PR #5, opened 2021-12-13
fix(NetPlayer): Prevent a crash when user select own player in singleplayer

xiaoxiao921:
.selected_player.net_player is null in singleplayer and its getting deferenced without being checked first


Yimura:
Thanks, code seems good!

PR #9, opened 2021-12-17
feat(Police): Fixed wanted level change.

revolutionxk:


Yimura:
Thanks, seems like a better way to reset and keep a wanted level locked!

PR #11, opened 2021-12-19
feat(Self): Better no ragdoll

karifeld:
much more simple & efficient


Yimura:
There's was reason I did no ragdoll the way I do it now and you'd be right to say it's incredibly ugly.

PR #21, opened 2021-12-26
added features

### LINKED ISSUE #20: window_handling, never calls ImGui::End
This if statement should call ImGui::End before the return statement.

https://github.com/Yimura/YimMenu/blob/9821a28e10e99dff1ab1ec109a49ec8390053e29/BigBaseV2/src/gui/window/window_handling.cpp#L14-L19

GibsonHF:
Include Persistcar, Give all weapons, clone player vehicle, teleport to waypoint hotkey, lsc tab, unlimited ammo and always jackpot casino with reset cooldown. Includes unfinished vehicle list spawner. also fixes https://github.com/Yimura/YimMenu/issues/20#issue-1088853932


GibsonHF:
Ignore all the previous git with vs was giving me a storm early on

Yimura:
I guess it's time to make some contributing guidelines and some other stuff to improve the flow of adding new features or doing bug fixes.

You seem pretty new with Git which is fine but make sure that you have set it up properly. Your name is in each commit, not sure if this is what you wanted. For it to work properly you need to set your Github username as your git name.

These commits can be squashed into one so you don't see what happened in between. Also you don't need to commit every 5 seconds whenever one feature is done you can commit after testing it's working correctly.

On the features you added I guess they're fine but I won't be accepting a PR like this. Try to work on a single feature for each PR as much as possible.

The hotkey feature is meh, I would probably work something out like all games have to make custom key binds for hotkeys so you can execute a preset function, not just assign a single hot key for teleport to waypoint.

In short you're adding waaaay too much at once. This is fine for a fork that drifts away from the main repo and that'll never merge into main.

There's so much more I could say because you've added so much in this one PR but I think you get the point.

GibsonHF:
Not a problem, yeah i've always kept things local and figured id give git a try but had lots of problems at the start. i appreciate the feedback and will take notes

Yimura:
Ye don't worry, my repo isn't setup with any guidelines for newcomers so there's no way to know how exactly stuff should be done.

I will be working on adding those for sure!

PR #27, opened 2021-12-29
feat(Teleport): Teleport to objective

ThatKakiGuy:
It works now 🎉

### DIFF
--- a/BigBaseV2/src/gui/window/main/tab_teleport.cpp
+++ b/BigBaseV2/src/gui/window/main/tab_teleport.cpp
@@ -1,7 +1,6 @@
 #include "main_tabs.hpp"
 #include "fiber_pool.hpp"
 #include "util/teleport.hpp"
-#include <vector>
 
 namespace big
 {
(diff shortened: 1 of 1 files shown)


Yimura:
Nice will take a look later on my computer.

Yimura (changes requested):
The button is still called waypoint.

Watch out with teleport to blip because it disregards the Z coordinate (height) and calculates its own based on ground coordinates.

Maybe make some kind of for loop with all the blip ID's instead of hardcoding them.

Also if one of the functions returns true exit the for loop this will prevent the others from executing and creating unnecessary calls.

ThatKakiGuy:
oh, right, I just copied the code from the waypoint button.

ThatKakiGuy:
fixed

Yimura (changes requested):
Looks much better.

Only could you use a static variable or static const and an int array instead of a vector. You don't need a dynamic array.

```cpp
static const int blips[] = { 1, 57, ... };
```

Aside from that I've got no complaints.

ThatKakiGuy:
Fixed :D

Yimura:
Ok now just remove the include for vector and you're good.

Yimura:
Perfect! I will merge it in master now.

PR #76, opened 2022-03-02
Session tab

### LINKED ISSUE #74: Move session join options
**Describe the bug**
Currently session join options have been moved under the "Self" tab but these don't belong here.

**Expected behavior**
It might be better to create a new view category for "lobby" options, where you could then use these same lobby join options.

xM4ddy:
#74

### DIFF
--- a/BigBaseV2/src/backend/backend.cpp
+++ b/BigBaseV2/src/backend/backend.cpp
@@ -61,13 +61,13 @@ namespace big
 			looped::weapons_delete_gun();
 			looped::weapons_force_crosshairs();
 			looped::weapons_gravity_gun();
-			looped::weapons_infinite_ammo();
 			looped::weapons_increased_damage();
+			looped::weapons_infinite_ammo();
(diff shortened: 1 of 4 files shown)


Yimura:
Bruh stop, leave dependency management to me. There's no need to update vendor currently.

Yimura:
Also a PR is for a single thing to be fixed, added or modified.

As the title of the PR states it's meant to fix issue #74 , but now you're also updating vendors, just stop.

xM4ddy:
Whoops didn't think this would go to here though it would keep the one I created the PR with. I probably need to make a new branch. Again new to github and making mistakes.

Yimura:
You can forcce push your branch to https://github.com/Yimura/YimMenu/pull/76/commits/f568d000b88345de718a35c4fbe323d2900036fa to fix the issue.

xM4ddy:
Sorry took me a sec.

Yimura (approved):

PR #104, opened 2022-03-10
F.A.R.T. Interface

### LINKED ISSUE #102: F.A.R.T. Interface
I'm always frustrated when a random kid walks up to me and explodes me. Then when I want to explode him using the menu, it takes a while to open the right player menu and go to the explode option.

What we could do is add Fast Acquiring Rendered Target, or F.A.R.T. for short. What F.A.R.T. would do is it would add a context menu (Like the one in gmod, see below for example) This way we can quickly do some stuff on the entity. Without having to open our menu.

**Provide Reasoning**
Because it makes the menu easier to use. It also has controller support, this way you can use some of the menu's features while using a controller.

**Additional context**
image

LiamD-Flop:
closes #102


GeopJr (approved):

Tayipc38 (approved):
I suggest that the context menu should have more basic options before initial release

VolkerSchlegel (approved):
Looks good to me, we should check if the entity is a player, because we can't delete a player so we shouldn't even display that option.

Yimura:
> Looks good to me, we should check if the entity is a player, because we can't delete a player so we shouldn't even display that option.

Don't worry this was an initial proof of concept internally we're already differentiating between different types of entities.

Yimura (approved):

PR #123, opened 2022-03-21
feat(session): Add Local Weather Options

xM4ddy:


xM4ddy:
Hmm some weird commit shit happened but still looks right.

BigBaseV2/src/views/view_session.cpp
```diff
@@ -64,6 +65,24 @@ namespace big
...
+				QUEUE_JOB_BEGIN_CLAUSE()
+				{
+					MISC::CLEAR_OVERRIDE_WEATHER();
+				}
+				QUEUE_JOB_END_CLAUSE
+			}
+
+			if (ImGui::ListBox("", &g->session.local_weather, session::weathers, 15))
+			{
+				session::local_weather();
```
LiamD-Flop: As you're using a native call this in a fiber_pool

LiamD-Flop (changes requested):
- [ ] Don't call natives in GUI thread

BigBaseV2/src/views/view_session.cpp
Yimura: Yes just the session_local weather call since that function uses a native.

LiamD-Flop (approved):
Weather be looking GOOD

PR #130, opened 2022-03-22
fix(local time): Smoof transition

xM4ddy:
^-^


Yimura:
I don't think this is behaving as intended.

1. When I modify time it's not smooth (unless you mean something else with the title)
2. If I disable the time override the original time doesn't apply from online
3. Re-enabling the time after it has been modified will reset the time back to the correct value

xM4ddy:
Cant be assed to worry tbh

Issue #117, opened 2022-03-19
Add OpenHookV (loading ASI mods)


Yimura:
**Is your feature request related to a problem? Please describe.**
Add OpenHookV / ScriptHookV support to YimMenu.

**Describe the solution you'd like**
BigBaseV2-fixed contains a branch which implements this exact feature.
https://bitbucket.org/gir489/bigbasev2-fix/commits/58d10c442e8d59c54edf3cb87cfffe968aa196cc

Thanks to @gir489returns for this.

**Provide Reasoning**
People keep asking for this feature, I tell them I know and so here is the issue to keep track of it and link to these people asking for it.

**Additional context**
Discord peeps

TheGreenBandit:
https://github.com/JackD111/OpenVHook this couls help on getting the asi loader up

FIying-Scotsman:
As a reference this has been done in BBV2 already:- https://bitbucket.org/gir489/bigbasev2-fix/commits/58d10c442e8d59c54edf3cb87cfffe968aa196cc

Should note - DX related modes **_do not work_** so some work would have to be put in to get it to share ImGui's DX hook.

FIying-Scotsman:
Should note for that SHV implementation linked above - DX related mods **_do not work_** so some work would have to be put in to get it to share ImGui's DX hook.

TheGreenBandit:
> https://github.com/JackD111/OpenVHook this couls help on getting the asi loader up

It would also be good to mention midt asi mods are detected for online like menyoo, as they use outdated things, so if you wanted to use asi's you would have to build undetected versions of them

FIying-Scotsman:
> > https://github.com/JackD111/OpenVHook this couls help on getting the asi loader up
> 
> It would also be good to mention midt asi mods are detected for online like menyoo, as they use outdated things, so if you wanted to use asi's you would have to build undetected versions of them

Menyoo patches the "is this thread networked?" check which is what gets you caught. For mods which are open source, rebuilding it without the patch is doable but not all mods are open source (and also do this as a way of keeping their mod strictly SP only).

While it would be harder to implement compared to just recompiling the mods - a more future-proof implementation would be to monitor for the patch (as Menyoo etc only patch it once on init) then 'undo' the patch back to the intended value immediately after it's been applied. 

[...]

PR #135, opened 2022-03-26
feat(ASI Loader):  Add ASI Loader

xM4ddy:
Has issues loading some ASIs mostly looks like ones that have a UI

#117


Yimura (approved):
Code looks nicely structured into YimMenu structure!

Yimura:
## ToDo:
 - [ ] Dynamically detect current DLL name when PE image patching
 - [ ] Restore ASIs on unload to prevent being unable to patch again if the current DLL name changes
 - [ ] DX Hooks for ASIs that have DX frontends

PR #139, opened 2022-03-30
Full API rewrite

Yimura:
This is a full API rewrite, as the current implementation use HTTP and not HTTPS.

The authenticate endpoint just gives a session as long as the RockstarID is known of the target user beforehand.

## New
 - HTTPS library (vendor)
 - URL parser/helper (vendor)
 - API rewrite

## ToDo
 - Currently calls to the API have been commented out
 - You don't get automatically signed in (opt-in)
 - Prepare Github Actions to include OpenSSL (1.1.1 as 3.X isn't supported yet)
 - Create instructions on installing OpenSSL

## References
 - https://github.com/Yimura/GTA5-Backend/pull/6


Sainan:
I cannot recommend using OpenSSL. It is the epitome of C developers. The DLL itself is a memory leak.

Yimura:
Yeah branch is dead, will probably never be merged in master.
What would you recommend though?

PR #146, opened 2022-04-24
get_blip_location fix

Aure7138:


Yimura:
Nice, will check it out when I've got the time.

PR #182, opened 2022-05-08
feat(Vehicle): Add auto drive

xM4ddy:
Should be working to auto drive to waypoint. Planning to add a thing to change the driving styles a bit and maybe add a wander or other driving stuff.

### DIFF
--- a/BigBaseV2/src/backend/backend.cpp
+++ b/BigBaseV2/src/backend/backend.cpp
@@ -85,6 +85,7 @@ namespace big
 
 		QUEUE_JOB_BEGIN_CLAUSE()
 		{
+			looped::vehicle_auto_drive();
 			looped::vehicle_despawn_bypass();
 			looped::vehicle_drive_on_water();
 			looped::vehicle_god_mode();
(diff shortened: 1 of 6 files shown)


LiamD-Flop (approved):

ghost:
🛠️
**On "Normal" mode, A-D doesn't hesitate to bump into traffic and some objects, even at low speeds.
On "Safe-ish", this ↑ does not happen, but it still tries to overtake stopped cars at red lights / stop signs.**

Yimura:
> 🛠️ **On "Normal" mode, A-D doesn't hesitate to bump into traffic and some objects, even at low speeds. On "Safe-ish", this ↑ does not happen, but it still tries to overtake stopped cars at red lights / stop signs.**

Pull Request are not issues, they are the work in progress on a feature. Please edit the contents of your existing issue #190 to readd to contents it previously had, I am unable to restore it.
Issues are issues for existing features, while Pull Requests are for new features that have already been worked out or still need to be worked out but still need to be merged into master.

This one has already been merged into master and should thus no longer receive any comments, if an issue relates to an issue you're making you can refer to it but that doesn't mean you should start commenting on that issue (that's why we separate it).

PR #191, opened 2022-05-10
feat(Improvement): Improved auto drive

### LINKED ISSUE #190: 🛠️ [to improve Auto-Drive]
**Describe the bug**
On "Normal" mode, A-D doesn't hesitate to bump into traffic and some objects, even at low speeds.
On "Safe-ish" mode, this ↑ does not happen, but it still tries to overtake stopped cars at red lights / stop signs.

**To Reproduce**
Nothing else to say, so far.

**Expected behavior**

Suggestion:
###
	static constexpr int driving_styles[] = { 443, 524349 };
	static constexpr char const* driving_style_names[] = { "Law-Abiding", "The Road Is Yours" };

**Screenshots**
❌

**Logs**
❌

**Build**
Select which applies for you:
- [x] I used YimMenu unmodified
- [ ] I use my own version of YimMenu

**Additional context**
❌

xM4ddy:
New settings for the auto drive styles that better reflect the options name. Also changes that are made to the settings while driving should take effect without needing to restart the auto drive option.

### DIFF
--- a/BigBaseV2/src/backend/backend.cpp
+++ b/BigBaseV2/src/backend/backend.cpp
@@ -86,7 +86,8 @@ namespace big
 
 		QUEUE_JOB_BEGIN_CLAUSE()
 		{
-			looped::vehicle_auto_drive();
+			looped::vehicle_auto_drive_to_waypoint();
+			looped::vehicle_auto_drive_wander();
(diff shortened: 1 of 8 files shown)


xM4ddy:
#190

PR #193, opened 2022-05-10
expand esp functionality

karifeld:
* changeable tracer position (float from 0.f to 1.f)
* color changing esp depending on distance
* health bar size relative to bounding box
* better color scheme (HUD_COLOR)
* shows players armor


Yimura (approved):

LiamD-Flop:
Why did you decide to change health bar on distance of the player?

LiamD-Flop:
With this type of health bar, you'd need to be on less then 30m from the player to see if he has lower health
image
image

LiamD-Flop (changes requested):
- [x] Health bar gets to small. Have it either minimum size or remove the scaling.
- [x] Distance based color should be able to toggle off.
- [x] When you disable the distance to be displayed it just removes the "|" but not the actual distance

karifeld:
sorry for the late response, to answer your question as for why the healthbar changes size

i run GTA 720p fullscreen, i made changes to it because if theres too many esp being rendered at the same location, it could easily block the view

heres what it looks like on 720p, now imagine this with more people being in that same spot
MH2LTmzfnL

but yes, i agree with the other things that you said
those should be togglable, i can try to work on it now

karifeld:
by "Better color scheme", i meant by changing the default ImU32 esp colors to the default hud colors defined in `common:/data/ui/hudcolor.dat` that is present in the game, see [https://docs.fivem.net/docs/game-references/hud-colors/](https://docs.fivem.net/docs/game-references/hud-colors/)

karifeld:
sorry if i sounded a bit weird, im not fluent in english
ill try to get the changes you requested to be done

karifeld:
the issues you mentioned should be addressed now @LiamD-Flop, let me know what you think

LiamD-Flop (approved):
Thanks for updating

PR #202, opened 2022-05-12
Delete Last Spawn

Aure7138:
Delete Last Spawn


Aure7138:
add ped list service

Aure7138:
put the ListBoxFooter to right place

PR #215, opened 2022-05-17
Fix RCOE-Patch leaving the game in a dirty state after unload

Sainan:


Yimura (approved):

Yimura:
Thanks a lot for your contribution, as of now this doesn't appear to be a problem to leave the game in a dirty state (in particular those bytes). It might be best to consider in the future to only patch the request control native when we actually require control and not permanently while the menu is loaded.

PR #223, opened 2022-05-21
feat(hotkeys): Teleport to waypoint

mentolixite:
I don't like the use of `GetAsyncKeyState()`, we should probably rely on `renderer::wndproc()` instead.

The problem is that calling `teleport::to_waypoint()` from there causes a crash.
Does perhaps the function call need to be queued in the thread pool?


xM4ddy:
Closed for same reason as #231. https://www.unknowncheats.me/forum/grand-theft-auto-v/476972-yimmenu-1-59-a-9.html#post3402800

PR #224, opened 2022-05-21
feat(Toxic): Send to Cayo Perico

mentolixite:


Yimura:
Unsure if we're going to accept any more toxic features, the PR can remain open but we're actively discussing on if we're still accepting toxic features in the future.

mentolixite:
Should I create a PR that adds a button for `toxic::bounty_player()` anyway?

xM4ddy:
Due to the mentioned reason it has been discussed and decided that this feature no longer fits in the vision of YimMenu. We will update the readme to reflect this and are probably going to remove the existing toxic options. Apologies for the change of policy.

Issue #119, opened 2022-03-20
Vehicle preview bug


Loko415:
**Describe the bug**
vehicle preview bug

**To Reproduce**
1. Go to spawner
2. Click on a vehicle
3. it shows the preview. search for another vehicle and its preview wont show. The old vehicle will only show. Fixed by closing and opening menu using insert

**Expected behavior**
it instantly switches to a new cars preview. Also noticed sometimes it shows the wrong car

**Screenshots**
none right now. But you can check yourself

**Logs**

no<paste your logs here>
no

**Build**
Select which applies for you:
- [ ] I used YimMenu unmodified
- [X] I use my own version of YimMenu

**Additional context**
none

Loko415:
i didnt assign you my bad idk why it said that

Yimura:
There might be a way to fix that but yes currently the vehicle preview is janky in certain ways.

1. Unloading while previewing leaves the vehicle behind
2. When scrolling the preview will sometimes show the incorrect vehicle, to fix this go of the list with your cursor and back on the list.
3. Swapping between tabs keeps the vehicles previewed

Mystro69:
#170 it should be fixed

Yimura:
> #170 it should be fixed

Yes but the issue remains open because of my comment mentioning other issues with the vehicle preview.
The main problem still being that the vehicle preview remains active when switching tabs.

LiamD-Flop:
Implemented by #226

PR #226, opened 2022-05-21
refactor: Modernized/optimized general code, bug fixing and more

karifeld:
refactor: Utilize self globals and gui::components in most code
fix(Vehicle Preview): addresses the issues mentioned in #119

fix(Infinite Clip): Fixed an issue where disabling it doesnt work
fix(No Ragdoll): Fixed an issue where disabling it doesnt work, removed unnecessary calls to natives as they are hardcoded to false for local ped.
fix(Spawn menu): Fixed placement of `ImGui::ListBoxFooter();` (It caused an issue where if `vehicles.json` was not present, The ListBox would be the wrong size)

fix: `self_gloabls.cpp` name typo

refactor(Mobile): Only call `TASK::CLEAR_PED_TASKS_IMMEDIATELY();` if pv_teleport_into global is true and player ped is in a vehicle
feat(Weapons): Remove current weapon

refactor: Added missing variable in calls to self globals
refactor: Utilize usage of ControllerInputs


xM4ddy:
Hey looking through the changes and wondering about your decision to do something like `Vehicle vehicle = self::veh;` rather than using `self::veh` throughout. (Not saying it's an issue just looking to clarify something.)

karifeld:
Well, its an optimization habit that ive carried over outside of Lua.

in Lua, if you were to run the following:
`for i = 1, 1000000 do 
   local x = math.sin(i) 
end`

`local sin = math.sin 
for i = 1, 1000000 do
    local x = sin(i) 
end`

The latter would run faster than the former. because Lua's main data structure is a table. if you were to a global function instead of a local, it will be slower because globals are stored in a table and are accessed via hash lookup. On the other hand, If you declare a local variable outside of a loop, the resulting code would be faster as they are accessed by index.

Now, im not expert but i dont think declaring a local variable really makes a difference in CPP.
which is why theres another reason, Readability and tidiness.

Compare both codes. one has a location variable declared and one does not.
	
	Vector3 vector;
	vector.x = self::pos.x * 2;
	vector.y = self::pos.y * 2;
	vector.z = self::pos.z * 2;
	
.

	Vector3 location = self::pos;
	
	Vector3 vector;
	vector.x = location.x * 2;
	vector.y = location.y * 2;
	vector.z = location.z * 2;
	
The latter looks neater and much more comprehensive than the former, Consistency is the most important aspect of a coding style. I think it should be best practiced even in other projects.

BigBaseV2/src/backend/looped/self/no_ragdoll.cpp
```diff
@@ -5,13 +5,6 @@ namespace big
...
 	void looped::self_no_ragdoll()
 	{
-		Ped player = PLAYER::PLAYER_PED_ID();
-
-		if (g->self.no_ragdoll) {
-			PED::SET_PED_CAN_RAGDOLL(player, !g->self.no_ragdoll);
-			PED::SET_PED_CAN_RAGDOLL_FROM_PLAYER_IMPACT(player, !g->self.no_ragdoll);
-			PED::SET_PED_RAGDOLL_ON_COLLISION(player, !g->self.no_ragdoll);
-		}
-
```
Yimura: You sure you want to always run this native, you're even executing the native when the feature is disabled, possibly causing side effects in missions. Have you properly tested this?

BigBaseV2/src/backend/looped/weapons/infinite_mag.cpp
```diff
@@ -5,8 +5,6 @@ namespace big
 {
 	void looped::weapons_infinite_mag()
 	{
-		if (g->weapons.infinite_mag) {
-			WEAPON::SET_PED_INFINITE_AMMO_CLIP(PLAYER::PLAYER_PED_ID(), g->weapons.infinite_mag);
-		}
```
Yimura: Forcefully running this native each tick even when disabled might cause side effects.

BigBaseV2/src/backend/looped/vehicle/fly.cpp
```diff
@@ -1,6 +1,7 @@
 #include "backend/looped/looped.hpp"
 #include "natives.hpp"
 #include "util/entity.hpp"
+#include "gta/enums.hpp"
```
Yimura: Sort includes alphabetically

Yimura (changes requested):
In general all looks good, if you resolve the few comments I have here and there then we can merge it into master.

Also CI is failing, fix the error the below error:
`src\backend\looped\vehicle\fly.cpp(14): error C4700: uninitialized local variable 'ped' used`

maybegreat48:
Infinite ammo (not infinite clip) can be simplified to ```WEAPON::SET_PED_INFINITE_AMMO(self::ped, g.weapons.infinite_ammo, NULL);```

karifeld:
@Yimura Changes have been made, let me know what you think.

@maybegreat48  Thanks for the contribution.

Yimura (approved):

Yimura:
If there are no more changes to be added then it's ready to be merged into master.

karifeld:
That should be it, you may merge at anytime

Yimura:
Alright @xM4ddy will take one final look at it and then it'll be merged.

xM4ddy:
> Infinite ammo (not infinite clip) can be simplified to `WEAPON::SET_PED_INFINITE_AMMO(self::ped, g.weapons.infinite_ammo, NULL);`

I'm trying to remember why I did that. It was actually a conscious choice actually. I may have just been some weird preference for the behavior of doing it the way I did.

Yimura:
There might be some odd scenario where the game resets the toggle.

xM4ddy:
Both seem to work so if I remember what it was and its important I'll just make another commit.

PR #244, opened 2022-05-23
Name and status variables for the Script system.

Tayipc38:
A simple change to the script system of bbv2 which will allow in-game manipulation.


BigBaseV2/src/script.hpp
```diff
@@ -8,20 +8,24 @@ namespace big
...
+		explicit script(func_t func, const char* name, bool* enabled, std::optional<std::size_t> stack_size = std::nullopt);
 		~script();
 
 		void tick();
 		void yield(std::optional<std::chrono::high_resolution_clock::duration> time = std::nullopt);
 		static script* get_current();
 		static void script_exception_handler(PEXCEPTION_POINTERS exp);
+		const char* m_name;
+		void toggle_script(bool toggle) { if (m_enabled != nullptr) *this->m_enabled = toggle; }
+		bool* get_enabled_pointer() { return m_enabled; }
```
Yimura: Instead of returning the pointer to the bool just dereference the pointer here and return its value.
Additionally you could check if the enabled pointer is valid before dereferencing and return true/false if its a nullptr.

Then instead of calling `get_enabled_pointer()` twice each time you need to check if a script is enabled, instead you could just call `is_enabled()`.

```cpp
bool is_enabled() const
{
    if (m_enabled == nullptr)
        return false;
    return *m_enabled;
}
```

BigBaseV2/src/script_mgr.hpp
```diff
@@ -12,8 +25,15 @@ namespace big
...
-
 		void tick();
+		std::vector<script*> get_scripts_data() {
+			std::vector<script*> retvalue{};
+			for (auto& s : m_scripts) { 
+				if(s->get_enabled_pointer())
+					retvalue.push_back(s.get());
+			}
+			return retvalue;
+		}
```
Yimura: Instead of creating a new vector you can return the entire script vector by reference, though then you'll have to filter through the scripts in the UI.

```cpp
std::vector<std::unique_ptr<script>>& get_scripts_data()
{
    return m_scripts;
}
```

BigBaseV2/src/script_mgr.hpp
```diff
@@ -4,6 +4,19 @@
...
+	inline bool g_script_weapons = true;
+	inline bool g_script_vehicles = true;
+	inline bool g_script_misc = true;
+	inline bool g_script_remote = true;
+	inline bool g_script_noclip = true;
+	inline bool g_script_lscustoms = true;
+	inline bool g_script_vehiclefly = true;
+	inline bool g_script_rgbrandomizer = true;
+	inline bool g_script_contextmenu = true;
+	inline bool g_script_turnsignal = true;
```
Yimura: In general I dislike having global variables like these but for now I guess it is fine.

No need to change it as of now.

BigBaseV2/src/script.hpp
```diff
@@ -8,20 +8,24 @@ namespace big
...
 	public:
-		explicit script(func_t func, std::optional<std::size_t> stack_size = std::nullopt);
+		explicit script(func_t func, const char* name, bool* enabled, std::optional<std::size_t> stack_size = std::nullopt);
 		~script();
 
 		void tick();
 		void yield(std::optional<std::chrono::high_resolution_clock::duration> time = std::nullopt);
 		static script* get_current();
 		static void script_exception_handler(PEXCEPTION_POINTERS exp);
+		const char* m_name;
```
Yimura: Use `const std::string_view`, string view offers more flexibility as you can pass `std::string` and `char[]` interchangeably it's also more efficient than `std::string` as string view doesn't make a copy of the value but instead references the original storage location of the string.

BigBaseV2/src/script.hpp
```diff
@@ -8,20 +8,24 @@ namespace big
 	public:
 		using func_t = void(*)();
 	public:
-		explicit script(func_t func, std::optional<std::size_t> stack_size = std::nullopt);
+		explicit script(func_t func, const char* name, bool* enabled, std::optional<std::size_t> stack_size = std::nullopt);
```
Yimura: Read below comment about `std::string_view`.

Yimura (changes requested):

BigBaseV2/src/script.hpp
Tayipc38: I don't think you quite understood what the pointer stands for. Look close enough and youll notice that some scripts do not have a bool. When this is the case the script will run regardless and to check this I need the pointer not the value.

image

image

BigBaseV2/src/script_mgr.hpp
Tayipc38: The reason for this little detour is to disregard scripts that require no bool and have no ability to be disabled. Such as the main scripts. Weeding those out before fetching will reduce confusion when using this function . Case: Fetching the scripts and accessing their bool pointer.

BigBaseV2/src/script_mgr.hpp
Tayipc38: This can be greatly optimized, i admit.

BigBaseV2/src/script.hpp
Yimura: Ye you could change the default behaviour of the is_enabled or add another member function to get the pointer. My point stands that currently you need to make 2 calls for something that can be handled internally by the script member function instead of having duplicate code of checking the pointer each time.

BigBaseV2/src/script_mgr.hpp
Yimura: Yes just like the above comment, you can add another member function to get the pointer then in the UI you can filter out the functions that don't have the enabled pointer, instead this way you're making an std::vector each frame your menu renders instead of passing by reference and just filtering out the scriptz that don't have the pointer.

BigBaseV2/src/script.hpp
Tayipc38: As a side note, the boolean can be defaulted within the function so passing a nullptr won't be necessary.

Tayipc38: > Ye you could change the default behaviour of the is_enabled or add another member function to get the pointer. My point stands that currently you need to make 2 calls for something that can be handled internally by the script member function instead of having duplicate code of checking the pointer each time.

2 calls? A dereference can be employed to both of these things:

`script::get_current()->get_enabled_pointer() //Pointer
*script::get_current()->get_enabled_pointer()//Value`

Surely it would be more verbose to have the pointer to the boolean than only its value

BigBaseV2/src/script_mgr.hpp
Tayipc38: > Yes just like the above comment, you can add another member function to get the pointer then in the UI you can filter out the functions that don't have the enabled pointer, instead this way you're making an std::vector each frame your menu renders instead of passing by reference and just filtering out the scriptz that don't have the pointer.

I am not knowledgable of how taxing a vector creation would be to utilize each frame, but I can outsource the task to the usage thereof: 

image

Note that some small changes would be required as this would not be accessing a vector of script pointers but rather the following at this point:

`std::vector<std::unique_ptr<script>> m_scripts;`

BigBaseV2/src/script.hpp
Tayipc38: > Ye you could change the default behaviour of the is_enabled or add another member function to get the pointer. My point stands that currently you need to make 2 calls for something that can be handled internally by the script member function instead of having duplicate code of checking the pointer each time.

Thats also the reason for the creation of the vector that only includes scripts that have bools. To avoid having to check for a valid pointer.

PR #261, opened 2022-05-29
feat(ViewPlayer): Give health, armour, ammo, weapons

mentolixite:
Please note that this doesn't work for local player.


Yimura (approved):

MoneyDropLobby:
REWARD_MONEY_VARIABLE how does that work? I need a little explanation

Yimura (changes requested):
Feature doesn't seem to work as intended, I tried to give all weapons but seems to only give 3 of the handguns and then it stops.
Maybe yield between weapons...

Neither does this feature work on the local player (which is fine).

mentolixite:
> Feature doesn't seem to work as intended, I tried to give all weapons but seems to only give 3 of the handguns and then it stops.

There appears to be some kind of cooldown, because when you spam health and armor it suddenly stops giving the reward.

> Maybe yield between weapons...

What do you mean?

Yimura:
> What do you mean?

Yielding between giving each weapon or between every 3 gifts.

Yimura:
In the current state I can't merge this feature into YimMenu please modify it so it's usable.

Yimura (changes requested):
1. Throwables aren't dropped with the `Give Weapons` button.
2. Dropping all ammo types seems like it only gives the clip size
3. Spamming the feature can crash a player

mentolixite:
> 1. Throwables aren't dropped with the `Give Weapons` button.
> 2. Dropping all ammo types seems like it only gives the clip size

Throwables?

> 3. Spamming the feature can crash a player

Should we implement a rate limit?

Yimura:
> Throwables?

Yes grenades, sticky bombs, proximity minies, etc..

> Should we implement a rate limit?

Yes we can't really release a feature which can crash players, either yield long enough that it no longer happens. Though you'll have to use some kind of global rate limit for pickups.

mentolixite:
> Yes grenades, sticky bombs, proximity minies, etc..

By "dropped" you mean "given"?

> Yes we can't really release a feature which can crash players, either yield long enough that it no longer happens. Though you'll have to use some kind of global rate limit for pickups.

We could store a timestamp in the `player` object and then compare against that.

Yimura:
> By "dropped" you mean "given"?

Well ye, same same.

> We could store a timestamp in the `player` object and then compare against that.

It might be time to move to service like structure perhaps?
I don't want to store values like that in the player object, though my vision on that could change in the future.
I would just have a global lock for giving pickups to players using some kind of service.

Clone69420:
To drop it to someone else it have to be synced

Yimura:
> To drop it to someone else it have to be synced

Context? It was previously established that this doesn't work on your local player.

Clone69420:
> > To drop it to someone else it have to be synced
> 
> Context? It was previously established that this doesn't work on your local player.

All pickups given by you to yourself or someone in your lobby it only works by syncing the pickup thats basicly the bypass for rockstars patch. But on your local player iam not quiet sure if u need to sync it too

Clone69420:
https://ibb.co/VvMW4h3
how to sync hashes ^^

Yimura:
> https://ibb.co/VvMW4h3 how to sync hashes ^^

And what favour are you doing by sharing that here?
How does it relate to this topic?
If you want to drop money it will never be added to YimMenu and that screenshot isn't useful without telling us the edition of your game, your base address and other information.

Neither is this signature related to `OBJECT::CREATE_AMBIENT_PICKUP`, this syncs directly with a player.
Unless I'm misunderstanding something.

Yimura (dismissed):
I changed the design to a service and increased the yield for ammo/weapon drops to 20ms and players do not appear to be crashing anymore with this.
Even when I was spamming the feature.

Yimura (approved):

PR #231, opened 2022-05-21
feat(Hotkeys) Added Hotkey support

MoneyDropLobby:

PR #264, opened 2022-06-01
feat(hotkeys): Teleport to waypoint, Teleport to objective

Aure7138:


Yimura:
As the previous PR's related to hotkeys, I will reject this PR with the same reasoning.

#223
#231 

I dislike how GetAsyncKeyState works and how you need to setup a single function that runs in a loop.

PR #271, opened 2022-06-04
feat(PlayerList): Show self at the top, separated by a divider

mentolixite:


Yimura:
While looking at your code I decided to redesign the player service as I was pretty unhappy of how it was implement (not your code, just the player service in general).

PR #284, opened 2022-06-24
Fix player still being able to perform certain in-game actions even though they are not supposed to

xiaoxiao921:
Multiple menu features disable in-game actions with native calls like `DISABLE_CONTROL_ACTION`.
Currently most of them are not working as expected.
My understanding is that some of the looped scripts that these natives are used in are "too big" and are not fast enough to be called every frame, making `DISABLE_CONTROL_ACTION` not work correctly.

Fix: Make `DISABLE_CONTROL_ACTION` calls in their own script thread

### DIFF
--- a/BigBaseV2/src/backend/backend.hpp
+++ b/BigBaseV2/src/backend/backend.hpp
@@ -2,18 +2,21 @@
 #include "common.hpp"
 #include "looped/looped.hpp"
 #include "script.hpp"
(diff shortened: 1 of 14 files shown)


Yimura:
Oh cool that's a great way of resolving that issue.

Seems lightweight as well.

Yimura (approved):

PR #289, opened 2022-06-27
Desync kick / MsgNetComplaint:

xiaoxiao921:
- range check player array
- null check player_ptr

### DIFF
--- a/BigBaseV2/src/hooks/receive_net_message.cpp
+++ b/BigBaseV2/src/hooks/receive_net_message.cpp
@@ -27,7 +27,6 @@ namespace big
 
 	bool hooks::receive_net_message(void* netConnectionManager, void* a2, rage::netConnection::InFrame* frame)
 	{
-
 		if (frame->get_type() == 4)
 		{
 			rage::datBitBuffer buffer((uint8_t*)frame->m_data, frame->m_length);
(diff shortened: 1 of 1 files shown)


Yimura (approved):

BigBaseV2/src/hooks/receive_net_message.cpp
```diff
@@ -59,16 +58,26 @@ namespace big
...
 							}
 
 						}
 
 						buffer.Seek(0);
 
-						players.at(0)->m_complaints = 10000;
+						if (!players.empty())
+						{
+							const auto player = players.at(0);
```
Yimura: Pass by reference?

BigBaseV2/src/hooks/receive_net_message.cpp
```diff
@@ -59,16 +58,26 @@ namespace big
...
+							if (player && player->is_valid())
+							{
+								player->m_complaints = 65535;
+								g_notification_service->push_warning("Blocked Kick", std::string("Blocked desync kick from ") + players.at(0)->get_name());
```
Yimura: Back to players.at(0) instead of u using player?

Yimura (changes requested):

BigBaseV2/src/hooks/receive_net_message.cpp
Yimura: Reference isn't really required since it's a pointer anyways.

Yimura (approved):

PR #290, opened 2022-06-27
Add vehicle jump

xiaoxiao921:


BigBaseV2/src/backend/looped/vehicle/vehicle_jump.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include "gta/enums.hpp"
+#include "natives.hpp"
+
+namespace big
+{
+	void looped::vehicle_jump()
+	{
+		if (!g->vehicle.vehicle_jump) return;
+
+		const auto& vehicle = self::veh;
```
Yimura: Passing this value by reference still causes a lookup in the global table, best to assign by value.

Yimura (changes requested):

Yimura (approved):

PR #291, opened 2022-06-27
Feat vehicle gun improv 

xiaoxiao921:
- add some delay between shoots so that freshly spawned car doesnt replace the last one
- mark veh as no longer needed so that we can always keep shooting
- fix game input not correctly disabled while typing vehicle gun model text

### DIFF
--- a/BigBaseV2/src/backend/looped/weapons/vehicle_gun.cpp
+++ b/BigBaseV2/src/backend/looped/weapons/vehicle_gun.cpp
@@ -3,7 +3,6 @@
 #include "util/math.hpp"
 #include "util/vehicle.hpp"
 #include <gui.hpp>
-#include <chrono>
 
 namespace big
 {
--- a/BigBaseV2/src/backend/looped/weapons/vehicle_gun.cpp
+++ b/BigBaseV2/src/backend/looped/weapons/vehicle_gun.cpp
@@ -2,7 +2,7 @@
 #include "core/enums.hpp"
 #include "util/math.hpp"
 #include "util/vehicle.hpp"
-#include <gui.hpp>
+#include "gui.hpp"
 
 namespace big
 {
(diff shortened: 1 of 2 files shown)


BigBaseV2/src/backend/looped/weapons/vehicle_gun.cpp
```diff
@@ -2,47 +2,62 @@
 #include "core/enums.hpp"
 #include "util/math.hpp"
 #include "util/vehicle.hpp"
+#include <gui.hpp>
+#include <chrono>
```
Yimura: Chrono is already included in common.hpp, redundant include?

BigBaseV2/src/backend/looped/weapons/vehicle_gun.cpp
```diff
@@ -2,47 +2,62 @@
...
+				dist * sin(pitch)
+			};
+
+			ENTITY::SET_ENTITY_ROTATION(veh, rot.x, rot.y, rot.z, rotation_order, 1);
+			ENTITY::SET_ENTITY_VELOCITY(veh, velocity.x, velocity.y, velocity.z);
+
+			// flagging the veh as no longer needed so that the game can remove it
+			// when reaching the maximum vehicle limit,
+			// allowing the vehicle gun to keep working
+			ENTITY::SET_VEHICLE_AS_NO_LONGER_NEEDED(&veh);
```
Yimura, line 57: Probably should be moved the vehicle::spawn helper function.

Though not necessarily in this PR.

Yimura (changes requested):

BigBaseV2/src/backend/looped/weapons/vehicle_gun.cpp
```diff
@@ -2,47 +2,61 @@
 #include "core/enums.hpp"
 #include "util/math.hpp"
 #include "util/vehicle.hpp"
+#include <gui.hpp>
```
Yimura: Why use `<>` on your include instead of `""` include?

Usually you'd use `<>` for Standard library headers and `""` for user defined headers.
It's better to use `""` for our own headers.

Yimura (changes requested):

BigBaseV2/src/backend/looped/weapons/vehicle_gun.cpp
xiaoxiao921: Good question, didnt even realize, I blame cpp for that shitty ass decade old import system

Yimura (approved):

PR #293, opened 2022-06-27
feat context_menu_service

xiaoxiao921:
- add box around selected entity
- add back the los check I removed by mistake
- add option to select entity types 
- show reticle in middle of screen

preview:
image


BigBaseV2/src/services/context_menu_service.cpp
```diff
@@ -69,6 +69,70 @@ namespace big
...
+
+		GRAPHICS::DRAW_LINE(edge5.x, edge5.y, edge5.z, edge6.x, edge6.y, edge6.z, 0, green, blue, 200);
+		GRAPHICS::DRAW_LINE(edge5.x, edge5.y, edge5.z, edge8.x, edge8.y, edge8.z, 0, green, blue, 200);
+		GRAPHICS::DRAW_LINE(edge6.x, edge6.y, edge6.z, edge7.x, edge7.y, edge7.z, 0, green, blue, 200);
+		GRAPHICS::DRAW_LINE(edge7.x, edge7.y, edge7.z, edge8.x, edge8.y, edge8.z, 0, green, blue, 200);
+
+		GRAPHICS::DRAW_LINE(edge1.x, edge1.y, edge1.z, edge7.x, edge7.y, edge7.z, 0, green, blue, 200);
+		GRAPHICS::DRAW_LINE(edge2.x, edge2.y, edge2.z, edge8.x, edge8.y, edge8.z, 0, green, blue, 200);
+		GRAPHICS::DRAW_LINE(edge3.x, edge3.y, edge3.z, edge5.x, edge5.y, edge5.z, 0, green, blue, 200);
+		GRAPHICS::DRAW_LINE(edge4.x, edge4.y, edge4.z, edge6.x, edge6.y, edge6.z, 0, green, blue, 200);
```
Yimura: We don't use natives to draw anything in-game, please use ImGui instead.

You can take a look at the ESP how it is made https://github.com/Yimura/YimMenu/blob/master/BigBaseV2/src/views/esp/view_esp.cpp

The context menu also has a UI function so if you need to draw anything you can do it there.
https://github.com/Yimura/YimMenu/blob/master/BigBaseV2/src/views/view_context_menu.cpp

Yimura (changes requested):

BigBaseV2/src/services/context_menu_service.cpp
```diff
@@ -2,6 +2,7 @@
 #include "natives.hpp"
 #include "pointers.hpp"
 #include "gta/replay.hpp"
+#include <gui.hpp>
```
Yimura: 🤔 `<>`

Yimura (changes requested):
Looks good, will definitely test this one before merging it though. (after you've made the changes)

BigBaseV2/src/services/context_menu_service.cpp
xiaoxiao921: Ah, I know why now, Intellisense default to this when you use for auto resolve missing import

Yimura: Bad defaults 😏

Yimura (approved):
Looks good, I tested it and had no problems.

Also nice addition where you can choose which entities are selectable.

PR #297, opened 2022-06-29
feat: switch to issue forms

GeopJr:
By switching to issue forms, fields can be/are required making autocloser useless.
Additionally it's probably more user friendly as users don't have to deal with markdown.

You can give it a try over at: https://github.com/geopjr-forks/YimMenu/issues/new/choose

Main changes to the current ones:

- added "Store" dropdown
- added a default title ([Bug]: ) (Forms wouldnt work without or an empty title)
- changed the "I used YimMenu unmodified" section to just a single checkbox asking users to make sure it occurs on an unmodified version of yimmenu


Yimura (approved):
Seems good to me :+1:

GeopJr:
(fpush: forcing render: markdown broke the format so removed)

Issue #258, opened 2022-05-27
More protection against crashes


cristyanul:
I found this list of crash scripts and I think it would help at protecting against them. Out of them, we have **Script Event Crash 2,       {0: =-1386010354} ** already
```
[Kick]
MARK_AS_MODDER
;name				params
KICK 0,				{0: =0x493fc6bb}

[Malicious Crash]
MARK_AS_MODDER
Script Event Crash 1,       {0: =962740265}
Script Event Crash 2,       {0: =-1386010354} 
Script Event Crash 5,       {0: =603406648}
Script Event Crash 6 v1,       {0: =704979198, 2: <=-1 | >=1}
Script Event Crash 6 v2,       {0: =704979198, 3: <=-2 | >=0}
Script Event Crash 6 v3,       {0: =704979198, 4: <=3 | >=5}
Script Event Crash 6 v4,       {0: =704979198, 5: <=172 | >=174}
Script Event Crash 6 v5,       {0: =704979198, 6: <=20 | >=510}
Script Event Crash 6 v6,       {0: =704979198, 7: <=62 | >=64}
Script Event Crash 6 v7,       {0: =704979198, 11: <=-1 | >=1}
Script Event Crash 6 v8,       {0: =704979198, 12: <=-1 | >=1}
Script Event Crash 6 v9,       {0: =704979198, 13: <=-1 | >=1}
Script Event Crash 7,       {0: =-1715193475}
Script Event Crash 8,       {0: =1258808115}
Script Event Crash 10,      {0: =998716537}
Script Event Crash 11,      {0: =163598572}
Script Event Crash 12 v1,      {0: =-1970125962, 2: <=-1 | >=50}
Script Event Crash 12 v2,      {0: =-1970125962, 3: <=-1 | >=50}
Script Event Crash 13,      {0: =-1056683619}
Script Host Crash,      {0: =1757755807}
Uh oh spaghettios,      {0: =-393294520}

[Crash 2]
Script Event Crash 9 v1,       {0: =2112408256, 2: <=-1986324736 | >=1747413822}
Script Event Crash 9 v2,       {0: =2112408256, 2: >-9999999, 2: <77777777}
Script Event Crash 9 v3,       {0: =2112408256, 3: <=-1986324736 | >=1777712108}
Script Event Crash 9 v4,       {0: =2112408256, 3: >-9999999, 3: <77777777}
[...]

PR #302, opened 2022-07-01
Implemented Issue #258

aa15032261:


maybegreat48:
Am a bit curious on how the huge amount of TSE hashes will be updated. I generally have a comment with some code/strings that will not change for non-public events. Also this will protect against the broadcast array size overwrite exploit someone uploaded to github a couple of months ago

DaniTRDev:
> Am a bit curious on how the huge amount of TSE hashes will be updated. I generally have a comment with some code/strings that will not change for non-public events. Also this will protect against the broadcast array size overwrite exploit someone uploaded to github a couple of months ago

Through pattern scanning inside scripts, I've made a script disassembler & assembler that can generate good signatures(it needs to be updated since I didn't add the new opcode bittest).
If anyone wants it I can make the repo public.

maybegreat48:
I made something similar [here](https://github.com/maybegreat48/ScrUpdate) but script events are not supported for now. Perhaps some kind of xref and switch case finder might allow that to be implemented

DaniTRDev:
Mine can be more or less adapted, I'll post the link here in a few minutes.

> I made something similar here <https://github.com/maybegreat48/ScrUpdate>
> but script events are not supported for now. Perhaps some kind of xref and
> switch case finder might allow that to be implemented
>

DaniTRDev:
I've been busy until now. Here's the [link](https://github.com/DaniTRDev/EzScriptCore) to the public repo.

BigBaseV2/src/hooks/script_event_handler.cpp
```diff
@@ -6,7 +6,7 @@ namespace big
 	void format_string(std::string_view player_name, std::string_view protection_type, bool should_log, bool should_notify)
 	{
 		if (should_log)
-			LOG(WARNING) << "BLOCKED_SCRIPT_EVENT";
+			LOG(WARNING) << fmt::format("BLOCKED_SCRIPT_EVENT From: {} Event Type: {}", player_name.data(), protection_type.data());
```
Yimura: Couldn't this be changed to
```cpp
LOG(WARNING) << "BLOCKED_SCRIPT_EVENT From: " << player_name << " Event Type: " << protection_type;
```

Yimura (changes requested):
Overall looks good, it appears that the previous TSE crashes were incorrectly implemented as well (on their notification part), seems I missed that and thanks for fixing that in this PR.

BigBaseV2/src/hooks/script_event_handler.cpp
aa15032261: Should've done this, don't know what I was thinking

Yimura (approved):

PR #303, opened 2022-07-01
Added bulletproof feature

aa15032261:
Added bulletproof feature which locks player's hp and armor to 100%.


Yimura:
If you can you can just add the feature to the bullet proof branch and then force push it.

PR #304, opened 2022-07-01
Added bulletproof feature.

aa15032261:
Added bulletproof feature which locks player's hp and armor to 100%.


BigBaseV2/src/backend/looped/self/bulletproof.cpp
```diff
@@ -0,0 +1,13 @@
...
+
+namespace big
+{
+	void looped::self_bulletproof()
+	{
+		if (g->self.bulletproof) {
+			ENTITY::SET_ENTITY_HEALTH(self::ped, PED::GET_PED_MAX_HEALTH(self::ped), 0);
+			PED::SET_PED_ARMOUR(self::ped, PLAYER::GET_PLAYER_MAX_ARMOUR(self::id));
+		}
+	}
```
Yimura: There's a better method than using natives in a loop, if your HP is drained fast enough you'll most likely still die.

```cpp
enum class eEntityProofs
{
	BULLET = 1 << 4,
	FIRE = 1 << 5,
	COLLISION = 1 << 6,
	MELEE = 1 << 7,
	GOD = 1 << 8,
	EXPLOSION = 1 << 11,
	STEAM = 1 << 15,
	DROWN = 1 << 16
};

g_local_player->m_damage_bits |= static_cast<int>(eEntityProofs::BULLET);
```
This code achieves the same or works even better while not using any natives at all.

Don't forget to check if `g_local_player != nullptr`

Yimura (changes requested):

BigBaseV2/src/backend/looped/self/bulletproof.cpp
Aure7138: add a waterproof
`waterproof = 1 << 24`

aa15032261: I'm curious what's the different between ENTITY::SET_ENTITY_INVINCIBLE and the god mode mask?

aa15032261: > add a waterproof `waterproof = 1 << 24`

I will add this to the file, thanks.

Yimura (approved):

BigBaseV2/src/backend/looped/self/godmode.cpp
```diff
@@ -15,5 +15,10 @@ namespace big
 
 			bLastGodMode = g->self.godmode;
 		}
+
+		if (g_local_player != nullptr)
+		{
+			g_local_player->m_damage_bits = g->self.proof_mask;
+		}
```
Yimura, line 19: Applying the proofs after god mode causes god mode to be disabled, since god mode uses `1 << 8`, either add god mode to the proof mask or call the god mode code after applying the mask.

I'd prefer you went with adding god mode to the mask.

Yimura (changes requested):

BigBaseV2/src/backend/looped/self/godmode.cpp
aa15032261: Fixed. Please check the latest commit.

Yimura (approved):

PR #316, opened 2022-07-04
fix(ViewPlayer): Make "never wanted" feature independent for each player

mentolixite:


Yimura:
@xM4ddy / @SkiddyToast how would you implement this sort of thing?
I'm intrigued in your implementation idea and not really disapproving this design, I myself have no idea how I'd do it.

SkiddyToast:
I practically do the same thing as this implementation. Add a bool to the player class and check the bool while looping over the players.

Yimura (approved):
As of now these changes are alright but I might revise where these values are stored.

PR #319, opened 2022-07-05
Added model hashtable

Yimura:
[Inspired by the code from MenyooSP](https://github.com/MAFINS/MenyooSP/blob/2c2ce48ba9379c8028bdb5c29d4985de93dc9648/Solution/source/Memory/GTAmemory.cpp#L836-L883).
## What is this?
The game has a structure in which it saves all the model info's.
There's a lookup table which is used to go from joaat to HashNode, though the first HashNode doesn't always hold the expected hash, thus you need to traverse the HashNode next members till you match the hash.
A HashNode holds the index of the item which can be used to access the member you need.

## Classes

```cpp
#pragma pack(push, 1)
class HashNode
{
public:
	int32_t m_hash; //0x0000
	uint16_t m_idx; //0x0004
	char pad_0006[2]; //0x0006
	HashNode* m_next; //0x0008
}; //Size: 0x0010
static_assert(sizeof(HashNode) == 0x10);

template<typename T>
class HashTable
{
public:
	T* m_data; //0x0000
	uint16_t m_size; //0x0008
	char pad_000A[14]; //0x000A
	uint64_t m_item_size; //0x0018
	char pad_0020[64]; //0x0020
	HashNode** m_lookup_table; //0x0060
	uint16_t m_lookup_key; //0x0068
}; //Size: 0x006A
// static_assert(sizeof(HashTable<void*>) == 0x6A); // compiler gives assert error without telling me what the problem is, the class is correct though.
#pragma pack(pop)
```

Link: https://github.com/Yimura/GTAV-Classes/blob/master/HashTable.hpp

## Usage
```cpp
// sample code to iterator models
for (int i = 0; i < m_model_table->m_size; ++i)
{
	for (auto node = m_model_table->m_lookup_table[i]; node; node = node->m_next)
	{
		if (const auto table_idx = node->m_idx; table_idx < m_model_table->m_size)
		{
			if (const auto model = m_model_table->m_data[table_idx]; model && model->m_model_type == eModelType::Vehicle)
			{

			}
		}
	}
}

// sample code to get a specific model
auto adder_hash = RAGE_JOAAT("adder");
for (auto i = m_model_table->m_lookup_table[adder_hash % m_model_table->m_lookup_key]; i; i = i->m_next)
{
	if (i->m_hash == adder_hash)
	{
		if (const auto model = m_model_table->m_data[i->m_idx]; model)
		{
			LOG(G3LOG_DEBUG) << "Found Model: " << HEX_TO_UPPER(model->m_model_hash) << " => type: " << (int)model->m_model_type;

			break;
		}
	}
}
```

### DIFF
--- a/BigBaseV2/src/pointers.cpp
+++ b/BigBaseV2/src/pointers.cpp
@@ -1,9 +1,6 @@
 #include "common.hpp"
 #include "pointers.hpp"
-
-#include "thread_pool.hpp"
 #include "memory/all.hpp"
-#include "util/system.hpp"
 
 namespace big
 {
--- a/BigBaseV2/src/pointers.cpp
+++ b/BigBaseV2/src/pointers.cpp
@@ -1,6 +1,9 @@
 #include "common.hpp"
 #include "pointers.hpp"
+
+#include "thread_pool.hpp"
 #include "memory/all.hpp"
+#include "util/system.hpp"
 
 namespace big
 {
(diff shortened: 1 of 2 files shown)


BigBaseV2/src/pointers.cpp
```diff
@@ -1,6 +1,9 @@
 #include "common.hpp"
 #include "pointers.hpp"
+
+#include "thread_pool.hpp"
```
SkiddyToast: Unused include

SkiddyToast (changes requested):

SkiddyToast (approved):
Looks good to me.

PR #325, opened 2022-07-06
fix(Notifications): Added titles to duplicate check

### LINKED ISSUE #249: Notification service duplicate detection bug
**Describe the bug**
The notification service denies duplicate messages to be queued. However the manner of how it concludes duplicates is faulty. It seems to only check for the content and not for the accompanied title wich results in undesired behaviour.

**To Reproduce**
1. Open yimmenu
2. Spam any message that has a different title but the same message
3. Notice how different messages with the same content but different title do not get queued.

**Expected behavior**
By all means, the manner of how the service should detect identical messages is to include the title or maybe a Primary Key.

**Screenshots**
Cant really show screenshots as the bug is of invisible nature.

**Logs**

```
<paste your logs here>
```

**Build**
Select which applies for you:
- [x] I used YimMenu unmodified
- [ ] I use my own version of YimMenu

**Additional context**
-

LiamD-Flop:


Yimura (approved):

LiamD-Flop:
Closes #249

PR #338, opened 2022-07-15
Fixed an issue where LSC is not working for some vehicles.

aa15032261:


Yimura:
Will check it later today, I noticed as well that some vehicles their mod kits wouldn't load or apply.

Yimura (approved):

Issue #253, opened 2022-05-25
Spawn Max Vehicle Crash


Vlhss:
**Describe the bug**
When spawning certain cars while having the spawn max option enabled it crashes

**To Reproduce**
1. Go to Vehicle->Spawn
2. Search the car Banshee
3. Enable the spawn maxed option
4. Spawn the car
5. Crash

**Expected behavior**
Not to crash

**Screenshots**
-

**Logs**
-

**Build**
Select which applies for you:
- [ ] I used YimMenu unmodified
- [x] I use my own version of YimMenu

**Additional context**
While I do use my own version of YimMenu I haven't tweaked with vehicles yet. Also I tried with the latest build from github once and it crashed

Yimura:
Going to close this in favour of making a new issue if the bug is recurring.

PR #341, opened 2022-07-15
Fix an issue where spawn max could crash the game for some vehicles

aa15032261:
Fixed issue #253. Also moved some vehicle functions to more appropriate location.

### DIFF
--- a/BigBaseV2/src/gta/VehicleValues.h
+++ b/BigBaseV2/src/gta/VehicleValues.h
@@ -386,9 +386,44 @@ enum VehicleModType
 
 enum CustomVehicleModType
 {
-    MOD_WHEEL_TYPE = -101,
-    MOD_PLATE_STYLE = -102,
-    MOD_WINDOW_TINT = -103
+    MOD_MODEL_HASH        = -100,
+
+    MOD_WHEEL_TYPE        = -101,
+    MOD_PLATE_STYLE       = -102,
+    MOD_WINDOW_TINT       = -103,
+
+    MOD_FRONTWHEEL_VAR    = -104,
+    MOD_REARWHEEL_VAR     = -105,
+
+    MOD_NEON_FRONT_ON     = -106,
+    MOD_NEON_BACK_ON      = -107,
+    MOD_NEON_LEFT_ON      = -108,
+    MOD_NEON_RIGHT_ON     = -109,
+    MOD_TIRE_CAN_BURST    = -110,
+
+    MOD_PRIMARY_COL       = -111,
+    MOD_SECONDARY_COL     = -112,
+    MOD_PEARLESCENT_COL   = -113,
+    MOD_WHEEL_COL         = -114,
+    MOD_INTERIOR_COL      = -115,
+    MOD_DASHBOARD_COL     = -116,
+    MOD_XENON_COL         = -117,
+
+    MOD_PRIMARY_COL_R     = -118,
+    MOD_PRIMARY_COL_G     = -119,
+    MOD_PRIMARY_COL_B     = -120,
+
+    MOD_SECONDARY_COL_R   = -121,
+    MOD_SECONDARY_COL_G   = -122,
+    MOD_SECONDARY_COL_B   = -123,
+
+    MOD_TIRESMOKE_COL_R   = -124,
+    MOD_TIRESMOKE_COL_G   = -125,
+    MOD_TIRESMOKE_COL_B   = -126,
+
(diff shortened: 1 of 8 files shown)


BigBaseV2/src/views/vehicle/view_pv.cpp
```diff
@@ -136,19 +136,19 @@ namespace big
 									spawn_plate = personal_veh->get_plate();
 								}
 
-								auto veh = vehicle::clone(veh_data, spawn_location, spawn_heading);
-
-								if (g->clone_pv.spawn_inside)
-								{
-									vehicle::telport_into_veh(veh);
-								}
+								auto veh = vehicle::clone_from_vehiclie_data(veh_data, spawn_location, spawn_heading);
```
Yimura: Typo in function name.

BigBaseV2/src/gta/VehicleValues.h
```diff
@@ -386,9 +386,44 @@ enum VehicleModType
...
+    MOD_SECONDARY_COL_G   = -122,
+    MOD_SECONDARY_COL_B   = -123,
+
+    MOD_TIRESMOKE_COL_R   = -124,
+    MOD_TIRESMOKE_COL_G   = -125,
+    MOD_TIRESMOKE_COL_B   = -126,
+
+    MOD_NEON_COL_R        = -127,
+    MOD_NEON_COL_G        = -128,
+    MOD_NEON_COL_B        = -129,
```
Yimura, line 426: Why are you using negative values? Is it just to not have an index collision in your std::map?

I don't have a problem with it, just odd choice...

Couldn't you use an enum class otherwise and not use manually set enum values and let the compiler decide? It just seems unnecessary to do it like this.

Anyways I'd like to just know your opinion on this.

Yimura (changes requested):

BigBaseV2/src/gta/VehicleValues.h
aa15032261: > Why are you using negative values? Is it just to not have an index collision in your std::map?

Yes, I pick negative values to avoid collision. The positive side is used by VehicleModType so I picked the negative side.

> Couldn't you use an enum class otherwise and not use manually set enum values and let the compiler decide? It just seems unnecessary to do it like this.

I guess you are right. It is probably unnecessary and just my personal preference. This structure is actually a preparation for the clone/save current vehicle feature I'm going to implement. I usually set values manually if the structure is going to be serialized so it won't mess with the serialization if I need to update it.

BigBaseV2/src/views/vehicle/view_pv.cpp
aa15032261: Thanks for catching that!

BigBaseV2/src/gta/VehicleValues.h
Yimura: Alright, you can keep it as-is then.

mentolixite: @aa15032261 Please note that @GibsonHF's fork already has the feature.

aa15032261: > @aa15032261 Please note that @GibsonHF's fork already has the feature.

Thanks for letting me know! I'll have a look

BigBaseV2/src/views/vehicle/view_spawn.cpp
```diff
@@ -104,18 +104,18 @@ namespace big
...
 							if (g->spawn.spawn_maxed)
 							{
 								vehicle::max_vehicle(veh);
 							}
 
 							vehicle::set_plate(veh, plate);
 
+							if (g->spawn.spawn_inside)
+							{
+								vehicle::telport_into_veh(veh);
```
Yimura: Typo

BigBaseV2/src/views/vehicle/view_pv.cpp
```diff
@@ -136,19 +136,19 @@ namespace big
...
 								if (g->clone_pv.spawn_maxed)
 								{
 									vehicle::max_vehicle(veh);
 								}
 
 								vehicle::set_plate(veh, spawn_plate);
+
+								if (g->clone_pv.spawn_inside)
+								{
+									vehicle::telport_into_veh(veh);
```
Yimura: Typo

Yimura (changes requested):

Yimura (approved):

PR #342, opened 2022-07-15
MingW compatible build attempt

Yimura:
Source Code should be compatible but I'm having some issues when linking it.

Use the following the command to prep the build environment:
```
premake5 gmake2
```

This is the command I used to compile it:
```bash
x86_64-w64-mingw32-make config=release -I /usr/x86_64-w64-mingw32/include/ -L /usr/x86_64-w64-mingw32/lib/
```

Multicore compile is broken though 2 cores seems to work `-j2`.


mentolixite:
Source code changes are good, regardless of the compiler.

I personally consider `clang-cl` way better than MinGW.

Yimura:
Wouldn't `clang` also just work?

I've just removed all Microsoft only exstensions from the source and made it more ISO C++.
Aside from that I had to fix a bunch of files from the old BigBaseV2. 

Edit:
As long as you've got the Microsoft and dx headers it should just work.

tupoy-ya:
Managed to build it, but it doesn't start.

To build i changed links in premake to this
```lua
links
{
    "fmtlib",
    "g3log",
    "ImGui",
    "MinHook",
    "stdc++",
    "wsock32",
    "ws2_32",
    "imagehlp",
    "d3dcompiler",
    "dwmapi"
}
```
and modified my `MakeFile` like this, maybe it can be done with premake but i haven't tried that.
```diff
- ALL_LDFLAGS += $(LDFLAGS) -L../bin/lib -L/usr/lib64 -m64 -flto -shared -s
+ ALL_LDFLAGS += $(LDFLAGS) -L../bin/lib -L/usr/x86_64-w64-mingw32/lib -m64 -flto -shared -s -mwindows
---
OBJECTS += $(OBJDIR)/vmt_hook.o
OBJECTS += $(OBJDIR)/wndproc.o
+ OBJECTS += $(OBJDIR)/invoker_asm.o
```

Then i manually compiled assembly by running `uasm -win64 invoker_asm.asm`
and moved "invoker_asm.o" to "YimMenu/bin/int/Release/BigBaseV2/invoker_asm.o".

Yimura:
Closing this PR as it has gone stale, probably will not try to swap the current version of YimMenu to a different build environment.

Maybe when switching base to one that uses CMake we can further pursue this.

PR #345, opened 2022-07-16
Improve and optimise session local_time

howdoiusekeyboard:
Local time has a transition to revert back to global session time when used in online.
Credits to @xM4ddy

### DIFF
--- a/BigBaseV2/src/backend/looped/session/local_time.cpp
+++ b/BigBaseV2/src/backend/looped/session/local_time.cpp
@@ -3,22 +3,24 @@
 
 namespace big
 {
-	static bool lastOverride = true;
+	static bool toggled = true;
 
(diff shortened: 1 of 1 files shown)


tupoy-ya:
weird i don't recall changing that

GeopJr (dismissed):
This has been PRed before #130 

Please see Yimura's review: https://github.com/YimMenu/YimMenu/pull/130#issuecomment-1076642040

howdoiusekeyboard:
It works as intended now

howdoiusekeyboard:
1. In multiplayer, if you override time and then disable it, the change isn't sophisticated, this will smoothen that transition.
2. In story mode if you disable override it will apply server time from online which is not required, this fixes it .

Yimura (approved):

PR #352, opened 2022-07-19
feat(Nightly): Added sigscanner to nightly

Yimura:
This PR adds a sigscanner to the nightly build of YimMenu.

The nightly build has been modified to always build a DLL regardless of if there were any changes in the last 24h.
There are 2 parallel jobs that depend on the binary being built, one runs a sigscanner to see if anything is detected.
The other checks if there were any changes made to the source code in the last 24h and create a new release if so.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -79,7 +79,7 @@ jobs:
 
       - name: Test file for detected bits
         id: detections
-        uses: yimura/gtav-sigscan-action@v0.0.1
+        uses: yimura/gtav-sigscan-action@v0.0.2
         with:
           file: ./YimMenu.dll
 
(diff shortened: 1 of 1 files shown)


GeopJr (changes requested):
How about reversing them?
Build nightly only after sigscan ran and passed?

Yimura:
> How about reversing them? Build nightly only after sigscan ran and passed?

2 problems:
1. Running the sigscanner on the latest nightly might be dated
2. We'd be building to then run the sigscanner on the binary and then release it anyways.

So we need to always build if we need to release or check the sigscanner, so we split the workflow in always building it.
Then check if the repository has had changes in the last 24h, if so create a new release.
And check the sigscanner on that same build regardless of if we need to create a release.

GeopJr (dismissed):

Yimura:
Gonna implement that in this PR, don't merge.

Yimura:
Now it'll always build the DLL, check if there were any detections from the sigscanner and parallel from this will check if there's a need to create a new release.

GeopJr (approved):

PR #353, opened 2022-07-19
Drive on water feature now supports flying vehicles.

aa15032261:
Also fixed an issue that causes vehicles trapped in the invisible object.


Yimura (approved):

BigBaseV2/src/backend/looped/vehicle/drive_on_water.cpp
```diff
@@ -5,61 +5,89 @@
 
 namespace big
 {
+	const Hash drive_on_water_surface_hash = rage::joaat("stt_prop_stunt_bblock_xl3");
```
Yimura: This can be a constexpr auto instead of const.

```cpp
constexpr auto drive_on_water_surface_hash = RAGE_JOAAT("stt_prop_stunt_bblock_xl3");
```

The difference being that constexpr is a compile time constant, meaning that the code is ran once when the program is being compiled, while const doesn't have such a restriction.
In this case the string of `stt_prop_stunt_bblock_xl3` would probably be compiled into the DLL instead of having the hash only be compiled into the DLL.

Yimura (changes requested):

Yimura (approved):

PR #365, opened 2022-07-28
Added weapons to gta_data_service.

aa15032261:


Yimura (dismissed):
Looks good, haven't tested yet.

karifeld (commented):
works good, one thing. it looks like youve added the `WEAPON_TACTICALRIFLE` (service carbine) to the `hash_blacklist` array by mistake.

that gun is unlockable in-game by normal means, i dont know if this is intentional or not

aa15032261:
> works good, one thing. it looks like youve added the `WEAPON_TACTICALRIFLE` (service carbine) to the `hash_blacklist` array by mistake.
> 
> that gun is unlockable in-game by normal means, i dont know if this is intentional or not

It was a mistake. Thanks for spotting that! It is fixed in the latest commit.

Yimura:
@karifeld thanks for testing it.
@aa15032261 I'll merge it then after I've had one last look.

BigBaseV2/src/services/gta_data/gta_data_service.cpp
```diff
@@ -85,27 +93,61 @@ namespace big
...
 
-	void gta_data_service::load_from_file(file file_to_load, file file_etag, std::string url, void(gta_data_service::* load_func)(), std::string data_name)
+		if (idx == -1)
+		{
+			return empty_weapon_item;
+		}
+		else
+		{
+			return m_weapon_item_arr[idx];
+		}
```
Yimura: redundant else as you've already returned (exit) out of the function in the if statement.

Instead you can just return at the end.

BigBaseV2/src/services/gta_data/weapon_item.cpp
```diff
@@ -0,0 +1,78 @@
...
+		}
+
+		if (item_json["Category"].is_null())
+		{
+			this->weapon_type = "NULL";
+		}
+		else
+		{
+			this->weapon_type = item_json["Category"];
+		}
```
Yimura: You can probably default the this->weapon_type = "NULL" and if the Category isn't null assign it.
```cpp
this->weapon_type = "NULL";
if (!item_json["Category"].is_null())
{
	this->weapon_type = item_json["Category"];
}
```

Yimura (changes requested):

Yimura (approved):

PR #375, opened 2022-07-29
Added ped model preview.

aa15032261:


Yimura (dismissed):

Yimura:
I crashed once after injecting it, haven't been able to recreate the crash yet.

Also the alpha fade-in is really slow, can you please make it faster again.
The game can only increment the alpha in really big numbers (20% or smth), so fading in slow is just jarring to look at.

aa15032261:
> I crashed once after injecting it, haven't been able to recreate the crash yet.
> 
> Also the alpha fade-in is really slow, can you please make it faster again. The game can only increment the alpha in really big numbers (20% or smth), so fading in slow is just jarring to look at.

Looks like SET_ENTITY_ALPHA doesn't work well with ped. Some clothes don't show when the ped is transparent. I still doubled the speed anyway and hopefully it looks better for you.

Yimura (approved):

PR #377, opened 2022-07-31
Added spawn ped feature.

aa15032261:
image


BigBaseV2/src/services/gui/gui_service.hpp
```diff
@@ -59,11 +67,14 @@ namespace big
 					{ tabs::HANDLING_SEARCH, {"Search Handling", view::handling_search } },
 				}}},
 				{ tabs::LSC, { "LS Customs", view::lsc }},
-				{ tabs::SPAWN, { "Spawn", view::spawn }},
+				{ tabs::SPAWN_VEHICLE, { "Spawn Vehicle", view::spawn_vehicle }},
 				{ tabs::PV, { "Personal Vehicle", view::pv }},
-				{ tabs::VEHICLE_FUN, { "Fun Features", view::vehicle_fun }},
 				{ tabs::PERSIST_CAR, { "Persist Car", view::persist_car }},
 			}}},
+			{ tabs::FUN, { "Fun Features", nullptr, {
```
maybegreat48: I think all the vehicle stuff should stay in Vehicle and the ped spawner can go in a new category called "World" or something like that

Yimura: Yes that does sound better, we can probably move a bunch of things to the world category.

aa15032261: Yea, that sounds great. I will change it real quick.

aa15032261: This is the updated layout:

image

BigBaseV2/src/services/gui/gui_service.hpp
```diff
@@ -91,7 +102,7 @@ namespace big
 		void set_nav_size(int);
 		void increment_nav_size();
 		void reset_nav_size();
-		std::unordered_map<tabs, navigation_struct> get_navigation();
+		std::map<tabs, navigation_struct> get_navigation();
```
maybegreat48: Can't we just return a reference? This is being called every tick

xiaoxiao921: And why the change to `map` ?

maybegreat48: Probably because iterating an unordered_map returns keys in undefined order

xiaoxiao921: Yes I know that but I simply wanted the reasoning behind the change.
Like what bug that change exactly fixed.

aa15032261: > Can't we just return a reference? This is being called every tick

Yes, I'll update that as well. I only changed unordered_map to map and didn't pay attention that it was return by value.

aa15032261: > Yes I know that but I simply wanted the reasoning behind the change. Like what bug that change exactly fixed.

Map allows us to control the order of the items in the menu, so the order stays the same even if we updated it.

aa15032261: get_selected_tab and get_navigation are now return by reference.

Yimura (dismissed):

howdoiusekeyboard:
Add an option to spawn without weapons?

aa15032261:
> Add an option to spawn without weapons?

Thanks for the suggestion! It’s been added in the latest commit.

Yimura (dismissed):

Yimura (approved):

PR #399, opened 2022-08-08
Improved detour handler, cleaned up hooks, and improved RCS hook

Vali0004:

PR #402, opened 2022-08-09
Patch: Is DLC Present

BugisoftRSG:
Is DLC Present Hook is initialized however never enabled/disabled


Yimura:
Already has been fixed in #399

PR #407, opened 2022-08-12
feat(RCS): Improved crash protections.

SkiddyToast:
Improves crash protections by checking data nodes for invalid data. The hashes in the cases can be generated by logging the `typeid` hashes and names.
Here's an example of all the hashes for the `CPlayerSyncTree`:
```txt
class CPlayerCreationDataNode: 0x76EF5527D75AD95D
class CGlobalFlagsDataNode: 0x6F7A9A62E2CBE35E
class CDynamicEntityGameStateDataNode: 0x30B991CE74E9A395
class CPhysicalGameStateDataNode: 0x47F43A8CFB6DDE4A
class CPedGameStateDataNode: 0x47BFD0C8D62C7F36
class CPedComponentReservationDataNode: 0x5D857E47FBF98CDB
class CEntityScriptGameStateDataNode: 0xD1E8079A0DB4A927
class CPlayerGameStateDataNode: 0xBF61B399A94E3D9B
class CPedAttachDataNode: 0xD3A0FBD0D1CFD23A
class CPedHealthDataNode: 0x4C41EF1C4CD4C639
class CPedMovementGroupDataNode: 0x83991F92B55E146F
class CPedAIDataNode: 0x7873EBF446254479
class CPlayerAppearanceDataNode: 0x2AEE7E3A2D2E84FF
class CPlayerPedGroupDataNode: 0x58C4254AD1421D7D
class CPlayerAmbientModelStreamingNode: 0xEE7A23F35C2CFFBD
class CPlayerGamerDataNode: 0x4FA9FE4133930B34
class CPlayerExtendedGameStateNode: 0x26E7730078ECF329
class CPedOrientationDataNode: 0xB52B7350012734C7
class CPedMovementDataNode: 0xEA5A5D9278B58744
class CPedTaskTreeDataNode: 0x7A7092123DD07228
class CPedTaskSpecificDataNode: 0xF971ADC568A24F53
class CPedTaskSpecificDataNode: 0xF971ADC568A24F53
class CPedTaskSpecificDataNode: 0xF971ADC568A24F53
class CPedTaskSpecificDataNode: 0xF971ADC568A24F53
class CPedTaskSpecificDataNode: 0xF971ADC568A24F53
class CPedTaskSpecificDataNode: 0xF971ADC568A24F53
class CPedTaskSpecificDataNode: 0xF971ADC568A24F53
class CPedTaskSpecificDataNode: 0xF971ADC568A24F53
class CSectorDataNode: 0x38335CCD1B62BF09
class CPlayerSectorPosNode: 0x5BE4BC10E1949905
class CPlayerCameraDataNode: 0x8630B8891E8F99ED
class CPlayerWantedAndLOSDataNode: 0x69C9A68C2078EA7D
class CMigrationDataNode: 0x538FD8BC57C8632E
class CPhysicalMigrationDataNode: 0xF1317AB524509688
class CPhysicalScriptMigrationDataNode: 0xA9E55C8D2B9E1759
```


pelecanidae:
BTW, to generate the same hashes MSVC uses for RTTI type IDs: https://godbolt.org/z/4jr6f6M56

```c++
constexpr uint64_t operator ""_fnv1a(char const *str, std::size_t len) {
  auto const fnv_offset_basis = 14695981039346656037ULL;
  auto const fnv_prime = 1099511628211ULL;

  auto value = fnv_offset_basis;
  for (auto i = 0; i < len; i++) {
    value ^= static_cast<size_t>(str[i]);
    value *= fnv_prime;
  }
  value ^= value >> 32;

  return value;
}

// ...

case "?AVCPlayerCreationDataNode@@"_fnv1a:
```

It uses the MSVC mangled symbol name, which follows the pattern "`?AV` + class name + `@@`" for trivial structs like the sync tree nodes.

Yimura:
@pelecanidae thanks a lot, that's an amazing addition.

Adler-D3R (dismissed):
Seems all good to me. Also, I trust Skiddy on his work so, even if I don't perfectly understand every single part of the code, it should be fine. I heard there were some false positives, patching this would be perfect. I didn't try it yet, but I should definitely, cuase this seem awesome.

Clone69420:
> CPlayerCreationDataNode

Nice Done, but i tested the Crash Models / Cage Models Protection and it didnt Blocked it

maybegreat48:
Hooking `read_bitbuffer_into_sync_tree` and filtering from there might be better as it isn't being called twice (one by the sync hook and one by the original) and also blocks crashes in takeover syncs (`GIVE_CONTROL_EVENT`). Having a separate blacklist for ped, vehicle and object improves performance. I don't have access to my old hard drive containing IDA for now so cannot really tell for sure

SkiddyToast:
> Hooking `read_bitbuffer_into_sync_tree` and filtering from there might be better as it isn't being called twice (one by the sync hook and one by the original) and also blocks crashes in takeover syncs (`GIVE_CONTROL_EVENT`). Having a separate blacklist for ped, vehicle and object improves performance. I don't have access to my old hard drive containing IDA for now so cannot really tell for sure

Hooking `read_bitbuffer_into_sync_tree` would be helpful, although I currently don't have a way of getting the sync sender from there. Will look into it later today.

Vali0004:
> > Hooking `read_bitbuffer_into_sync_tree` and filtering from there might be better as it isn't being called twice (one by the sync hook and one by the original) and also blocks crashes in takeover syncs (`GIVE_CONTROL_EVENT`). Having a separate blacklist for ped, vehicle and object improves performance. I don't have access to my old hard drive containing IDA for now so cannot really tell for sure
> 
> Hooking `read_bitbuffer_into_sync_tree` would be helpful, although I currently don't have a way of getting the sync sender from there. Will look into it later today.

Most of them can use net objects or something that inherits a class with their id. That can be used with the CNetworkPlayerMgr

Yimura:
I fixed the merge-conflicts, didn't bother to check if the code was still up-to-date with our current dependencies.

Vali0004:
> BTW, to generate the same hashes MSVC uses for RTTI type IDs: https://godbolt.org/z/4jr6f6M56
> 
> ```c++
> constexpr uint64_t operator ""_fnv1a(char const *str, std::size_t len) {
>   auto const fnv_offset_basis = 14695981039346656037ULL;
>   auto const fnv_prime = 1099511628211ULL;
> 
>   auto value = fnv_offset_basis;
>   for (auto i = 0; i < len; i++) {
>     value ^= static_cast<size_t>(str[i]);
>     value *= fnv_prime;
>   }
>   value ^= value >> 32;
> 
>   return value;
> }
> 
> // ...
> 
> case "?AVCPlayerCreationDataNode@@"_fnv1a:
> ```
> 
> It uses the MSVC mangled symbol name, which follows the pattern "`?AV` + class name + `@@`" for trivial structs like the sync tree nodes.

```cpp
inline consteval uint64_t generateMsvcRTTI(std::string_view str) {
	auto const fnvOffsetBasis = 14695981039346656037ULL;
	auto const fnvPrime = 1099511628211ULL;
	auto basisCopy = fnvOffsetBasis;
	for (size_t i{}; i != (sizeof(str) - 1); ++i) {
		basisCopy ^= static_cast<size_t>(str[i]);
		basisCopy *= fnvPrime;
	}
	basisCopy ^= basisCopy >> 32;
	return basisCopy;
}
inline consteval uint64_t operator ""_generateMangledRTTI(char const* str, size_t) { return generateMsvcRTTI(str); }
```

BigBaseV2/src/hooks/protections/received_clone_sync.cpp
```diff
@@ -20,21 +35,317 @@ namespace big
...
+		{
+			for (auto child = node->m_first_child; child; child = child->m_next_sibling) {
+				CPhysicalAttachDataNode* attach_node = get_attach_node_from_object(child);
+				if(attach_node != nullptr)
+					return attach_node;
+			}
+		}
+		else if (node->IsDataNode())
+		{
+			if (typeid(*node).hash_code() == 0xE6B80DD4C167CFF) //CPhysicalAttachDataNode
```
Yimura: Make use of
```cpp
"?AVCPhysicalAttachDataNode@@"_fnv1a
```

BigBaseV2/src/hooks/protections/received_clone_sync.cpp
```diff
@@ -20,21 +35,317 @@ namespace big
...
+		{
+			for (auto child = node->m_first_child; child; child = child->m_next_sibling) {
+				CPedAttachDataNode* attach_node = get_ped_attach_node_from_object(child);
+				if (attach_node != nullptr)
+					return attach_node;
+			}
+		}
+		else if (node->IsDataNode())
+		{
+			if (typeid(*node).hash_code() == 0xD3A0FBD0D1CFD23A) //CPedAttachDataNode
```
Yimura: Make use of
```cpp
"?AVCPedAttachDataNode@@"_fnv1a
```

BigBaseV2/src/hooks/protections/received_clone_sync.cpp
```diff
@@ -20,21 +35,317 @@ namespace big
...
+
+	inline bool is_model_a_cage_model(uint32_t model) {
+		for (int i = 0; i < sizeof(cage_models) / sizeof(uint32_t); i++)
+		{
+			if (cage_models[i] == model) return true;
+		}
+		return false;
+	}
 	
+	CPhysicalAttachDataNode* get_attach_node_from_object(rage::netSyncNodeBase* node)
```
Yimura: Possibly offer a function that replaces `get_attach_node_from_object` and `get_ped_attach_node_from_object`, though I'm not familiar with these protections so it might not be possible to implement such a function.
```cpp
template<typename T>
T* get_data_node(rage::netSyncNodeBase* node, uint64_t hash_code)
{
	if (node->IsParentNode())
	{
		for (auto child = node->m_first_child; child; child = child->m_next_sibling)
		{
			T* data_node = get_data_node<T>(child);
			if (data_node != nullptr)
				return data_node ;
		}
	}
	else if (node->IsDataNode())
	{
		if (typeid(*node).hash_code() == hash_code) //CPedAttachDataNode
			return reinterpret_cast<T*>(node);
	}
	return nullptr;
}
```

usage
```
const auto ped_attach_data_node = get_data_node<CPedAttachDataNode>(node, "?AVCPedAttachDataNode@@"_fnv1a);
```

BigBaseV2/src/hooks/protections/received_clone_sync.cpp
```diff
@@ -20,21 +35,317 @@ namespace big
...
+					{
+						if (g->notifications.invalid_sync.log)
+							LOG(WARNING) << "Invalid sector position." << " From : " << sender->get_name();
+						if (g->notifications.invalid_sync.notify)
+							g_notification_service->push_warning(fmt::format("Invalid sector position from {}", sender->get_name()), "Invalid sector position.");
+						return true;
+					}
+					break;
+				}
+			}
```
Yimura: This is already become a lot of small checks, though it might be a hassle splitting each and every check into smaller files.

Maybe we could switch this over to an std::unordered_map where `typeid(*node).hash_code()` is the key and the value is a function which checks the data node.

BigBaseV2/src/hooks/protections/received_clone_sync.cpp
```diff
@@ -20,21 +35,317 @@ namespace big
...
+	}
+
+	bool check_node(rage::netSyncNodeBase* node, CNetGamePlayer* sender, uint16_t object_id)
+	{
+		if (node->IsParentNode())
+		{
+			bool block = false;
+			for (auto child = node->m_first_child; child; child = child->m_next_sibling) {
+				if (check_node(child, sender, object_id))
+					block = true;
```
Yimura: return true in the for loop immediately? I'm not familiar but it seems like a waste of checking the other nodes if we're going to block it anyways.

Yimura (changes requested):

Aboba7778:
I tested this code on cage spawn, but cage successfully spawned and notify triggered after 3 seconds. What could be the problem?

SkiddyToast:
> I tested this code on cage spawn, but cage successfully spawned and notify triggered after 3 seconds. What could be the problem?

This is only on Received Clone Sync ATM, later this can be implemented into Received Clone Create.

BigBaseV2/src/hooks/protections/received_clone_sync.cpp
SkiddyToast: Nah

Yimura (approved):
This is safe to merge into master.

Merging it into master will give more people access to experiment with the protection possibilities this hook offers.

PR #409, opened 2022-08-12
add SC TV SessionType

Aure7138:


Aure7138:
Something went wrong, please don't merge, I should have only changed two files

PR #411, opened 2022-08-13
Improved error handling of gta_data_service.

aa15032261:
Fixed an issue where the game crashes when throwing invalid json format exception.

### DIFF
--- a/BigBaseV2/src/services/gta_data/gta_data_service.cpp
+++ b/BigBaseV2/src/services/gta_data/gta_data_service.cpp
@@ -1,4 +1,3 @@
-#include <filesystem>
 #include "api/remote.hpp"
 #include "file_manager.hpp"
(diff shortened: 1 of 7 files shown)


BigBaseV2/src/services/gta_data/gta_data_service.cpp
```diff
@@ -1,3 +1,4 @@
+#include <filesystem>
```
Yimura: Isn't this already included by common.hpp?

Yimura (changes requested):

BigBaseV2/src/services/gta_data/gta_data_service.cpp
aa15032261: Oh my bad, i didnt check.

Yimura (approved):

PR #423, opened 2022-08-22
feat(util): Added model hash table helper

Yimura:
Name of the helper is up for debate, same for the functions.

Point of this class is to provide a wrapper that doesn't require you to use or wait for natives like IS_MODEL_VALID.


BigBaseV2/src/util/models.hpp
```diff
@@ -0,0 +1,64 @@
...
+			if (const auto model = model_info::get_model(hash); model)
+				return true;
+			return false;
+		}
+
+		template<typename T = CBaseModelInfo*>
+		static T get_model(const rage::joaat_t hash)
+		{
+			const auto model_table = g_pointers->m_model_table;
+			for (auto i = m_model_table->m_lookup_table[adder_hash % m_model_table->m_lookup_key]; i; i = i->m_next)
```
Yimura: Still using `adder_hash` instead of parameter `hash`.

BigBaseV2/src/util/models.hpp
```diff
@@ -0,0 +1,64 @@
...
+			{
+				if (i->m_hash == hash)
+				{
+					if (const auto model = m_model_table->m_data[i->m_idx]; model)
+					{
+						return model;
```
Yimura: `dynamic_cast<T>(model)` or `reinterpret_cast<T>(model)`

BigBaseV2/src/util/models.hpp
```diff
@@ -0,0 +1,64 @@
...
+					}
+				}
+			}
+			return nullptr;
+		}
+
+		static CVehicleModelInfo* get_vehicle_model(const rage::joaat_t hash)
+		{
+			if (const auto model = model_info::get_model<CVehicleModelInfo*>(hash); model && model->m_model_type == eModelType::Vehicle)
```
Yimura: There are sub-types of Vehicle that need to be checked as well.

Yimura (commented):
Taking some notes so I don't forget.

BigBaseV2/src/util/models.hpp
Vali0004: I sent all of these fixes to Skiddy Toast, it removed the require for that native. I asked him to add it into his PR because I didn’t want to make it hard for anyone. Except he never finished it. I went ahead and fixed other things.

Vali0004: At least, inside the RCS hook there was some issues I overlooked.

Yimura: If I understand what you're saying I should finish this work but not implement the fix with the native yet.

And wait for SkiddyToast to fix that?

SkiddyToast (approved):
Looks good to me.

PR #428, opened 2022-08-29
Improve protections

maybegreat48:
Added protections for the OOM kick and a few TSEs.
Also why do we have a toggle for crash protections anyway? Nobody in their right mind will even think of disabling them.


SkiddyToast (approved):
LGTM

Yimura (approved):
Thanks for your contribution, all looks good.

PR #431, opened 2022-08-29
feat(Ped): Play Animation

### LINKED ISSUE #187: 💡 [Animations]
👎

   [Self]
         ↳ [Animations]

**Provide Reasoning**
They're always nice to have, for various reasons, and it would be safer and easier than loading an outdated SP ASI.

**Additional context**
_e.g. : Add a premade list, and/or "Play by name" + "Save & Load / Favorite" buttons._

Aure7138:
[Imgur](https://imgur.com/aoHEc4c)


Aure7138:
There are too many animations in animDictsCompact.json and it will freeze the imgui interface

Yimura:
You can try using the thread pool to handle parsing of the JSON file.

Aure7138:
Not at load time, but at display time, I used data/animations.hpp instead

Aure7138:
If all the animations in json are displayed, just display, without filtering, the interface is already very stuck

PR #437, opened 2022-09-01
Fix drawing self on ESP

kalkafox:
It is my understanding that `g_player_service->players()` does not include the local player. Because of this, the ESP was unable to draw despite the option toggled on or off.

My solution to the issue was to relocate the contents within the original loop into a new `draw_player` method that accepts a `player_ptr` reference and the pointer to `draw_list` respectively. This is then called for ourselves using `get_self()` and then called for every player in the session.

I should add that I am still very new to the language, so please forgive any crude habits and feel free to contribute as needed.

Take care.

EDIT: (Drafting to fix up the commit log a bit.)
EDIT2: `plyr->is_valid()` is now called during the local player check to fix a crash.


Yimura (approved):
Code looks good, still need to test it.

Yimura:
Draw self seems to work though people need to keep in mind that their global render distance needs to be set to 0 for it to work.

image

PR #451, opened 2022-09-19
Add world model spawning

xiaoxiao921:
Testing the feature:
Add in `BigBaseV2\saved_json_vehicles`
Change extension to json
[maze_bank_test.txt](https://github.com/YimMenu/YimMenu/files/9600761/maze_bank_test.txt)

### DIFF
--- a/BigBaseV2/src/pointers.cpp
+++ b/BigBaseV2/src/pointers.cpp
@@ -92,8 +92,14 @@ namespace big
 			m_model_spawn_bypass = ptr.add(8).as<PVOID>();
 		});
 
+		// World Model Spawn Bypass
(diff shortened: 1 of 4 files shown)


SkiddyToast:
IDK if it's needed, but you should also make sure that the original model bypass is being used when spawning the model, just in case.

xiaoxiao921:
I thought about it but it seems we only use the bypass for spawning peds (ped.hpp), and considering model attachment on the json car spawner never used the bypass too I don't think its needed ? I could be wrong

SkiddyToast:
I guess that is the case, I'll go ahead and do a review.

SkiddyToast (approved):
LGTM

PR #457, opened 2022-09-27
vehicles, peds, weapons local cache

xiaoxiao921:
Create a local cache once per game update.
The only issue currently is that if the user inject in SP and go to MP a crash will trigger because of some rpf file that got mounted twice is needed for the loading, its pretty meh but i'm not sure how to fix it, the fidevice system is a huge can of worm in itself and i'm out of energy for this right now. Atleast this crash will trigger "only" once per update because once the cache is done for the current update it just query the json files from disk


Yimura:
I'm sure if we'd look at how to fix that crash we can figure it out.

BigBaseV2/src/services/gta_data/gta_data_service.cpp
```diff
@@ -135,81 +129,453 @@ namespace big
...
+								veh["WheelType"] = wheel_type.substr(4);
+							else
+								veh["WheelType"] = nullptr;
+
+							const std::string flags_str = item.child("flags").text().as_string();
+							veh["Flags"] = split_str_into_json_array(flags_str, " ");
+
+							veh["Seats"] = VEHICLE::GET_VEHICLE_MODEL_NUMBER_OF_SEATS(model_hash);
+
+							veh["Price"] = -1; // todo: how DurtyFree generate it?
```
Yimura: CHandlingData has this information I believe.

https://github.com/Yimura/GTAV-Classes/blob/1da236c3c6ca5586dc9e1069dde1fbfd49641882/vehicle/CHandlingData.hpp#L74

BigBaseV2/src/services/gta_data/gta_data_service.cpp
```diff
@@ -135,81 +129,453 @@ namespace big
...
+								veh["ManufacturerDisplayName"] = nullptr;
+							}
+
+							const auto vehicle_class = std::string(item.child("vehicleClass").text().as_string());
+							if (vehicle_class.size() > 3)
+								veh["Class"] = vehicle_class.substr(3);
+							else
+								veh["Class"] = nullptr;
+
+							veh["ClassId"] = 0; // todo: how DurtyFree generate it?
```
Yimura: CVehicleModelInfo probably.

https://github.com/Yimura/GTAV-Classes/blob/1da236c3c6ca5586dc9e1069dde1fbfd49641882/vehicle/CVehicleModelInfo.hpp#L48

BigBaseV2/src/services/gta_data/gta_data_service.cpp
```diff
@@ -135,81 +129,453 @@ namespace big
...
+							}
+							veh_hashes[model_hash] = {};
+
+							veh["Hash"] = model_hash;
+							veh["SignedHash"] = (int32_t)model_hash;
+							std::stringstream s;
+							s << HEX_TO_UPPER(model_hash);
+							veh["HexHash"] = s.str();
+
+							veh["DlcName"] = "TODO"; // todo: how DurtyFree generate it?
```
Yimura: Unsure I'd like to know as well.

BigBaseV2/src/services/gta_data/gta_data_service.cpp
```diff
@@ -135,81 +129,453 @@ namespace big
...
+							ped["ClipDictionaryName"] = item.child("ClipDictionaryName").text().as_string();
+							ped["BlendShapeFileName"] = item.child("BlendShapeFileName").text().as_string();
+							ped["ExpressionSetName"] = item.child("ExpressionSetName").text().as_string();
+							ped["ExpressionDictionaryName"] = item.child("ExpressionDictionaryName").text().as_string();
+							ped["ExpressionName"] = item.child("ExpressionName").text().as_string();
+
+							ped["Pedtype"] = item.child("Pedtype").text().as_string();
+
+							ped["MovementClipSet"] = item.child("MovementClipSet").text().as_string();
+							ped["MovementClipSets"] = nullptr; // todo
```
Yimura: To Do.

Yimura (changes requested):
Looks good, making these comments just to keep track of them so don't worry about the TODO comment.

I will jump into the PR myself as well to take a look at what I can do.

SkiddyToast:
I'd also recommend moving the `fiDevice` and `GXT2` related classes into the GTAV-Classes repo.

Yimura:
Yeah, right now for simplicity of implementing it I'm going to merge the classes into the repo and at later moment we can move them to the other repo.

maybegreat48:
Haven't tested this but loading MP DLC through `DLC::ON_ENTER_MP` beforehand may stop the crash

Yimura:
Will test that afterwards. Right now I'm busy revamping the whole GTA Data service class. Throwing away the need for JSON files and instead store it binary.

I'm so glad this thing works though as is.

Yimura:
99% finished

Here and there can be optimised some more where we're copying vectors over where we might get away by moving the memory.
I crashed while previewing the vehicles, didn't check which one I crashed one (probably next-gen that we're not filtering out).

Current structure uses the following binary header format:
|Order|Field Name|Field Size (bytes)|Type|
|--|--|--|--|
|1|cache version|4|`uint32`|
|2|game build|4|`uint32`|
|3|online version|4|`float`|
|4|data size|8|`uint64`|
|5|data| / | `any` |

Yimura:
Seems the next-gen vehicles are causing crashes again; I have no idea from where I need to parse the bones of a vehicle model.

These are the current methods of checking for invalid vehicles, I checked handling.meta for monetaryValue but that one never seemed to be 0:
https://github.com/YimMenu/YimMenu/blob/d01afda8d360baa527099d8952eb745cf75d5dee/BigBaseV2/src/services/gta_data/gta_data_service.cpp#L238-L240

Edit:
I've found the arbitergt.ytf file which contains a stub bone it seems.
image

Yimura:
Should be ready to merge, @xiaoxiao921 tell me if you see anything.

GeopJr:
Could you bump pugixml to HEAD (`86c91051541d18a0d24b837a866cf0306fc8db1a`)? (it had some bug fixes since `521b2cd`)

Yimura (approved):

Yimura:
Going to merge soon, I want to retire the proxy server in the next 2 weeks.

PR #458, opened 2022-09-28
fix(Spectate & FreeCam): Dying while going far away from our ped

Yimura:
I think this fixes all occurrences of spectating and dying while going far away from your current active ped.


Yimura:
Should be ready for review.

Yimura:
Nvm, forgot about noclip.

Yimura:
Doesn't seem to work in combination with ENTITY::FREEZE_ENTITY_POSITION

PR #486, opened 2022-10-20
Fix: Properly restore the byte patches when unloading

xiaoxiao921:


BigBaseV2/src/memory/byte_patch.hpp
```diff
@@ -2,31 +2,64 @@
...
+		/// <summary>
+		/// Does not guarantee restoration of bytes, only after the object is deconstructed will the bytes be restored.
+		/// </summary>
+		void restore() const
+		{
+			if (const auto it = std::find(m_patches.begin(), m_patches.end(), this); it != m_patches.end())
+			{
+				m_patches.erase(it);
+			}
+		}
```
Yimura: I can probably call `it->reset()` on the iterator to guarantee the patch is restored.

BigBaseV2/src/memory/byte_patch.hpp
```diff
@@ -0,0 +1,65 @@
+#pragma once
+
+namespace memory
+{
+	class byte_patch
+	{
+	public:
+		template <typename TAddr>
+		byte_patch(TAddr address, std::remove_pointer_t<std::remove_reference_t<TAddr>> value)
```
xiaoxiao921: Ctor should prolly be private? Otherwise user may create an instance and it won't be added to the vector

BigBaseV2/src/memory/byte_patch.hpp
```diff
@@ -0,0 +1,65 @@
...
+			: m_address(address)
+		{
+			constexpr auto size = sizeof(std::remove_pointer_t<std::remove_reference_t<TAddr>>);
+			m_original_bytes.resize(size);
+			memcpy(m_original_bytes.data(), m_address, size);
+
+			*address = value;
+		}
+
+		virtual ~byte_patch()
```
xiaoxiao921, line 8: Why virtual?

BigBaseV2/src/memory/byte_patch.hpp
Yimura: Yeah sounds good.

BigBaseV2/src/memory/byte_patch.hpp
Yimura: If someone were to inherit the class this ensures the order of destructors called is respected.

We can probably use the `final` keyword on the class to prevent inheritance.
It's also why I've made the vector of byte_patch protected instead of private for possible future inheritance.

BigBaseV2/src/memory/byte_patch.hpp
Yimura: Done ✔️

BigBaseV2/src/memory/byte_patch.hpp
Yimura: Done ✔️

Yimura (approved):
Seems to work as intended without crashing.

PR #488, opened 2022-10-20
Feat scripting wren

xiaoxiao921:


GeopJr:
Is there any way to avoid hardcoding this path? https://github.com/YimMenu/wren/commit/d90a80c3ecab425a2e177425146df3fa154d90d6#diff-a7d1696a72e749a63aa7e1debc44c76929e014cd2617bd74d2cb98d0f086f261R201

It would require all downstreams that have changed its name from `BigBaseV2` to also fork (& maintain) YimMenu/wren.

GeopJr:
So, I fixed some of the tests on wren (idk why their Windows action does not run tests), however there are 2 remaining that should be addressed:

1. `test/regression/494.wren`:

```
Missing expected error on line 1.
```
I believe that this is an important one as it should definitely error, right?:
```
0[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
// expect error line 1
```
[more info](https://github.com/wren-lang/wren/commit/8fae8e4f1e490888e2cc9b2ea6b8e0d0ff9dd60f)

2. `test/language/bom.wren`

```
Unexpected output on stderr:
      free(): invalid pointer
      Expected return code 0 and got -6. Stderr:
      free(): invalid pointer
      
      Missing expected output "ok" on line 2.
```

xiaoxiao921:
Not sure why both of these fails, first test could be related to the max param change, second test could be related to the `syntaxSugar` stuff I added

thanks for fixing the tests btw, also Fn.new still has the limit of 16 max params, it probably should be noted somewhere, you can see why in wren_core.c `FUNCTION_CALL(vm->fnClass, "call(_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_)", fn_call16);`
I wasnt sure how to not having to generate each of the call so I simply left it as is as the whole reason I augmented the limit was for native foreign method interop

xiaoxiao921:
its also worthy to note that currently there is an issue where emitted line errors can be not matching with what the user have when they use the list deconstruct syntax sugar, implementing on the parser/lexer level is much much more annoying due to no AST building

xiaoxiao921:
If you want to play around with that branch, put those two files in appdata/bigbasev2/scripts/wren/
[wren_scripts.zip](https://github.com/YimMenu/YimMenu/files/9844905/wren_scripts.zip)

GeopJr:
Ok, so I fixed both failing tests (on the first one, the expected line had changed, on the second one, we have to remove the BOM bytes before consuming the source),

but apparently some new ones related to ASSERT popped up (https://github.com/YimMenu/wren/actions/runs/3304535591/jobs/5454028410) (that didn't even run previously due to the other 2 failing tests), which I'll leave up to you and Yimura to deal with.

Please do bump wren to the HEAD (`5c6f2145f030eb5fda0e9d09e2bbd35276669207`).

xiaoxiao921:
ASSERT change i'm just gonna simply revert it, it was just a dirty for me of debugging wren internals while having it injected in GTA

tupoy-ya:
Accidentally pushed here instead of my fork

Yimura:
No problem, just seems like all you did was a merge.

yasmasdas:
Any progress planned?

Issue #440, opened 2022-09-05
[Request]: I would like to make a proposal for the repair of the vehicle


Jo3y4r3aL:
### Problem

Can u turn the Repair Vehicle button into an automatic repair?

### Solution

Just add this feature to the vehicle list

### Reason

This can give people like me who have an obsessive-compulsive disorder a good experience
It can also help players better appreciate their cars

TheGreenBandit:
sechsmenu includes this feature in their fork

PR #493, opened 2022-10-21
feat: vehicle auto-repair

GeopJr:
Upstreamed from https://github.com/howdoiusekeyboard/SechsMenu (Thanks!)

closes: #440

### DIFF
--- a/BigBaseV2/src/backend/looped/vehicle/keep_vehicle_repaired.cpp
+++ b/BigBaseV2/src/backend/looped/vehicle/keep_vehicle_repaired.cpp
@@ -1,0 +1,13 @@
+#include "backend/looped/looped.hpp"
+#include "util/vehicle.hpp"
+
+namespace big
+{
+	void looped::vehicle_keep_vehicle_repaired()
+	{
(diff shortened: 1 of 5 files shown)


xiaoxiao921 (dismissed):

BigBaseV2/src/backend/looped/vehicle/keep_vehicle_repaired.cpp
```diff
@@ -0,0 +1,13 @@
+#include "backend/looped/looped.hpp"
+#include "util/vehicle.hpp"
+
+namespace big
+{
+	void looped::vehicle_keep_vehicle_repaired()
+	{
+		bool toggle = VEHICLE::GET_DOES_VEHICLE_HAVE_DAMAGE_DECALS(self::veh);
+		if (g->vehicle.keep_vehicle_repaired && toggle) {
```
Yimura: Can this native call be inlined in the if statement? That way we wouldn't be making unnecessary native calls if the keep_vehicle_repaired isn't set to true.

Yimura (changes requested):

Yimura (approved):

PR #502, opened 2022-10-22
feat(Spoofing): Add local user name spoofing

### LINKED ISSUE #417: [Request]: A fix for name spoofing to work locally
### Problem

When spoofing the username, it doesn't spoof it locally. When you type in chat, it shows the spoofed username for everyone else. But for you, it still shows your R* username.

### Solution

Essentially would just love to see it spoofed everywhere.

### Reason

At times, I can't really tell if has worked or not. I would have to have someone read the username off to me in-game, or a friend I am playing with has to tell me. I can't tell as its not spoofed for myself.

### Additional context

_No response_

Yimura:
### Changes
* Spoofs your local name by using natives and writing to a single struct (couldn't spoof name in chat otherwise)
* Add tooltip to IP spoofing to mention that it may cause you to get dropped from a session while connecting
* Expanded explanation in spoofing section to include that spoofing is server-side and not client-side.

Closes #417


yasmasdas:
That scaleform is MULTIPLAYER_CHAT and messages are added by the method ADD_MESSAGE(player, message, scope, teamOnly, eHudColour) why dont you just also hook CALL_SCALEFORM_MOVIE_METHOD_WITH_STRING

Yimura:
There's also ONLINE => Players => Names

yasmasdas:
You could go and be very wacky and just if any of the params of CALL_SCALEFORM_MOVIE_METHOD_WITH_STRING contain your name replace it with a spoofed one. Might not be smart

Sys0xdbg:
idk how this GitHub thing work but the way I do it.

```cpp
void change_name_locally(const std::string& name, CNetGamePlayer* data)
{
	auto len = std::min(name.size(), (std::size_t)18);
	std::memcpy(data->player_info->m_name, name.data(), len);
	data->player_info->m_name[len] = '\0';

}
```
it changes name from scalefrom and pause menu. but not in chat or social club when u press home button.

Yimura:
That should deal with spoofing the name in chat, though keep in mind that is seems to overwrite some struct that's used in a bunch of functions.

Fuscina:
Keep in mind r* check for spoofed names.

Yimura:
> Keep in mind r* check for spoofed names.

Yeah it might trigger a spoof detection, though changing this name only modifies it locally.
I didn't check if they check the name offset at any point.

All other spoofing methods I did by hooking natives so it shouldn't affect the structures they rely on.

Adler-D3R (commented):
Works fine for me. One more little detail could be also change the name value in the Assign Player Index Hook, but that wouldn't very useful since you only see the message one time when you join the session. But yeah, it works great.

Adler-D3R:
> idk how this GitHub thing work but the way I do it.
> 
> ```c++
> void change_name_locally(const std::string& name, CNetGamePlayer* data)
> {
> 	auto len = std::min(name.size(), (std::size_t)18);
> 	std::memcpy(data->player_info->m_name, name.data(), len);
> 	data->player_info->m_name[len] = '\0';
> 
> }
> ```
> 
> it changes name from scalefrom and pause menu. but not in chat or social club when u press home button.

Yup, it works great and even changes our name in the playerlist. However, it could create a conflict with the name restoration in Send Net Info To Lobby Hook.

Yimura:
Should be ready to merge, uses natives for the most part and patches one struct to make it so your name is also spoofed in the chat box. The struct will be correctly restored when unloading YimMenu so there should be no lasting effects.

Also added some clarification to the Spoofing submenu so people are aware it's supposed to work server-side only and added a tooltip to the IP spoofing option to warn it may causes you to get dropped from a session while connecting.

PR #523, opened 2022-10-26
bypass c4 limit

xiaoxiao921:


BigBaseV2/src/views/self/view_weapons.cpp
```diff
@@ -19,6 +20,18 @@ namespace big
...
+			if (g->weapons.bypass_c4_limit)
+			{
+				*g_pointers->m_bypass_max_count_of_active_sticky_bombs = 99;
+			}
+			else
+			{
+				*g_pointers->m_bypass_max_count_of_active_sticky_bombs = 4;
+			}
+		}
+
```
Yimura: This could be written differently:
```cpp
if (ImGui::Checkbox("Bypass C4 Limit", &g->weapons.bypass_c4_limit))
{
    const auto limit = g->weapons.bypass_c4_limit ? 99 : 4;
    *g_pointers->m_bypass_max_count_of_active_sticky_bombs = limit;
}

if (ImGui::Checkbox("Bypass C4 Limit", &g->weapons.bypass_c4_limit))
{
    *g_pointers->m_bypass_max_count_of_active_sticky_bombs = g->weapons.bypass_c4_limit ? 99 : 4;
}
```

Compiler probably optimises it the same so it's up to you.

BigBaseV2/src/pointers.cpp
```diff
@@ -413,6 +413,15 @@ namespace big
 		auto mem_region = memory::module(nullptr);
 		main_batch.run(mem_region);
 
+		if (auto pat = mem_region.scan("41 80 78 28 04 0F 85 F5 01 00 00"))
+		{
+			m_bypass_max_count_of_active_sticky_bombs = pat.add(4).as<uint8_t*>();
+
+			// declare it right now even though we write the same value
+			// so that it get cleaned up in the dctor
+			memory::byte_patch::make(m_bypass_max_count_of_active_sticky_bombs, 0x04);
```
Yimura: You're saving the bypass state to JSON but not using it on startup/inject.
```cpp
memory::byte_patch::make(m_bypass_max_count_of_active_sticky_bombs, g->weapons.bypass_c4_limit ? 99 : 4);
```

Yimura (changes requested):

BigBaseV2/src/pointers.cpp
xiaoxiao921: Actually it should be `memory::byte_patch::make(m_bypass_max_count_of_active_sticky_bombs, *m_bypass_max_count_of_active_sticky_bombs);`

Yimura (approved):

PR #529, opened 2022-10-27
feat: version badges

GeopJr:
Draft until we decide on badge style and metadata.json key names.

The badges should look like this when merged:

 

I tried to get the colors from the r* logo but they are hard to see (white on orange) so maybe we should choose a different set of colors.

Also for badge style I went with "for the badge", but here's the others:

- plastic

- flat

- flat-square

- social

(badges can be used outside of github and will always display the latest version from the metadata.json)


Yimura:
I like the flat-square design, for me the colours look good, but I can understand that on other screens it might look washed out or be completely unreadable.

The JSON name is generalised enough that we can use it for other stuff in the future (versions.json might be too constricting).

GeopJr:
do the key names sound right?

```json
{"online": "1.63",
"game": "2699"}
```

maybe

```json
{
    "game": {
        "version": "1.63",
        "build": "2699"
    }
}
```
instead?

Yimura (approved):

PR #533, opened 2022-10-28
feat(repo): add scripts

Yimura:
Unsure about the location of the scripts as you'd need to open a terminal at the root of the repo and not inside of the scripts directory.

Then execute those scripts from the root:
```bat
scripts/gen_proj.bat
```

I also added open_cmd_prompt.bat because I'm lazy but I'm definitely up for removing it entirely from the root dir.

### DIFF
--- a/open_cmd_prompt.bat
+++ b/open_cmd_prompt.bat
@@ -1,0 +1,1 @@
+start cmd.exe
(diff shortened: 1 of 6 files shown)


GeopJr (approved):

SkiddyToast (approved):
LGTM

PR #534, opened 2022-10-29
feat(spoofing): Spoof godmode

ghost:
Went a little crazy on the commits by accident. But here we are :)
All credits due to [maybegreat48](https://github.com/maybegreat48)


ghost:
Unsure what's going on, builds completely fine on my end. ill look into it tomorrow

BigBaseV2/write_player_game_state_data_node.cpp
```diff
@@ -0,0 +1,14 @@
+#include "hooking.hpp"
+
+namespace big
+{
+	bool hooks::write_player_game_state_data_node(rage::netObject* player, CPlayerGameStateDataNode* node)
+	{
+		auto ret = g_hooking->m_write_player_game_state_data_node_hook.get_original<decltype(&write_player_game_state_data_node)>()(player, node);
+
+		if (g->spoofing.spoof_hide_god)
+			node->m_is_invincible = false;
```
SkiddyToast: I'd reccomend adding these, as they would prevent potential future methods (such as `SET_ENTITY_PROOFS`) from being detectable:
```c++
a2->m_bullet_proof = false;
a2->m_melee_proof = false;
a2->m_steam_proof = false;
a2->m_explosion_proof = false;
a2->m_water_proof = false;
a2->m_fire_proof = false;
a2->m_collision_proof = false;
```

Yimura (approved):
Looks good but for the next time, could you check the repo out locally?

It'll help you with testing the build and making sure files are in the correct directories.

BigBaseV2/write_player_game_state_data_node.cpp
Yimura: Can be added in a future PR.

ghost:
you should be returning the og at the start and not at the end otherwise the values dont get spoofed

ghost:
> you should be returning the og at the start and not at the end otherwise the values dont get spoofed

That's not true. I've already tested my commits. It works.

PR #541, opened 2022-10-30
feat(self): no_collision

Aure7138:


BigBaseV2/src/backend/looped/self/no_collision.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+	void looped::self_no_collision()
+	{
+		if (g_local_player == nullptr) return;
+
+		bool bNoCollsion = g->self.no_collision;
+
+		auto set_collison = [](float value) 
+		{
+			static int offsets[] = { 0x30, 0x10, 0x20, 0x70, 0x00, 0x2C };
```
xiaoxiao921: Let's not do that and use a struct instead

BigBaseV2/src/backend/looped/self/seat_belt.cpp
```diff
@@ -0,0 +1,24 @@
+#include "backend/looped/looped.hpp"
+
+namespace big
+{
+	static bool bLastSeatBelt = false;
+
+	void looped::self_seat_belt()
```
xiaoxiao921: I thought we already had seat belt under Vehicle ?

xiaoxiao921 (changes requested):

BigBaseV2/src/backend/looped/self/seat_belt.cpp
Aure7138: my bad, didnt check vehicle view

BigBaseV2/src/backend/looped/self/no_collision.cpp
Aure7138: Ok, will make a PR to GTAV Class in quick

Aure7138:
Yes, I think it's done

xiaoxiao921 (approved):

Yimura (approved):

PR #543, opened 2022-10-30
fix c4 bypass limit setting value not loaded by the time the pointers class try to use it

xiaoxiao921:


BigBaseV2/src/main.cpp
```diff
@@ -54,6 +54,9 @@ BOOL APIENTRY DllMain(HMODULE hmod, DWORD reason, PVOID)
 
 			try
 			{
+				g->load();
+				LOG(INFO) << "Settings Loaded.";
```
tupoy-ya, line 60: I think
```cpp
g->load();
LOG(INFO) << "Settings Loaded.";
```
should be after:
```cpp
LOG(INFO) << "Yim's Menu Initializing";
```

xiaoxiao921: Good catch

Yimura (approved):

PR #546, opened 2022-10-30
feat(Script): Don't open Visual Studio if it's open already

Yimura:
It was starting to annoy me that the script would open Visual Studio regardless of if it was open or not.

This PR will change its behaviour to do nothing if it's already open.
Please tell me if you'd like for me to keep the timeout or just close the window instantly.

### DIFF
--- a/script.bat
+++ b/script.bat
@@ -12,4 +12,4 @@ IF %ERRORLEVEL% NEQ 0 (
 
 echo "Opening project..."
 tasklist /FI "WINDOWTITLE eq BigBaseV2 - Microsoft Visual Studio" /FI "STATUS eq running" 2>NUL | find /I /N "devenv.exe" >NUL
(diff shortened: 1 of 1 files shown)


tupoy-ya:
> Please tell me if you'd like for me to keep the timeout or just close the window instantly.

I think it should be "Press any key to continue".

Also a flag to run msbuild instead of visual studio could be useful.

Yimura:
> I think it should be "Press any key to continue".

You'd prefer it'd be PAUSE instead? Sure, I don't mind either.

> Also a flag to run msbuild instead of visual studio could be useful.

MSBUILD isn't in path on Windows, will have to do some additional research regarding this.

ghost:
Unrelated to this pr, but noticed a mistake in the README.md. Too small to make a dedicated PR

```
related that might get the menu targetted by Take2.
```
should be
```
related that might get the menu targeted by Take2.
```
targetted -> targeted

Yimura:
We can add MSbuild support in a later PR under the `scripts/` directory.

PR #551, opened 2022-11-05
refactor!: Replace premake5 with CMake.

tupoy-ya:
# Reason
I want to make YimMenu MinGW compatible and it's much easier to do with CMake.

Related:
+ https://github.com/YimMenu/YimMenu/pull/342#issuecomment-1268391294

# Additional context
- [x] Compiles
- [x] Runs
- [x] CI
- [x] Test with v142 tool set
- [x] Test with v143 tool set
- [x] Update `README.md`
- [x] Make everything in vendor directory appear in vs
- [x] Update script.bat
- [ ] Maybe remove some `CMakePredefinedTargets`
- [ ] Maybe use `YimMenu/bin` as output directory
- [ ] Maybe remove Debug build
- [ ] Don't increase amount of people failing to build YimMenu by 90%
- [x] Warnings as errors


GeopJr (changes requested):
Could you also get rid of gitsubmodules and instead use [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html)?

README.md
```diff
@@ -31,33 +31,49 @@ If you haven't installed git on your system go and do so it is **REQUIRED** for
...
 
 Go into the directory you just cloned:
 ```bash
 cd YimMenu
 ```
 
-Generate project files:
+#### Generate project files:
+
+Manualy:
```
TheGreenBandit, line 54: manually*

tupoy-ya:
Building `MinSizeRel` is detected by sigscanner
```bash
$ ./sigscan
(BigBaseV2_MinSizeRel.dll) { 48 8b c4 48 81 ec b8 00 00 00 0f 29 70 e8 } (14) (v2699) (PAGE_EXECUTE_READWRITE) (~0.000 kb region)
```
Should i remove it?

GeopJr (dismissed):

maybegreat48:
> Building `MinSizeRel` is detected by sigscanner

The detected function is ImGui::ScaleValueFromRatioT<double, double, double> if you are curious

CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+
+# Fetch content
+message("\nFetching modules")
+include(scripts/MinHook.cmake)
+include(scripts/g3log.cmake)
+include(scripts/pugixml.cmake)
+include(scripts/json.cmake)
+
+message("\nFetching custom modules")
+include(scripts/ImGui.cmake)
```
GeopJr:
```suggestion
include(scripts/imgui.cmake)
```

case-sensitive (at least on unix)

CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+    "${gtav_classes_SOURCE_DIR}"
+    "${minhook_SOURCE_DIR}/include"
+    "${imgui_SOURCE_DIR}"
+)
+
+target_precompile_headers(BigBaseV2 PRIVATE "${SRC_DIR}/common.hpp")
+target_link_libraries(BigBaseV2 PRIVATE pugixml minhook g3log imgui d3dcompiler)
+
+# Disable warnings
+set_target_properties(BigBaseV2 PROPERTIES COMPILE_FLAGS "/wd4996") # C4996: deprecated in C++17
```
Yimura: Where does this warning occur? Let's try to fix it instead of silencing it, we're on C++20 anyways.

BigBaseV2/src/main.cpp
```diff
@@ -31,10 +31,10 @@ BOOL APIENTRY DllMain(HMODULE hmod, DWORD reason, PVOID)
 		DisableThreadLibraryCalls(hmod);
 
 		g_hmodule = hmod;
-		g_is_steam = GetModuleHandle(L"steam_api64.dll") != NULL;
+		g_is_steam = GetModuleHandle("steam_api64.dll") != NULL;
```
Yimura: We can remove this as all there only exists one version of GTA currently (since b2699).

Yimura (changes requested):
I'm fine with keeping the `build/` folder instead of `bin/`.

Regarding build types `Debug/Release/MinSizeRel/RelWithDebInfo`, keep them all.
It should be up to the developer to choose the correct one.

Since we're doing this big change, why don't we remove the top-level `BigBaseV2/` folder and just straight up go with `src/`.
Continuing this, YimMenu is barely anything like BBv2, how about we switch the name in the CMake project to YimMenu instead?
Possibly also changing the README title as it is misleading and instead reference BBv2 as our "roots" somewhere further down in the README.

Possibly also remove the `.bat` scripts from the repository, I'll probably be using VS Code instead of Visual Studio from now on and most of the problems that people had were from the submodules which will be removed/handled by CMake instead.

tupoy-ya:
> Since we're doing this big change, why don't we remove the top-level `BigBaseV2/` folder and just straight up go with `src/`. 

I think this will remove github file history and probably make it hard to merge for forks.

GeopJr:
It will probably make it slightly hard for downstreams (they'll have to move additional files on their own) but history-wise, git (& frontends) will know the files got renamed (/moved). e.g. https://github.com/YimMenu/YimMenu/commits/master/BigBaseV2/src/detour_hook.cc

`detour_hook.cpp` got renamed to `detour_hook.cc` but the history is still available.

On git, you can pass the `--follow` flag: `git log --follow ./BigBaseV2/src/detour_hook.cc`.

CMakeLists.txt
tupoy-ya: This warning doesn't occur anymore but did when i added this.

xiaoxiao921:
Removing the top folder would make adding other projects under the same solution more annoying

Yimura:
> Removing the top folder would make adding other projects under the same solution more annoying

That's true but when will that happen with YimMenu?
IMO it's better to split other projects to another repo instead of putting them in the same repo anyways, I know it's a thing for Visual Studio, but I don't plan on using Visual Studio after this is merged in.

tupoy-ya:
> Possibly also changing the README title as it is misleading and instead reference BBv2 as our "roots" somewhere further down in the README.

Should i copy-paste [UC](https://www.unknowncheats.me/forum/grand-theft-auto-v/476972-yimmenu-1-63-a.html) version of BigBaseV2 mention into `README.md`?

Because i'm too stupid (or just lazy) to write it myself.

Yimura:
> Should i copy-paste [UC](https://www.unknowncheats.me/forum/grand-theft-auto-v/476972-yimmenu-1-63-a.html) version of BigBaseV2 mention into `README.md`?
> 
> Because i'm too stupid (or just lazy) to write it myself.

Yep, sounds good.

tupoy-ya:
Should i add `.vscode/` and `.cache/` from clangd to `.gitignore`?

Yimura (dismissed):

tupoy-ya:
But this PR doesn't add linux support.
I will open MinGW support PR after this is merged.

Yimura (approved):
Approving for now, if people still have something to say about what they'd like to be done differently have 6h left before I merge.

PR #553, opened 2022-11-06
feat(self): Added HUD options.

SpaghettDev:
Added some HUD options. These include:
- Hiding Radar
- Hiding Ammo
- Hiding specific HUD components
- And hiding everything


Yimura (dismissed):
Code looks clean, will test it in-game just to see it doesn't crash my game immediately 😜

BigBaseV2/src/views/self/view_self.cpp
```diff
@@ -158,6 +159,58 @@ namespace big
...
+		ImGui::SliderInt(
+			"##hud_comp_slider",
+			&g->self.selected_hud_component,
+			1,
+			(int)HudComponents::HUD_WEAPONS,
+			(
+				g->self.selected_hud_component >= 1 &&
+				g->self.selected_hud_component <= (int)HudComponents::HUD_WEAPONS
+			) ? hud_component_names[g->self.selected_hud_component] : hud_component_names[0]
+		);
```
Yimura: Weird way of drawing something that could be a ListBox or ComboBox.

Please change to one of these instead.

Yimura (changes requested):

BigBaseV2/src/views/self/view_self.cpp
SpaghettDev: I was gonna make it a listbox, but made it a slider instead because I thought it would take up too much of the screen. Will change nonetheless.

SpaghettDev: Done. Made it a ComboBox. Sorry for being late, I was at school practically all day.

Yimura (approved):
Looks good now :+1:

PR #554, opened 2022-11-06
feat(pointers): handle_to_ptr

Aure7138:


Yimura:
Seems like a duplicate of `get_script_handle_t` or am I wrong?

Aure7138:
You are right, I didn't notice its signature

PR #578, opened 2022-11-11
feat(Renderer): Added Cyrillic to ImGui font

### LINKED ISSUE #512: [Bug]: Russian characters aren't displayed correctly
### Describe the bug

Russian characters are displayed as question marks. Similar to the cjk problem shown here https://www.unknowncheats.me/forum/3551540-post347.html which was fixed in #487 

### Steps To Reproduce

- Generate cache with the game language set as Russian.
- Type Russian characters into any input box. 

### Expected Behavior

Show russian characters normally.

### Screenshots and/or Logs

image

### Store

Steam

### Additional context

_No response_

### Build

- [X] I used YimMenu unmodified

Yimura:
Fixes #512 

image

### DIFF
--- a/src/renderer.cpp
+++ b/src/renderer.cpp
@@ -38,10 +38,13 @@ namespace big
 		ImGui_ImplDX11_Init(m_d3d_device.Get(), m_d3d_device_context.Get());
 		ImGui_ImplWin32_Init(g_pointers->m_hwnd);
 
-		file chinese_font_file("C:/Windows/Fonts/msyh.ttc");
+		folder windows_fonts(
+			std::filesystem::path(std::getenv("SYSTEMROOT")) / "Fonts"
+		);
+
+		file chinese_font_file = windows_fonts.get_file("./msyh.ttc");
 		if (!chinese_font_file.exists())
-			chinese_font_file = { "C:/Windows/Fonts/msyh.ttf" };
-		
+			chinese_font_file = windows_fonts.get_file("./msyh.ttf");
 		auto font_file = std::ifstream(chinese_font_file.get_path(), std::ios::binary | std::ios::ate);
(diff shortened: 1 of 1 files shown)


src/renderer.cpp
```diff
@@ -38,15 +38,25 @@ namespace big
 		ImGui_ImplDX11_Init(m_d3d_device.Get(), m_d3d_device_context.Get());
 		ImGui_ImplWin32_Init(g_pointers->m_hwnd);
 
-		file font_file_path("C:/Windows/Fonts/msyh.ttc");
-		if (!font_file_path.exists())
-			font_file_path = { "C:/Windows/Fonts/msyh.ttf" };
-		auto font_file = std::ifstream(font_file_path.get_path(), std::ios::binary | std::ios::ate);
-		const auto font_data_size = static_cast<int>(font_file.tellg());
-		const auto font_data = std::make_unique<std::uint8_t[]>(font_data_size);
```
xiaoxiao921, line 46: Is assuming the drive letter is C correct here?

Yimura: I hope people have their operating system on `C:`, I think there are a lot of programs that would break if people didn't install their stuff on `C:`

xiaoxiao921: C is the default, but you can change it without any issue even though that's quite uncommon

Yimura: We can probably switch to std::getenv or smth
https://github.com/YimMenu/YimMenu/blob/e0cfb36b8fe7155510e4b9bf639713ac26f5dae7/src/main.cpp#L39

Yimura: Seems we can use `SYSTEMROOT` to get for example `C:/Windows/`

xiaoxiao921 (approved):

PR #581, opened 2022-11-11
feat(ScriptEventHandler): Block switch player to animal

Yimura:
Kids seem to really like this one.

### DIFF
--- a/src/hooks/protections/script_event_handler.cpp
+++ b/src/hooks/protections/script_event_handler.cpp
@@ -105,6 +105,17 @@ namespace big
 				return true;
 			}
 			break;
+		case eRemoteEvent::GiveCollectible:
+			if (g->protections.script_events.switch_player_model)
+			{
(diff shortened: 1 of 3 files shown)


src/hooks/protections/script_event_handler.cpp
```diff
@@ -105,6 +105,17 @@ namespace big
 				return true;
 			}
 			break;
+		case eRemoteEvent::GiveCollectible:
+			if (g->protections.script_events.switch_player_model)
+			{
+				if (args[2] == 8)
```
BugisoftRSG, line 111: Baguette

BugisoftRSG (approved):

src/hooks/protections/script_event_handler.cpp
Yimura: Thanks, very insightful indeed.

PR #588, opened 2022-11-12
Add script patcher and update protections

### LINKED ISSUE #276: Block Script Host Kick.
No, it 's not.

I'd like to know if it would be possible to add a Script Host Kick protection to the menu since it is very used by P2C, and quite annoying. I know how to do it, it's about settings 1 or 2 globals to a certain value. We can find them in Freemode Script, quite easy. The issue is that, if I'm not wrong, you need to hijack the Freemode Script to change those globals values.

**Provide Reasoning**
I'd like this ot be added because I find it useful.

**Additional context**
Adding this could make Yim users targeted by modders, cause their are Forcing Script Host (the protection needs to do this).

maybegreat48:
* Script host kick protection
* Reveal OTR players
* Block lost connection kick when host
* Redesigned native hooking
* Added ability to hide spectate
* Hide godmode now saved to JSON
* Added CMakeSettings.json for Visual Studio support
Online features not thoroughly tested. Unload seems to work. Fixes #276

### DIFF
--- a/src/hooks/protections/receive_net_message.cpp
+++ b/src/hooks/protections/receive_net_message.cpp
@@ -13,7 +13,7 @@ namespace big
 		uint32_t extended{};
 		if ((buffer.m_flagBits & 2) != 0 || (buffer.m_flagBits & 1) == 0 ? (pos = buffer.m_curBit) : (pos = buffer.m_maxBit),
 			buffer.m_bitsRead + 15 > pos || !buffer.ReadDword(&magic, 14) || magic != 0x3246 || !buffer.ReadDword(&extended, 1)) {
-			msgType = rage::eNetMessage::CMsgInvalid;
+			msgType = rage::eNetMessage::MsgInvalid;
 			return false;
 		}
 		length = extended ? 16 : 8;
@@ -21,6 +21,15 @@ namespace big
 			return true;
 		else
 			return false;
+	}
+
+	void gamer_handle_deserialize(rage::rlGamerHandle& hnd, rage::datBitBuffer& buf)
+	{
+		if ((hnd.m_platform = buf.Read<uint32_t>(8)) != 3)
+			return;
+
+		buf.ReadInt64((int64_t*)&hnd.m_rockstar_id, 64);
+		*reinterpret_cast<uint32_t*>(reinterpret_cast<__int64>(&hnd) + 9) = buf.Read<uint32_t>(8);
 	}
 
 	bool hooks::receive_net_message(void* netConnectionManager, void* a2, rage::netConnection::InFrame* frame)
@@ -43,7 +52,7 @@ namespace big
 			{
 				switch (msgType)
 				{
-				case rage::eNetMessage::CMsgTextMessage:
+				case rage::eNetMessage::MsgTextMessage:
 				{
 					if (g->session.log_chat_messages)
 					{
@@ -57,7 +66,7 @@ namespace big
 					}
 					break;
 				}
-				case rage::eNetMessage::CMsgTextMessage2:
+				case rage::eNetMessage::MsgTextMessage2:
 				{
 					if (g->session.log_text_messages)
 					{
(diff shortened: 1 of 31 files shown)


src/pointers.cpp
```diff
@@ -434,6 +434,18 @@ namespace big
...
 		});
 
+		// Init Native Tables
+		main_batch.add("INT", "8B CB E8 ? ? ? ? 8B 43 70 ? 03 C4 A9 00 C0 FF FF", [this](memory::handle ptr)
+		{
+			m_init_native_tables = ptr.add(3).rip().as<PVOID>();
+		});
+
+		// Script VM
+		main_batch.add("VM", "E8 ? ? ? ? 48 85 FF 48 89 1D ? ? ? ?", [this](memory::handle ptr)
```
xiaoxiao921: useless ? ? ? ? at the end

src/hooks/protections/receive_net_message.cpp
```diff
@@ -23,6 +23,15 @@ namespace big
 			return false;
 	}
 
+	void gamer_handle_deserialize(rage::rlGamerHandle& hnd, rage::datBitBuffer& buf)
+	{
+		if ((hnd.m_platform = buf.Read<uint32_t>(8)) != 3)
```
xiaoxiao921: ideally we'd want a variable for this `3`

src/hooks/protections/receive_net_message.cpp
```diff
@@ -23,6 +23,15 @@ namespace big
...
+			return;
+
+		buf.ReadInt64((int64_t*)&hnd.m_rockstar_id, 64);
+		*reinterpret_cast<uint32_t*>(reinterpret_cast<__int64>(&hnd) + 9) = buf.Read<uint32_t>(8);
```
xiaoxiao921: ideally we'd use a proper struct for this

src/services/script_patcher/script_patcher_service.hpp
```diff
@@ -0,0 +1,93 @@
+#pragma once
+#include <script/scrProgram.hpp>
+#include "memory/pattern.hpp"
+
+namespace big
+{
+	struct script_data
```
xiaoxiao921: Should be a class there?

xiaoxiao921 (commented):
Small nitpicks but overall very nice PR!
I'm neither approving or requesting explicit changes as other reviewers may find it un/acceptable to merge as is.

src/hooks/protections/receive_net_message.cpp
SkiddyToast: Why would we want a variable for the 3? It will never change.

xiaoxiao921: https://en.wikipedia.org/wiki/Magic_number_(programming)

src/services/script_patcher/script_patcher_service.hpp
maybegreat48: It just holds data so I think a struct is fine (please correct me if I'm wrong though)

xiaoxiao921: I said this because of the `public` access modifier, thinking it was intended, it doesnt matter much

maybegreat48: Whoops, didn't see that! Thanks

src/services/script_patcher/script_patch.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
+#include <memory/pattern.hpp>
```
Yimura: Using `<` include while somewhere else `"` is used to include this same header.

I understand the usage of `<` includes for stuff that comes from "modules" (GTAV-Classes).

Yimura (changes requested):
This is some good stuff, only have one small thing I noticed.

Yimura (approved):

PR #606, opened 2022-11-17
Improve protections again and player database

maybegreat48:
* Improved protections
* Added player database service, save players permanently
* Added block join
* Added player magnet
* Fixed idle kick


tupoy-ya:
A button to add player manually would be nice.

Also this text doesn't fit.
image

Yimura:
Currently there are some false positives on checks for infinite attachments and rope crashes.
We either need to improve detections of these or not mark people as modder.
Marking people prematurely as modder might get them targeted.

image

src/hooks/protections/receive_net_message.cpp
```diff
@@ -64,7 +65,7 @@ namespace big
 						buffer.ReadString(message, 256);
 						buffer.ReadQWord(&unk, 64);
 						buffer.ReadBool(&is_team);
-						LOG(INFO) << "[CHAT] from " << player->get_name() << ": " << message << (is_team ? " [TEAM]" : " [ALL]");
+						LOG(INFO) << "[CHAT] from " << player->get_name() << ": " << message << (is_team) ? " [TEAM]" : " [ALL]";
```
tupoy-ya: This adds weird numbers at the ends of messages:
```
[17:58:38][INFO/receive_net_message.cpp:68]: [CHAT] from no2: АХАХХАХАХААХ247
[17:58:40][INFO/receive_net_message.cpp:68]: [CHAT] from no1: ну го на снайпах0
[17:58:43][INFO/receive_net_message.cpp:68]: [CHAT] from no2: бездарность 0
[17:58:47][INFO/receive_net_message.cpp:68]: [CHAT] from no2: хааххаха0
[17:58:52][INFO/receive_net_message.cpp:68]: [CHAT] from no2: удачи хуйлуша 0
```
Sorry for a bad example.

Also this `is_team` gives a lot of false positives and probably isn't actually `is_team`.

Yimura: This is what IDA shows for the game parsing the CMsgTextMessage, it seems to read a bool of some sorts.

image

SpaghettDev:
damn, what a coincidence. I was literally making an internal player database in my fork and woke up to find that you already made one lol

Yimura (approved):
Regarding the false positives, rope crash should never trigger unless the user is a modder.
The infinite attachment crash false positives appear to have significantly dropped of after the latest changes. (I haven't had it trigger for me since but I tested it only to a limited capacity)

I want to be clear the we should never false mark people as modder when a certain protection still has false positives.

PR #621, opened 2022-11-20
feat(protection): sound overload crash protection

Aure7138:


MarkEcza (commented):
functionality wise it works perfectly fine.

Aure7138:
[cout.log](https://github.com/YimMenu/YimMenu/files/10049491/cout.log)
I think I forgot this, this is the crash log

xiaoxiao921:
This is not getting merged in it's current state, don't remove the constructor of byte patch and simply add an overload which take an an array of byte instead

Aure7138:
You are right, it was too late that day and I just thought of this method with no experience, will change it back immediately

PR #628, opened 2022-11-21
feat(self) & refactor: Add force show HUD component and replaced some hard-coded values.

SpaghettDev:
Now you can force show any HUD component.


Yimura:
I don't quite understand how this is supposed to work.
Does it force show all HUD elements when I click the toggle? Or do I have to press show after toggling force show all elements?

For example when I clicked hide it would show all elements, is this intended?

Please offer some additional explanation.

SpaghettDev:
This adds a "Force show HUD component" that forces the HUD element chosen to be forcefully shown. An example would be the "Show reticle" in the weapons tab. But instead of only having the option to show the reticle, you can force any HUD component to be shown.

The component is forcefully shown if its value in the "hud_components_states" array is false (because true means hide).

Edit: Actually I get what you mean, working on a fix.

SpaghettDev:
This might be a fix, not sure. Something's not right, and I'm not sure what it is. Now force showing anything at all will show the reticle and cash. I'll fix this when I have time.

SpaghettDev:
Should be fixed now.

Yimura (dismissed):
Code makes sense, using the UI and figuring out how it all works is a different matter.
Took me some time visually confirm it to be working in-game.

SpaghettDev:
> Code makes sense, using the UI and figuring out how it all works is a different matter.
> Took me some time visually confirm it to be working in-game.

I was going to add another bool array to hold the elements to be force shown, but  thought against it because it would just be too much for a simple feature.

Yimura (approved):

PR #632, opened 2022-11-22
chore: Renamed ptr_to_handle & some small refactoring

Aure7138:


src/views/self/view_weapons.cpp
```diff
@@ -21,9 +21,8 @@ namespace big
 		ImGui::Checkbox("Enable Special Ammo", &g->weapons.ammo_special.toggle);
 
 		if (ImGui::Checkbox("Bypass C4 Limit", &g->weapons.bypass_c4_limit))
-		{
-			*g_pointers->m_bypass_max_count_of_active_sticky_bombs = g->weapons.bypass_c4_limit ? 99 : 4;
-		}
+			g->weapons.bypass_c4_limit ? g_pointers->m_bypass_max_count_of_active_sticky_bombs->apply() : g_pointers->m_bypass_max_count_of_active_sticky_bombs->restore();
```
Yimura: Just use if else statement, there's no need to use a ternary operator here.

Yimura (changes requested):

Yimura (approved):

PR #634, opened 2022-11-24
hotkey_service

### LINKED ISSUE #248: Hotkey System
Without Hotkeys my Life is terrible 

Some Kind of hotkey System or More hotkeys for noclip..

**Provide Reasoning**
Improves the gameplay 

**Additional context**
X

Aure7138:
Closes #248


Yimura (changes requested):
Overall, the code is nice and well-structured however it's making use of GetAsyncKeyState which I dislike, just as all the previous times people have use GetAsyncKeyState to implement this.

Please WndProc instead of using GetAsyncKeyState.

Using WndProc gives several benefits over using GetAsyncKeyState:
1. Only key presses sent to GTA V game window are picked up
2. You can filter between key UP / key DOWN
3. WndProc is event based (you don't need to loop all registered keys and check for each of them if they have been pressed)
4. WndProc contains the specific key that has been pressed (do a lookup std::map::find using the key that was pressed)

Shouldn't be too hard to convert from using GetAsyncKeyState to using WndProc.

Aure7138:
I keep the vector, because the hotkey can change dynamically, and what I store is the address of the hotkey

Yimura (approved):

PR #245, opened 2022-05-23
remove(Toxic): remove toxic options

xM4ddy:

PR #636, opened 2022-11-25
crash protection

CRGBS:
andti model crash


0xEmiil:
i thought replay interface wasnt a realiable option for that kind of protections and why dont do with pools instead of allocate the hashes of the obj?

srry for my bad english~

CRGBS:
> 

because i am not code pro
maybe it is good idea
waiting for other pro do it

i am bad english too

Yimura:
I have a couple of things to say about this PR.

This seems like it's old code from YimMenu, like very old.
The replay interface was removed because it's a very bad way of filtering crashes, if you're removing crash models in the replay interface it's already too late and your game should've already crashed.
Whose code is this? These commits were made on the 3rd of June, this is ages ago. (the crash hashes for example and usage of fmt::format instead of std::format)
Then there's also the toxic code we removed from YimMenu in #245 that has been added again.
Lastly, you've added the GTAV-Classes repo as a submodule again, while we've stepped off from using git submodules and instead started using CMake FetchContent.

PR #665, opened 2022-12-08
Imrpove CEventNetwork and add comments sEntityDamagedData

Vali0004:
Mapped out all the types of eEventNetworkType and added them to an enum.


Yimura (approved):
I enjoy how R* employees where consistent in their event names. /s

PR #714, opened 2022-12-17
feat(Api): Add more api endpoints to the scui api

BugisoftRSG:
- Send socialclub message
- Get UGC Job Data via scui api and download metadata from prod.cloud


Yimura (dismissed):

src/views/network/view_player_database.cpp
```diff
@@ -108,6 +109,21 @@ namespace big
 					session::join_by_rockstar_id(current_player.rockstar_id);
 				});
 
+				static char message[20];
```
maybegreat48: Just 20 chars?

BugisoftRSG: I could swear I looked up in gta for the max chars 255 but guess I havent replaced it thanks for pointing out

maybegreat48:
Are you sure you've really implemented the UGC thing?

BugisoftRSG:
> Are you sure you've really implemented the UGC thing?

Im currently on it its more a todo list. The language from the metadata file can't be queried efficently so guess I have loop over the languages if you dont know another solution. The Job Details (scui) doesnt contain it and QueryContentByContentId (QueryContent, pros.ros) neither they are also sending all languages to the server and no the response also doesnt contain it

(all tools I know about just loop over it and search for a 200 response)

Yimura (approved):

PR #715, opened 2022-12-17
feat(Protection): Add bypass for rare report metric

### LINKED ISSUE #702: [DETECTED] YimMenu is being detected!
YimMenu is being detected by the sigscanner, my advise is not to use if you care about your account until a bypass is implemented.

We removed the previous bypass as its specific implementation was being detected by the anti-cheat. We're currently looking into other possiblities to fix this issue.

Yimura:
This should fix a detection that rarely happens but does appear to occur.

Huge thanks to @brunph for their excellent research in finding this bypass!

Fixes #702


maybegreat48 (commented):
LGTM

PR #717, opened 2022-12-18
refactor(globals): use macro's for to_json/from_json

### LINKED ISSUE #374: [Request]: Clean JSON serialization & deserialization
### Problem

The current implementation of "from_json" & "to_json" methods inside settings are not the best option at all.

### Solution

Instead of manually declaring those functions it would be better to use 

`NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(Class, members, ...)` for structs (or classes that only contains public members)
`NLOHMANN_DEFINE_TYPE_INTRUSIVE(Class, members, ...)` for classes (or structs that contains private members)

### Reason

Code will be simplified and it will be easier to read and understand.

### Additional context

More info: [official nlohman's guide to these macros](https://github.com/nlohmann/json#simplify-your-life-with-macros)

Yimura:
This PR should be compatible with the old settings except for enum values which will be reset to their default values.
The global will no longer be accessible as a pointer and instead be changed to a static instance.

```cpp
// instead of 
g->

// use 
g.
```

Closes #374

### DIFF
--- a/src/backend/backend.cpp
+++ b/src/backend/backend.cpp
@@ -13,7 +13,6 @@ namespace big
 
 		while (g_running) 
 		{
-			g->attempt_save();
 			looped::system_self_globals();
 			looped::system_update_pointers();
 			looped::system_desync_kick_protection();
--- a/src/backend/looped/player/good_options.cpp
+++ b/src/backend/looped/player/good_options.cpp
@@ -24,17 +24,17 @@ namespace big
 
 		g_player_service->iterate([](const player_entry& entry)
 		{
-			if ((g->session.off_radar_all || entry.second->off_radar) && offRadarPlayer == entry.second->id())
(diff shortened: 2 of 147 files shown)


Yimura:
@tupoy-ya, might revert the global pointer back to a static instance.

```cpp
inline menu_settings* g{};
g->
// to
inline menu_settings g;
g.
```

tupoy-ya:
> @tupoy-ya, might revert the global pointer back to a static instance.
> 
> ```c++
> inline menu_settings* g{};
> g->
> // to
> inline menu_settings g;
> g.
> ```

Is there a reason to do so?

maybegreat48:
Doing that will allow us to take the address of a setting in something like a static constructor. This will allow us to do things like create classes for each feature which could read/write to settings directly

tupoy-ya:
Seems useful so i'm fine with it.

~~I just have a bad history with breaking `g`  changes.~~

PR #738, opened 2022-12-22
fix #737

### LINKED ISSUE #737: [Request]: Block all advertising SMS
### Problem

Some text messages are sent through script events, which will not be intercepted by receive_net_message

### Solution

This is the event used to send SMS in 1.64 `1519903406`, just need to switch a few sessions to get it

### Reason

Advertising bot is annoying

### Additional context

Imgur
Imgur

Aure7138:


src/core/globals.hpp
```diff
@@ -113,42 +113,7 @@ namespace big
 				NLOHMANN_DEFINE_TYPE_INTRUSIVE(received_event, clear_ped_task, kick_vote, report_cash_spawn, modder_detect, request_control_event, vehicle_temp_action)
 			} received_event{};
 
-			struct script_event_handler
```
maybegreat48: Why did you remove this?

Aure7138: In the notification options, there are dozens of options that need to be checked, or maybe I just need to add a check all

Yimura: On one side I agree there are a lot of different logging options for each script event and simplifying it would be nice.
However I don't want to get notified of ALL script events being blocked by YimMenu.

Some are actively malicious and others may falsely trigger and aren't actually meant for you.

Yimura (approved):

PR #740, opened 2022-12-23
feat(Protections): Reenabled CanApplyData

Yimura:
Passes the inject, unload and re-inject test without crashing.
Did not test if any of the old protections caused false positives.


Yimura:
Merging this into master, did not check if these protections are all up-to-date.
We've just made it possible for them to work again.

PR #756, opened 2022-12-27
Update readme to include new library and fix issue #748

### LINKED ISSUE #748: [Bug]: Seatbelt and Drive Underwater are stored under the same variable
### Describe the bug

Checking "Drive Underwater" also forces "Seatbelt" to be checked.

### Steps To Reproduce

1. Go to Vehicle tab
2. Click on Seatbelt
3. Witness Drive underwater magically enable itself
4. Moment

### Expected Behavior

THE HACK WORKS AND ISN'T PASTED

Wasn't the menu already proofread?
(https://github.com/YimMenu/YimMenu/commit/54b3267537e3bf07f97d6a0ddb59aacf0a54d0c1)

### Screenshots and/or Logs

Gif of the phenomenon

https://i.imgur.com/jUt3nEX.gif

### Store

Steam

### Additional context

Used latest nightly build as of 12/24

### Build

TheGreenBandit:
Fixes #748


TheGreenBandit:
doing this because i didnt use yim for a bit and came back confused, when i checked readme i saw i had everything, figured out that i needed the new library and that it was not in the readme, so added it to the readme to not confuse people like me

GeopJr:
I don't think curl should be listed as a dependency:

- cpr *should* install it https://github.com/libcpr/cpr#cmake
- curl is included with Windows since Jan 22 https://curl.se/windows/microsoft.html
- I believe "Git for Windows" also installs curl

Yimura:
Yeah curl is definitely not an install requirement, it's why we went with cpr since it sets up almost everything for you.

TheGreenBandit:
> Yeah curl is definitely not an install requirement, it's why we went with cpr since it sets up almost everything for you.

really? when i tried to build with cmake I got errors, once i ran the exe in curl i was able to build Yim again with no problem

README.md
```diff
@@ -81,8 +89,8 @@ CMake is used to generate our project files, if you haven't used it before we wi
   ```
 
   Now, you will be able to open the solution, and build it.
-
-
+  ##NOTE## You will have cd build and cmake again if you add any files or you will get an error when building. //is this also required on the visual studio extension?
```
Yimura: Yes it's always required, Visual Studio, VS Code or any other IDE.
CMake instructs our compiler which files should be compiled.

README.md
```diff
@@ -116,11 +124,11 @@ Below is an incomplete list of feature that I believe are notable to this "base"
  - Custom [settings](src/core/globals.hpp) with deep compare if changes were made include auto saving
  - Clear and well structured source code
  - Includes a thread pool to queue tasks that shouldn't block the game thread, very similar to fiber pool
- - Updated natives.hpp from https://nativedb.spyral.dev
+ - Updated natives.hpp from https://nativedb.spyral.dev (We do not use fivem natives currently)
```
Yimura: Don't need to add this, YimMenu will never use FiveM natives. Only the ones officially in the game.

README.md
```diff
@@ -23,11 +23,18 @@ So here I am with an up-to-date menu focusing on protecting the user from toxic
...
  * [Git](#Git)
  * [CMake](#CMake)
 
 If you have both of the above requirements you can skip to [cloning the git repo and generating the project files](#cloning-and-generating-project-files).
 
+### Curl
+
+Curl is used to grab things like authorization tickets making pulling rockstar ids and much more possible. It is required to download as cpr, a library we use needs it, if you haven't used it before you will need to download and install it.
+
+[Download Link](https://curl.se/download.html)
```
Yimura: Your operating system needs to really be behind on updates for curl not be present, additionally YimMenu only targets Windows10 and higher operating systems.

TheGreenBandit: oh no i use a custom os of windows so that was probably why

TheGreenBandit:
removed faq, should be ready to commit

TheGreenBandit:
also closes #748

Yimura (approved):
Looks good to merge after CI has finished.

PR #770, opened 2022-12-29
Asynchronous pattern scanning using std::async.

MarkEcza:
Well, i'm open to any suggestion. Code might need improvements regarding i'm not the best. But it should improve the performance as of my testing.


src/memory/batch.cpp
```diff
@@ -3,45 +3,79 @@
...
+#endif
+		}
+#if LOG_SCAN_TIME
+		auto stop = std::chrono::high_resolution_clock::now();
+		auto duration = std::chrono::duration_cast<std::chrono::microseconds>(stop - start);
+		LOG(INFO) << "Finished pattern scanning in " << duration.count() << " microseconds.";
+#endif
+		
+		m_entries.clear();
+		m_futures.clear();
```
pelecanidae: You'll want to move this line above L72. Otherwise, you can't guarantee that the futures are actually complete and you're only measuring the time it takes to create the futures.

MarkEcza: Now i think of it yes. Also thanks for the suggestion. But it should be much faster since it's assigning one new thread to one scan(pattern) instance no?

MarkEcza: > You'll want to move this line above L72. Otherwise, you can't guarantee that the futures are actually complete and you're only measuring the time it takes to create the futures.

It should be fine now with the latest commit.

Yimura (dismissed):

Yimura:
I've made it so the asynchronously executed function returns a boolean that way we can properly exit the batch with a return value if all patterns have been found.
In pointers.cpp I've added a throw but because of g3log which hardcore catches that an decides to kill the process it doesn't unload YimMenu cleanly.

Anyways we might switch to a non-g3log solution in the future and use something else to dump offsets when GTA crashes.

Yimura (approved):
I've also reenabled logging the offsets to console, there's no humanly noticeable delay.

PR #777, opened 2022-12-30
refactor: Load tooltips from commands and disable unusable command buttons

pelecanidae:
- Replaces manual `SetTooltip` calls now that most checkboxes/buttons are backed by commands with their own descriptions
- Adds a `disable_unless` component wrapper that prevents players from being able to press buttons they can't use
  - This both greys out buttons and prevents them from being pressed
  - An unfortunate consequence is that tooltips can't be shown, as disabled components aren't marked as hovered in imgui. I don't think this is a big deal though.


Yimura (approved):
Sounds good to me.

Issue #801, opened 2023-01-02
Interior and cutscene check for spoof godmode


TheGreenBandit:
### Problem

Right now if our player is in interior or a cutscene we wont appear as godmode

### Solution

Add detection for if in cutscene or in interior
if(g.spoofing.spoof_god_mode && !in_interior && !in_cutscene) replace the original part in hooks with this
Will post a interior check and cutscene check in a bit

### Reason

It would help us stay undetected around other modders

TheGreenBandit:
const char* InCutscene(int player)
	{
		if (NETWORK::NETWORK_IS_PLAYER_IN_MP_CUTSCENE(player))
		{
			return true;
		}
		else
		{
			return false;
		}
	}

TheGreenBandit:
bool is_player_in_interior(Player player)
	{
		int interior = INTERIOR::GET_INTERIOR_FROM_ENTITY(self::ped);
		if (!interior) {
			interior = *script_global(2689235).at(player, 453).at(248).as<int*>();
		}
		return interior;
	}

Hydra-Source:
> 	const char* InCutscene(int player)
> 
> 	{
> 
> 		if (NETWORK::NETWORK_IS_PLAYER_IN_MP_CUTSCENE(player))
> 
> 		{
> 
> 			return true;
> 
> 		}
> 
> 		else
> 
> 		{
> 
> 			return false;
> 
> 		}
> 
> 	}

Why const char as a function? Lol.
Also you can just do:

return NETWORK::.......;

No need for if statements.

TheGreenBandit:
yeah, kinda shit code cause i dont knowmuch

Hydra-Source:
> 	bool is_player_in_interior(Player player)
> 
> 	{
> 
> 		int interior = INTERIOR::GET_INTERIOR_FROM_ENTITY(self::ped);
> 
> 		if (!interior) {
> 
> 			interior = *script_global(2689235).at(player, 453).at(248).as<int*>();
> 
> 		}
> 
> 		return interior;
> 
> 	}
> 
> 
> 
> these may have to be modified to fit yimmenu

I don't think you need that script global.

TheGreenBandit:
like this? 

	const char* InCutscene()
	{
		char cutscene = if(NETWORK::NETWORK_IS_PLAYER_IN_MP_CUTSCENE(self::ped))
                return cutscene;
		}
	}
bool is_player_in_interior()
{
[...]

PR #803, opened 2023-01-02
feat(Spoofing): add interior and cutscene check

TheGreenBandit:
in the hook however it is stating that the variables are undefined, needs to be fixed before it can be merged, also would like to add the checks to player info in this pr


TheGreenBandit:
solves #801 when finished

src/util/ped.hpp
```diff
@@ -23,6 +23,10 @@ namespace big::ped
 		return true;
 	}
 
+	bool is_player_in_cutscene(Ped player) { return NETWORK::NETWORK_IS_PLAYER_IN_MP_CUTSCENE(player); }
+
+	bool is_player_in_interior(Ped player) { return (INTERIOR::GET_INTERIOR_FROM_ENTITY(player) > 0) ? true : false; }
```
pelecanidae: Both of these methods should be `inline` if they're going into a header file

src/hooks/spoofing/write_player_game_state_data_node.cpp
```diff
@@ -1,13 +1,14 @@
...
+#include "util/ped.hpp"
 
 namespace big
 {
 	bool hooks::write_player_game_state_data_node(rage::netObject* player, CPlayerGameStateDataNode* node)
 	{
 		auto ret = g_hooking->get_original<write_player_game_state_data_node>()(player, node);
 
-		if (g.spoofing.spoof_hide_god)
+		if (g.spoofing.spoof_hide_god && !is_player_in_cutscene && !is_player_in_interior(self::ped))
```
pelecanidae: `is_player_in_cutscene` is missing a method invocation, e.g. `is_player_in_cutscene(self::ped)`.

Moreover, it's awkward to call natives from this context. We could either reverse the natives and figure out what they're reading (not sure), or store these values in a global variable and have a loop update them from inside of a fiber.

src/util/ped.hpp
pelecanidae: Nit: we can get rid of the ternary and just do `return (INTERIOR::GET_INTERIOR_FROM_ENTITY(player) != 0)`

TheGreenBandit: originally i had them as inline, removed inline trying to fix the definition error and never did it

src/hooks/spoofing/write_player_game_state_data_node.cpp
TheGreenBandit: > is_player_in_cutscene is missing a method invocation, e.g. is_player_in_cutscene(self::ped).

same thing here, was trying to debug forgot to put back,  yes we can but I have no knowledge of how to do that

TheGreenBandit: leaving this open so others can see

BugisoftRSG: g_local_player->m_player_info->m_game_state == eGameState::InMPCutscene

TheGreenBandit:
i will add the checks to player info later just as extra info to use

TheGreenBandit:
> i will add the checks to player info later just as extra info to use

idk how to do it, the commit is ready for its original purpose however

.github/workflows/rebase_upstream.yml
```diff
@@ -0,0 +1,18 @@
+# .github/workflows/sync.yml
```
GeopJr: Please remove this action. This *is* the upstream repo.

Don't worry about keeping your PR up-to-date, Yimura or another member will merge upstream before merging.

TheGreenBandit: Oh yeah, sorry was trying to implement a feature for myself, will delete

Yimura (approved):

src/hooks/spoofing/write_player_game_state_data_node.cpp
```diff
@@ -1,13 +1,18 @@
 #include "hooking.hpp"
 #include <datanodes/player/CPlayerGameStateDataNode.hpp>
+#include "util/ped.hpp"
 
 namespace big
 {
+	bool is_player_in_cutscene(Ped player) { return NETWORK::NETWORK_IS_PLAYER_IN_MP_CUTSCENE(player); }
```
Aure7138: Wrong native parameter passed, it should be
```cpp
bool is_player_in_cutscene(Player player) { return NETWORK::NETWORK_IS_PLAYER_IN_MP_CUTSCENE(player); }
// and the use
is_player_in_cutscene(self::id);
```

src/hooks/spoofing/write_player_game_state_data_node.cpp
```diff
@@ -1,13 +1,18 @@
 #include "hooking.hpp"
 #include <datanodes/player/CPlayerGameStateDataNode.hpp>
+#include "util/ped.hpp"
```
Aure7138: Is this header file required? because it looks like you just need to add ```#include "natives.hpp"```

Yimura (approved):

PR #836, opened 2023-01-08
(refactor & fixes) Improvements and fixes

ghost:
Added nullptr checks before calling `g_hooking->get_original<write_player_gamer_data_node>()` to prevent potential crashes.

Added a check to only spoof Rockstar dev and QA status if `g.spoofing.spoof_rockstar_id` is true.

Moved the clan data spoofing code block inside the if `(g.spoofing.spoof_crew_data)` check to ensure that clan data is only spoofed when the user has enabled the feature.

Added a check to only set the `m_is_cheater` field to `g.spoofing.spoof_cheater` if `g.spoofing.spoof_cheater` is true. This ensures that the `m_is_cheater` field is not set to true if the user has not enabled the feature.

Added a default value for the `m_clan_created_time` field to prevent it from being uninitialized in the event that `g.spoofing.spoof_crew_data` is false.

Replaced calls to `strcpy`with calls to `std::strcpy` for improved type safety.

Added a default value for the `m_clan_name` field to prevent it from being uninitialized in the event that `g.spoofing.rockstar_crew` is false.

### DIFF
--- a/src/hooks/spoofing/write_player_gamer_data_node.cpp
+++ b/src/hooks/spoofing/write_player_gamer_data_node.cpp
@@ -3,40 +3,36 @@
 
 namespace big
 {
-    void hooks::write_player_gamer_data_node(rage::netObject* player, CPlayerGamerDataNode* node)
-    {
-        g_hooking->get_original<write_player_gamer_data_node>()(player, node);
+	void hooks::write_player_gamer_data_node(rage::netObject* player, CPlayerGamerDataNode* node)
+	{
+		g_hooking->get_original<write_player_gamer_data_node>()(player, node);
 
-        // move these variables to a separate function
-        const bool spoof_rockstar_id = g.spoofing.spoof_rockstar_id;
-        const bool spoof_rockstar_dev = spoof_rockstar_id && g.spoofing.spoof_rockstar_dev;
-        const bool spoof_rockstar_qa = spoof_rockstar_id && g.spoofing.spoof_rockstar_qa;
+		if (g.spoofing.spoof_crew_data)
+		{
+			node->m_clan_data.m_clan_member_id = 1;
+			node->m_clan_data.m_clan_id = 41564112;
+			node->m_clan_data.m_clan_id_2 = 41564112;
+			node->m_clan_data.m_clan_color = 420;
+			node->m_clan_data.m_clan_member_count = 15;
+			node->m_clan_data.m_clan_created_time = 420;
 
-        if (g.spoofing.spoof_crew_data)
-        {
-            node->m_clan_data.m_clan_member_id = 1;
-            node->m_clan_data.m_clan_id = 41564112;
-            node->m_clan_data.m_clan_id_2 = 41564112;
-            node->m_clan_data.m_clan_color = 420;
-            node->m_clan_data.m_clan_member_count = 15;
-            node->m_clan_data.m_clan_created_time = 420;
+			memcpy(node->m_clan_data.m_clan_tag, g.spoofing.crew_tag.c_str(), std::min(g.spoofing.crew_tag.size(), sizeof(node->m_clan_data.m_clan_tag)));
+        		node->m_clan_data.m_is_system_clan = g.spoofing.rockstar_crew;
+        		node->m_clan_data.m_is_clan_open = g.spoofing.square_crew_tag;
 
-            strcpy(node->m_clan_data.m_clan_tag, g.spoofing.crew_tag.c_str());
-            node->m_clan_data.m_is_system_clan = g.spoofing.rockstar_crew;
-            node->m_clan_data.m_is_clan_open = g.spoofing.square_crew_tag;
+			if (g.spoofing.rockstar_crew)
+			{
+				strcpy(node->m_clan_data.m_clan_name, "Rockstar");
+			}
+		}
 
-            if (g.spoofing.rockstar_crew)
-            {
-                strcpy(node->m_clan_data.m_clan_name, "Rockstar");
-            }
-        }
+		node->m_is_cheater = g.spoofing.spoof_cheater;
 
-        node->m_is_cheater = g.spoofing.spoof_cheater;
-        
-        if (g.spoofing.spoof_rockstar_id)
-	    {
-		    node->m_is_rockstar_dev = g.spoofing.spoof_rockstar_dev;
-		    node->m_is_rockstar_qa = g.spoofing.spoof_rockstar_qa;
-	    }
-    }
+		// spoof r* dev and qa only when spoofing rid because every player in the session will send metrics if you join with that enabled
+		if (g.spoofing.spoof_rockstar_id)
+		{
+			node->m_is_rockstar_dev = g.spoofing.spoof_rockstar_dev;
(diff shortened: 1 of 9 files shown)


xiaoxiao921:
Unless i'm missing something, all this PR does is wasting extra cpu cycles, also, where are the nullchecks you are talking about?

ghost:
> Unless i'm missing something, all this PR does is wasting extra cpu cycles, also, where are the nullchecks you are talking about?

Instead of running the looped functions every frame, the improved code only runs the functions when `g_running` is `true`. This means that the looped functions will only be run when they are needed, rather than constantly running and potentially wasting CPU cycles.

As for the null checks:

- Before calling the original `write_player_gamer_data_node` function, a null check was added to ensure that `player` is not null

- Before accessing `node->m_clan_data`, a null check was added to ensure that `node` is not null

src/hooks/spoofing/write_player_gamer_data_node.cpp
```diff
@@ -3,36 +3,37 @@
...
-		if (g.spoofing.spoof_rockstar_id)
-		{
-			node->m_is_rockstar_dev = g.spoofing.spoof_rockstar_dev;
-			node->m_is_rockstar_qa = g.spoofing.spoof_rockstar_qa;
-		}
-	}
+        node->m_is_cheater = g.spoofing.spoof_cheater;
+
+        node->m_is_rockstar_dev = spoof_rockstar_dev;
+        node->m_is_rockstar_qa = spoof_rockstar_qa;
```
Yimura: There was a reason these were wrapped in an if-statement that checked if your Rockstar ID was spoofed.
It's because that setting these will cause players to send out metrics snitching on your RID.

There was even a comment in the old code explaining this.
https://github.com/YimMenu/YimMenu/blob/86c44f38b5a480cb13138d177a64659f4aa3ac71/src/hooks/spoofing/write_player_gamer_data_node.cpp#L31

src/hooks/spoofing/write_player_gamer_data_node.cpp
```diff
@@ -3,36 +3,37 @@
...
+            node->m_clan_data.m_clan_color = 420;
+            node->m_clan_data.m_clan_member_count = 15;
+            node->m_clan_data.m_clan_created_time = 420;
 
-			if (g.spoofing.rockstar_crew)
-			{
-				strcpy(node->m_clan_data.m_clan_name, "Rockstar");
-			}
-		}
+            strcpy(node->m_clan_data.m_clan_tag, g.spoofing.crew_tag.c_str());
```
Yimura: strcpy should be safe to use as std::string::c_str() is guaranteed to have string termination char, alternatively use memcpy that limits to the size of `m_clan_tag` if you're worried about writing too much.

src/hooks/spoofing/write_player_gamer_data_node.cpp
```diff
@@ -3,36 +3,37 @@
...
-			node->m_clan_data.m_clan_member_id = 1;
-			node->m_clan_data.m_clan_id = 41564112;
-			node->m_clan_data.m_clan_id_2 = 41564112;
-			node->m_clan_data.m_clan_color = 420;
-			node->m_clan_data.m_clan_member_count = 15;
-			node->m_clan_data.m_clan_created_time = 420;
+        // move these variables to a separate function
+        const bool spoof_rockstar_id = g.spoofing.spoof_rockstar_id;
+        const bool spoof_rockstar_dev = spoof_rockstar_id && g.spoofing.spoof_rockstar_dev;
+        const bool spoof_rockstar_qa = spoof_rockstar_id && g.spoofing.spoof_rockstar_qa;
```
Yimura: No need to move these outside of their respective scopes, the compiler will optimise this as it'll deem fit.

LiamD-Flop:
> Instead of running the looped functions every frame, the improved code only runs the functions when `g_running` is `true`. This means that the looped functions will only be run when they are needed, rather than constantly running and potentially wasting CPU cycles.

> As for the null checks:
> 
> * Before calling the original `write_player_gamer_data_node` function, a null check was added to ensure that `player` is not null
> * Before accessing `node->m_clan_data`, a null check was added to ensure that `node` is not null

1. g_running is set to true when the menu is fully loaded. So adding a check for this doesn't seem to achieve what you are saying as this will always be true so the method will still always be running. When g_running is false the menu will do it's cleanup so no reason for it to have a check here. Therefor adding this will add a useless check to the method.

2. I don't see you ever implementing the g_running check so I guess disregard 1. *confused ?*
3. You say you are implementing null checks but again. I never see them implemented in this 34 line code.

As a conclusion the only change I noted in this code is that you moved an if statement to inline and run it twice instead of once. So as far as "optimising" goes, it seems like nothing really changed and if the compiler doesn't catch it there is actually more checks that are happening for no reason.

If you actually implemented the things you noted here, then I think you should check you git status as these changes do not seem to be committed.

src/hooks/spoofing/write_player_gamer_data_node.cpp
ghost: Should be fixed.

src/hooks/spoofing/write_player_gamer_data_node.cpp
ghost: Decided to replace with memcpy.

src/hooks/spoofing/write_player_gamer_data_node.cpp
ghost: I have reverted it, should be fixed now.

src/hooks/spoofing/write_player_gamer_data_node.cpp
```diff
@@ -16,9 +16,9 @@ namespace big
 			node->m_clan_data.m_clan_member_count = 15;
 			node->m_clan_data.m_clan_created_time = 420;
 
-			strcpy(node->m_clan_data.m_clan_tag, g.spoofing.crew_tag.c_str());
-			node->m_clan_data.m_is_system_clan = g.spoofing.rockstar_crew;
-			node->m_clan_data.m_is_clan_open = g.spoofing.square_crew_tag;
+			memcpy(node->m_clan_data.m_clan_tag, g.spoofing.crew_tag.c_str(), std::min(g.spoofing.crew_tag.size(), sizeof(node->m_clan_data.m_clan_tag)));
+        		node->m_clan_data.m_is_system_clan = g.spoofing.rockstar_crew;
+        		node->m_clan_data.m_is_clan_open = g.spoofing.square_crew_tag;
```
Aure7138, line 21: Using strcpy without checking the size is fatal, but there is already a check here https://github.com/YimMenu/YimMenu/blob/86c44f38b5a480cb13138d177a64659f4aa3ac71/src/views/network/view_spoofing.cpp#L65-L67, so memcpy is not needed

Yimura: Though the code you linked to checks against the m_name member var and not the m_clan_tag member.

Aure7138: just edit

ghost:
It seems that none of the changes I proposed are necessary so I will close this pull request.

PR #838, opened 2023-01-09
feat(Self): Part Water (Moses Mode)

xM4ddy:


TheGreenBandit:
i already tried this on my branch, doesnt work, will test urs tho

TheGreenBandit:
also why have you not put it as a command? Here's the feature you made but in the new command style

```
#include "backend/looped/looped.hpp"
#include "natives.hpp"
#include "backend/looped_command.hpp"

namespace big
{
	class partwater : looped_command
	{
		using looped_command::looped_command;

		virtual void on_tick() override
		{
			float offset[] = { -4, 4 };
			for (int i = 0; i < 5; i++)
			{
				if (i < 2)
				{
					self::pos.x += offset[(i % 2 == 0)];
				}
				else if (i < 4)
				{
					self::pos.y += offset[(i % 2 == 0)];
					WATER::MODIFY_WATER(self::pos.x, self::pos.y, -500000.0f, 0.2f);
				}
			}
		}
	};

	partwater g_invisibility("partwater", "Part Water", "Makes you like Moses", g.self.part_water);
}
```

src/core/globals.hpp
```diff
@@ -270,7 +271,7 @@ namespace big
 
 			NLOHMANN_DEFINE_TYPE_INTRUSIVE(self,
 				clean_player, force_wanted_level, free_cam, invisibility, local_visibility, never_wanted, no_ragdoll,
-				noclip, off_radar, super_run, no_collision, unlimited_oxygen, no_water_collision, wanted_level, god_mode,
+				noclip, off_radar, super_run, no_collision, unlimited_oxygen, no_water_collision, wanted_level, god_mode, part_water,
```
TheGreenBandit, line 274: why add it here and not at the bottom?

TheGreenBandit:
check out latest commit in the branch

https://github.com/TheGreenBandit/YimMenu/tree/maddywater

src/core/globals.hpp
Yimura: Position doesn't really matter as there's no real order in the variables, if there was some I'd mind but here it wouldn't change much.

xM4ddy:
@TheGreenBandit 
> fix the position of the modifywater and how far it goes down

I think this may be about as good as it is going to work without clearing out vast swathes of ocean
> the player becoming invisible

Didn't run into this in testing. Steps to reproduce?

TheGreenBandit:
> Didn't run into this in testing. Steps to reproduce?
> 

I will try to reproduce with stock settings and see if it works

src/core/globals.hpp
xM4ddy: Didn't matter + for a long time they were alphabetical so I just put it before the P's w/o looking.

Yimura: Yeah I put them alphabetically out of habit but nobody else cares about this lmao

Yimura (dismissed):

TheGreenBandit:
Hey @xM4ddy can you fix the hotkey names view_settings in this pr? I didn't notice how inconsistent they were

ghost:
This implementation seems a little jank, you need a beach for it to actually start properly, and if you go a little deeper it's radius will get smaller. You can't run because it can't catch up too..

image

src/backend/looped/self/part_water.cpp
```diff
@@ -0,0 +1,30 @@
...
+				{
+					self::pos.x += offset[(i % 2 == 0)];
+				}
+				else if (i < 4)
+				{
+					self::pos.y += offset[(i % 2 == 0)];
+					WATER::MODIFY_WATER(self::pos.x, self::pos.y, -500000.0f, 0.2f);
+				}
+			}
+		}
```
pelecanidae: I don't follow what's going on here. The loop unrolls to this:

```
self::pos.x += 4;

self::pos.x += -4;

self::pos.y += 4;
WATER::MODIFY_WATER(self::pos.x, self::pos.y, -500000.0f, 0.2f);

self::pos.y += -4;
WATER::MODIFY_WATER(self::pos.x, self::pos.y, -500000.0f, 0.2f);
```

- Mutating `self::pos` is weird, you probably want to make a locally-scoped copy and mutate it
- The first two loop iterations do nothing
- The last loop iteration does nothing

xM4ddy: Awesome catch the native call was in the wrong place, was working too fast.

xM4ddy:
@FireFox101889 
> This implementation seems a little jank, you need a beach for it to actually start properly, and if you go a little deeper it's radius will get smaller. You can't run because it can't catch up too..

Fixed bug should be more effective now. Mind re-testing?

BugisoftRSG:
> @FireFox101889
> 
> > This implementation seems a little jank, you need a beach for it to actually start properly, and if you go a little deeper it's radius will get smaller. You can't run because it can't catch up too..
> 
> Fixed bug should be more effective now. Mind re-testing?

Now its better but once the water gets a bit deeper and your head is below the water level its over

ghost:
still just doesn't seem fast enough. and yeah once underwater it's definitely over
image

src/backend/looped/self/part_water.cpp
```diff
@@ -0,0 +1,32 @@
...
+				else if (i < 4)
+				{
+					coords.y += offset[(i % 2 == 0)];
+				}
+				WATER::MODIFY_WATER(coords.x, coords.y, -500000.0f, 0.2f);
+			}
+		}
+	};
+
+	partwater g_invisibility("partwater", "Part Water", "Makes you like Moses", g.world.water.part_water);
```
LiamD-Flop: Copy paste forgot to rename variable

LiamD-Flop (changes requested):

src/backend/looped/self/part_water.cpp
Yimura: 👀

LiamD-Flop (approved):

LiamD-Flop:
Like others commented. This only seems to work in undeep water. sewer water will not part. In deep water, water will go down in the area you stand but only for the first +- player height.

LiamD-Flop:
I tried editing the values to make it work better but it didn't really change much. So I guess for now this is as good as it's going to be.

PR #846, opened 2023-01-11
Use cdn url

oakboat:
raw.​githubusercontent.​com is not reachable in China.


oakboat:
image
cdn version.

Yimura (approved):
Thanks a lot that solves that.

PR #850, opened 2023-01-11
feat(misc): Added a quick search bar

LiamD-Flop:
Quick search bar to execute commands on the go :)

Haven't done cpp in a while so feel free to suggest stuff.

Before PR becomes open:
- [x] Hotkey to open search bar
- [x] Execute commands on enter
- [ ] Suggest commands while typing
- [ ] Implement history of previously executed commands when there are no suggestions
- [ ] Wait for translation service to be merged with this branch


src/renderer.cpp
```diff
@@ -165,10 +165,7 @@ namespace big
 		for (const auto& cb : m_wndproc_callbacks)
 			cb(hwnd, msg, wparam, lparam);
 
-		if (g_gui->is_open())
-		{
-			ImGui_ImplWin32_WndProcHandler(hwnd, msg, wparam, lparam);
-		}
+		ImGui_ImplWin32_WndProcHandler(hwnd, msg, wparam, lparam);
```
ghost, line 168: I've did precisely all of this when maybegreat48 gave me a hard on for commands. I came across an issue with inputbox glitching out and not responding. Unsure why, Pissed me off... I recommend making this not always on btw. The user can press things if any imgui window is open without knowledge. Can come back to haunt you in the future

LiamD-Flop: Yeah I discussed this with Yimura and we came to the conclusion that ImGui probably used key down events. That when I press enter on a command it will close my quick search. But the last key event ImGui received is an enter down. So when I open it again it will continue from having received this event and just start acting like a stuck key. That's our guess.

Yimura: As Flop said, this will fix a number of issues that ImGui would randomly have (stuck input fields that can't be modified, keys that remain pressed down, etc...). In our testing it did not negatively impact performance and I'd think that the ImGui team would properly optimise this function as in their implementation there's no need to disable calls to this function.

The only reason this was ever like this is because pocakking designed it that way 4 years ago.

src/gui.cpp
```diff
@@ -165,6 +166,8 @@ namespace big
 
 	void gui::wndproc(HWND hwnd, UINT msg, WPARAM wparam, LPARAM lparam)
 	{
+		if (VK_OEM_3 == wparam)
+			std::cout << "Yeet";
```
Yimura: Debug comment?

src/views/core/view_console.cpp
```diff
@@ -0,0 +1,37 @@
+#include "views/view.hpp"
+#include "pointers.hpp"
+#include "services/console/console_service.hpp"
+#include "backend/command.hpp"
```
Yimura: Already included by `services/console/console_service.hpp`

src/gui.cpp
```diff
@@ -15,10 +15,11 @@ namespace big
 	{
 		g_renderer->add_dx_callback(view::gta_data, -1); // -1 highest priority of drawing
 		g_renderer->add_dx_callback(view::notifications, -2); // second highest priority
+		g_renderer->add_dx_callback(view::console, -3); // third highest priority
 		g_renderer->add_dx_callback([this]
 		{
 			dx_on_tick();
-		}, -3); // 3rd highest priority
+		}, -4); // 4rd highest priority
```
Yimura: 4th not 4rd

src/renderer.cpp
ghost: But it also handles mouse input. Which can close, minimize, resize and press widgets etc for a window. That was the premise of my argument..

LiamD-Flop: From what I can see this shouldn't be an issue as the gui doesn't exist when you don't have it open. So it shouldn't be able to interact with any of it. And you're not able to interact with the items that do continuously draw as they get drawn from scratch and don't use any of the ImGui built in components, neither do they listen to keypresses.

ghost:
```c++
float draw_x = ImGui::GetCursorScreenPos().x;
float draw_y = ImGui::GetCursorScreenPos().y;

for (int i = 0; i < possible_commands.size(); i++)
{
    command* cmd = possible_commands[i]; 
    std::string content = cmd->get_label();

    draw_y += 5.f; // padding between

    draw->AddRectFilled(
	{ draw_x, draw_y },
	{ draw_x + 5.f + ImGui::CalcTextSize(content.c_str()).x + 5.f, draw_y + 5.f + ImGui::CalcTextSize(content.c_str()).y + 5.f  },
	ImColor{ 0, 0, 0, 200 },
	6.f,
	0
    );
    draw->AddText(
	{ draw_x + 5.f, draw_y + 5.f },
	ImColor{ 255, 255, 255, 200 },
	content.c_str()
    );
    draw_y += 5.f + ImGui::CalcTextSize(content.c_str()).y + 5.f; // total height of box
}
```
here's a little something i put together in 10 minutes. Just an example, edit to your liking. It uses the cursor position which needs to be be directly after ImGui::InputText 
image

src/backend/looped/self/off_radar.cpp
```diff
@@ -20,5 +20,5 @@ namespace big
 		}
 	};
 
-	off_radar g_off_radar("otr", "Off Radar", "Hides your blip from other players", g.self.off_radar);
+	off_radar g_off_radar("otr", "Off the Radar", "Hides your blip from other players", g.self.off_radar);
```
TheGreenBandit, line 23: Capitalize "the"

ghost: "Do not capitalize an article (a, an, the) unless it is first or last in the title. Do not capitalize a coordinating conjunction (and, or, nor, but, for, yet, so) unless it is first or last in the title."

LiamD-Flop: No fix needed as this PR will only be merged after the translation service gets added. Which will generalize this text anyways.

Yimura:
Closing in preference to #1029

PR #855, opened 2023-01-13
Fixes Invisable Animals

Dafoosh:
Animal models had some draw issue's.
Code may not be the best lol buts its how ive gotten it to work correctly feel free to fix anything you see wrong or inefficient.


ghost (commented):
I don't think the if statement in necessary because it returns false regardless.

src/util/ped.hpp
```diff
@@ -19,7 +19,19 @@ namespace big::ped
 		PLAYER::SET_PLAYER_MODEL(self::id, hash);
 		script::get_current()->yield();
 		STREAMING::SET_MODEL_AS_NO_LONGER_NEEDED(hash);
-
+		if (ENTITY::GET_ENTITY_MODEL(self::ped) != ENTITY::GET_ENTITY_MODEL(self::ped)) {
+			return false;
+		}
```
LiamD-Flop: I might be confused but it looks like this will never be true.

LiamD-Flop (dismissed):

src/util/ped.hpp
Yimura: Correct since both calls are made on the same tick, it's best to store model from the last tick and compare against that one.

Yimura:
@Dafoosh please review the requested changes or this PR can't be merged into master.

Dafoosh (commented):
Requested Changes Made ☺.

Yimura (approved):

PR #875, opened 2023-01-17
Chore(Rearrange): Moved Files Around YimMenu To Better Suit Their Purpose

TheGreenBandit:


TheGreenBandit:
unfinished as i need to write the code for the includes, posting now for comments on file placement or extra files requested to be moved

FireFox101889:
Ewww. Everything just looks... wrong...

TheGreenBandit:
> Ewww. Everything just looks... wrong...

ideas to improve it?

Yimura:
I don't see a need to this this immediately, let's take some time before we restructure because it'll affect forks of YimMenu trying to sync up with master.

TheGreenBandit:
> I don't see a need to this this immediately, let's take some time before we restructure because it'll affect forks of YimMenu trying to sync up with master.

true, close if you like, we can do this in multiple prs, only thing i really would like to rearrange is native hooks in hooks and gui.cpp and .hpp in the gui folder

Yimura:
native hooks is fine in a separate folder as they're native hooks and not conventional hooks, they way they're made differs a lot as well compared to other hooks (VMT, Detour, IAT, EAT, ...)

TheGreenBandit:
> native hooks is fine in a separate folder as they're native hooks and not conventional hooks, they way they're made differs a lot as well compared to other hooks (VMT, Detour, IAT, EAT, ...)
> 
> Regarding to the gui.cpp/hpp file they'll be moved when restructuring the UI.

ah true, we should also move renderer.cpp and hpp too then? I am happy to wait to see new ui when it's finished

PR #884, opened 2023-01-19
feat: stat editor

Aure7138:


yasmasdas:
Maybe a danger warning would be nice? some stats are honeypot stats or heavilly monitored

Aure7138:
> Maybe a danger warning would be nice? some stats are honeypot stats or heavilly monitored

Yes, it's dangerous to cap it more than it should, my knowledge about stat is only from https://www.unknowncheats.me/forum/grand-theft-auto-v/461672-gtahax-1-64-external-thread-3-a.html, I don't know how to implement your idea, maybe I should add "Be aware of stat limits..."

Aure7138:
> > 
> 
> Yes, it's dangerous to cap it more than it should, my knowledge about stat is only from https://www.unknowncheats.me/forum/grand-theft-auto-v/461672-gtahax-1-64-external-thread-3-a.html, I don't know how to implement your idea, maybe I should add "Be aware of stat limits..."

yasmasdas:
There are multiple stats which are known to get people banned but I just mean like maybe a label above it saying that you should be aware that there is a risk with some setting stats and that it should be used with caution that should be enough. Something along those lines

Yimura (approved):

PR #907, opened 2023-01-25
Added Cage Player option

Dafoosh:
Toxic option

### DIFF
--- a/src/backend/looped/self/request_services.cpp
+++ b/src/backend/looped/self/request_services.cpp
@@ -1,0 +1,29 @@
+#include "backend/command.hpp"
+#include "script_global.hpp"
+#include "util/mobile.hpp"
+
+namespace big
+{
+	class boat_pickup : command
+	{
+		using command::command;
+
+		virtual void execute(const std::vector<std::uint64_t>&, const std::shared_ptr<command_context> ctx)
+		{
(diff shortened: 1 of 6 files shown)


Seanghost117:
You are creating 2 files that already exist, and one of them, you edited (being toxic player options) and create. Watch your file locations and how you are merging. Also, there isn't a need to make the same function in ped.hpp that exists in entity.hpp, except with an extra check. You also don't even need to make that function, the function you pasted above it has that check because you are stealing other's outfits, so it shouldn't steal yours if you click it.

cage_player.cpp
```diff
@@ -0,0 +1,27 @@
...
+{
+	class cage_player : player_command
+	{
+		using player_command::player_command;
+
+		virtual void execute(player_ptr player, const std::vector<std::uint64_t>& _args, const std::shared_ptr<command_context> ctx)
+		{
+			
+			
+
```
TheGreenBandit, line 17: very werid pr here with lots of unneeded spaces as seen where this comment is, duplicated files set in the src, entire pr should be fixed,

Yimura (changes requested):
Please check your directory structure, 3 files were moved to the root of the project for some reason.

- cage_player.hpp
- ped.hpp
- player_toxic.hpp

Remove those and I'll have another look at your PR.

PR #910, opened 2023-01-27
feat(ui): add ingame overlay

Primexz:
Hey,
I have now decided to actively contribute to YimMenu.

Let's start with a small ingame overlay with some info like FPS / game version and so on.

### DIFF
--- a/src/views/core/view_overlay.cpp
+++ b/src/views/core/view_overlay.cpp
@@ -26,7 +26,7 @@ namespace big
 				ImGui::Text(std::format("Players: {}/{}", network_player_mgr->m_player_count, network_player_mgr->m_player_limit).c_str());
 
 			if (g.window.ingame_overlay.show_time)
-				ImGui::Text(std::format("Time: {:%d-%m-%Y %H:%M:%OS}", std::chrono::system_clock::now()).c_str());
+				ImGui::Text(std::format("Time: {:%d-%m-%Y %H:%M:%OS}", std::chrono::current_zone()->to_local(std::chrono::system_clock::now())).c_str());
 
 			if (auto replay_interface = *g_pointers->m_replay_interface; g.window.ingame_overlay.show_replay_interface)
 			{
--- a/src/views/core/view_overlay.cpp
+++ b/src/views/core/view_overlay.cpp
@@ -1,8 +1,6 @@
 #include "views/view.hpp"
 #include "pointers.hpp"
-#include "util/math.hpp"
 #include "gta_util.hpp"
-#include "util/misc.hpp"
 #include "gui.hpp"
 
 namespace big
@@ -25,6 +23,8 @@ namespace big
 
 			if (CNetworkPlayerMgr *network_player_mgr = gta_util::get_network_player_mgr())
 				ImGui::Text(std::format("Players: {}/{}", network_player_mgr->m_player_count, network_player_mgr->m_player_limit).c_str());
+
+			ImGui::Text(std::format("Time: {:%d-%m-%Y %H:%M:%OS}", std::chrono::system_clock::now()).c_str());
 
             ImGui::Separator();	
 
--- a/src/views/core/view_overlay.cpp
+++ b/src/views/core/view_overlay.cpp
@@ -1,0 +1,49 @@
+#include "views/view.hpp"
+#include "pointers.hpp"
+#include "util/math.hpp"
+#include "gta_util.hpp"
+#include "util/misc.hpp"
+#include "gui.hpp"
+
+namespace big
+{
+	void view::overlay()
+	{
+		if (!g.window.ingame_overlay.opened || g_gui->is_open())
+            return;
+
+		ImGui::SetNextWindowPos(ImVec2(10.0f, 10.0f), ImGuiCond_Always, ImVec2(0.0f, 0.0f));
+		ImGui::SetNextWindowBgAlpha(0.3f);
+
+		if (ImGui::Begin("overlay", nullptr, ImGuiWindowFlags_NoMove | ImGuiWindowFlags_NoTitleBar | ImGuiWindowFlags_NoResize | ImGuiWindowFlags_AlwaysAutoResize | ImGuiWindowFlags_NoSavedSettings | ImGuiWindowFlags_NoFocusOnAppearing | ImGuiWindowFlags_NoNav))
+		{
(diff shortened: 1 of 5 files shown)


yasmasdas:
Maybe it would be nice if you could customize what can be shown in this overlay? like if you only want it to show fps and players thats a option.
also the real life time would be nice

TheGreenBandit:
u can do the above comment by doing
if(g.window.ingame_overlay.ped_interface)
{
     if(replay_interface->m_ped_interface)
     {
             ImGui::Text(std::format("Ped Pool: {}/{}", replay_interface->m_ped_interface->m_cur_peds, replay_interface->m_ped_interface->m_max_peds).c_str());
     }
}
stuff like that for each part you are displaying

Seanghost117:
You should probably add in the ability to edit the position of the overlay or at least preset it to the four corners of the screen. Just saying, especially online, the text can get mushed by the player list. Also instead of nesting as much, you should opt to condense that, something like:
(Also I would allow changing colors of specific text but that's just me, sometimes certain colors can feel better or worse on a specific person's eyes)
```
if ( initialize and update the replay interface / check if its valid )
{
       initialize your three interfaces
       if ( check if they are valid or not)
       {
             Add Text (current / max );
       }
}

**Comment by TheGreenBandit** [comment]
> You should probably add in the ability to edit the position of the overlay or at least preset it to the four corners of the screen. Just saying, especially online, the text can get mushed by the player list. Also instead of nesting as much, you should opt to condense that, something like: (Also I would allow changing colors of specific text but that's just me, sometimes certain colors can feel better or worse on a specific person's eyes)
> 
> ```
> if ( initialize and update the replay interface / check if its valid )
> {
>        initialize your three interfaces
>        if ( check if they are valid or not)
>        {
>              Add Text (current / max );
>        }
> }
> ```

could be fixed by just removing the flag ImGuiWindowFlags_NoMove

**Comment by Primexz** [comment]
Cool ideas here.
I'll get right on it ^^

File: src/views/core/view_overlay.cpp [hunk 1]
```diff
@@ -0,0 +1,54 @@
...
+{
+	void view::overlay()
+	{
+		if (!g.window.ingame_overlay.opened || g_gui->is_open())
+            return;
+
+		ImGui::SetNextWindowPos(ImVec2(10.0f, 10.0f), ImGuiCond_Always, ImVec2(0.0f, 0.0f));
+		ImGui::SetNextWindowBgAlpha(0.3f);
+
+		if (ImGui::Begin("overlay", nullptr, ImGuiWindowFlags_NoMove | ImGuiWindowFlags_NoTitleBar | ImGuiWindowFlags_NoResize | ImGuiWindowFlags_AlwaysAutoResize | ImGuiWindowFlags_NoSavedSettings | ImGuiWindowFlags_NoFocusOnAppearing | ImGuiWindowFlags_NoNav))
```
TheGreenBandit, line 16: looks great so far, only thing i would change is removing the no move flag

Primexz: All right, I will remove the flag later.
Another question: do we want to enable the overlay by default?
Currently it is disabled by default.

Seanghost117: If it’s ok the left side I recommend we either bump it under the water mark or move it’s position and enable it by default. It’s helpful but also people won’t go out of their way to find something they don’t know about.

TheGreenBandit: > All right, I will remove the flag later. Another question: do we want to enable the overlay by default? Currently it is disabled by default.

i would have it enabled

Primexz: I have thought a bit about the positioning.

On the left side we have the problem that the notifications from GTA are rendered there. 
Without the NoMove flag, users could adjust the position themselves, which I personally would also find good.
The negative point is that users would have to bring the overlay into focus first, as it is rendered behind Yim.
Unfortunately there is not much space to click on the overlay..

EDIT:
I have now added a flag that allows users to set whether the overlay is rendered when Yim is open.
By default this flag is disabled.
So by default the position can't be changed but with the flag turned on the users can move the overlay freely.
I think this is an acceptable solution.

Seanghost117: I think we still have an issue of the clipping, my honest suggestion is the watermark isn’t exactly necessary, the unload button can be removed and even made a hotkey as well, yim menu itself can be added into the overlay and the name was never important and this would fix the issue even more so and also would just benefit the overall ui with the addition of the overlay.

Primexz: I would say it would not be the best idea to integrate the menu directly into the overlay. The idea was that the overlay is always visible and always shows the basic info regardless of whether you are actively doing something in the menu or not.

Seanghost117: Yes I understand that, I’m saying there is a problem with it clipping when the menu is open, that is due to the watermark, my suggestion is to just remove the watermark, change where the unload button is in the menu or add a hornet for it and to add the text yim menu not yim menu itself into the overlay. Now we don’t have a clipping issue and it can be freely adjusted with no clashing. And the watermark doesn’t detract from the overall ui, it’s big and bulky as is.

Primexz: Your idea sounds good so far.
Before I make such "major" changes to the UI, I'll wait for @Yimura and his opinion on this topic.

Yimura (approved):
Nice design, I like it.

image

src/views/core/view_overlay.cpp
Yimura: I don't want to make any major changes to the UI as the translation branch is still not merged, I should probably just merge it and then slowly transition other not-yet translated parts later on.

There's also the route that we might just redo the entire UI at some point.

Primexz: Basically, I think the current UI looks good and is also relatively user-friendly.
Unfortunately, there are only minor problems such as with the overlay to integrate it reasonably into the UI.

Seanghost117: If the UI does get changed I have some ideas. As of current a lot, especially tabs like self are just to the point where they take up a vast majority of the screen. It has a nice minimal design on init but I feel once you go into a tab its almost as if things are just "put there". If the UI is changed, I've started building an icon library, hacking the child to be more of use (because as imgui is right now there is some issue with using child, especially on the x-axis.

PR #912, opened 2023-01-29
Fix remove all weapons

tyackman:


Aure7138:
I noticed you used fix for the title, did the original one not work or

tyackman:
No, it did not work.

Aure7138:
I remember that it is working for self, and it should be working for NPCs. If you are referring to players, just remind players that they can protect against this event

tyackman:
Yes I am aware, it is for other players who do not block the event. Thank you.

Adler-D3R:
This Native does not work on players. However, removing weapons one by one will work. The fix he provided should be working fine.

Primexz:
Maybe we could check here if we run the function on ourselves or on other players.

SkiddyToast:
Try this, there might be a check for players in the native:

```c++
// E9 ? ? ? ? 48 83 B8 B0 10 00 00 00 .add(1).rip()
using remove_all_weapons_event_trigger_t = void(*)(CPed* pPed);
```

PR #917, opened 2023-01-30
feat(self): add ptfx player effects

Primexz:
idk some ptfx effects for our local player.
i don't know if the whole thing makes sense but here we go.


Seanghost117:
If you don't mind can I add to this / fix this up?

Primexz:
> If you don't mind can I add to this / fix this up?

np

Yimura (dismissed):
Looks good, waiting for @Seanghost117 to add their changes.

Seanghost117:
> Looks good, waiting for @Seanghost117 to add their changes.

Almost done just revising some small bugs and testing

Seanghost117:
The only problem I can see right now, for me personally is just the effect of the name drop-down (the strings including underscores). I don't like the strings being that way but it's not all too serious. If anyone cared we can easily just add friendly names with them or I was working on a little string return that should remove all the characters before the first _, replace all the underscores with spaces, and uppercase the words. The only problem is some assets like "scr_indep_firework_(...)" this does not work, although it can be easily worked around by either checking if there are more than 3 underscores and removing everything before the third or just checking if it matches a specific case such as the one above. It's not all that serious and will be on my to-do list later, it can be merged as is though.

Edit: I see I would also have to format the string to include: Friendly name from return + ## asset name, because some assets would be the same if I proceeded with the idea, case and point being eject clown and muz clown, again not a serious issue but in case I forget

Primexz:
> The only problem I can see right now, for me personally is just the effect of the name drop-down (the strings including underscores). I don't like the strings being that way but it's not all too serious. If anyone cared we can easily just add friendly names with them or I was working on a little string return that should remove all the characters before the first _, replace all the underscores with spaces, and uppercase the words. The only problem is some assets like "scr_indep_firework_(...)" this does not work, although it can be easily worked around by either checking if there are more than 3 underscores and removing everything before the third or just checking if it matches a specific case such as the one above. It's not all that serious and will be on my to-do list later, it can be merged as is though.

Due to the small number of effects, I also agree that the whole thing is currently not yet necessary. Improvements can always be made afterwards. :)


EDIT:
The UI also looks relatively simple to use for users. Very nice improvements @Seanghost117  :=)

image

TheGreenBandit:
personally i would add variability for the user of where they can set the effect, for example have it on the head and hand, have one ptfx on the leg and a different ptfx on the torso, but it looks awesome so far 👍

src/backend/looped/self/ptfx.cpp
```diff
@@ -0,0 +1,39 @@
...
+			PedBones::SKEL_Head,
+			PedBones::SKEL_L_Hand,
+			PedBones::SKEL_R_Hand,
+			PedBones::SKEL_L_Foot,
+			PedBones::SKEL_R_Foot
+		};
+
+        void show_ptfx_effect(const char* fx_name, const char* name)
+        {
+			for (const auto& ptfx_bone : ptfx_bones)
```
Yimura: mixed indentation (tabs and spaces)

Yimura (changes requested):
Overall good code just one mixed indentation.

src/backend/looped/self/ptfx.cpp
Seanghost117: Forgot to fix that when I made my edits, I’ll let @Primexz handle that one, don’t wanna Pr off a Pr again

Primexz: I am currently working on the idea of @Seanghost117 .
I will fix the indents in the commit as well

Yimura:
If CI fails ignore it, that's probably my fault.

Primexz:
PR should be ready.
Friendly Names are not perfect yet bot O.K.

Seanghost117:
Maybe we should hold off on the friendly name idea and merge it as it is before, better to have a polished product then not. Let’s just stick to the original strings maybe if it doesn’t look right especially if more ptfx names get merged in.

Primexz:
> 

We can use the history to work later on that "feature".

EDIT:
Only for completeness:
Other contributors can use [this](https://github.com/DurtyFree/gta-v-data-dumps/blob/master/particleEffectsCompact.json) to add new PTFX effects.

Yimura (approved):

PR #919, opened 2023-01-30
fix(self): fix auto tp to waypoint

Primexz:
With the old implementation we have the problem that the place we are teleported to is not the one we set (in freemode).
I have now used the same function that we use for a button click. This seems to work much better


Yimura (approved):
Sounds good, I agree the old method was sorta hacky.

PR #924, opened 2023-01-31
feat(self): add superman mode

Primexz:
Yet another pretty simple but cool feature.


Yimura (commented):
I really hate GetAsyncKeyState and have been stubborn in the past when features used it.

For now the code is good as-is, I'll test it later if it works well in-game and is intuitive.

Primexz:
> GetAsyncKeyState

Do you know any good alternatives for GetAsyncKeyState?

Yimura:
> Do you know any good alternatives for GetAsyncKeyState?

WndProc, the renderer exposes an add_wndproc_callback function, with that you can easily add a callback to receive keypresses.
Take a look at hotkey service, it makes use of it.

maybegreat48:
Using the `PAD` natives would get rid of `GetAsyncKeyState` and would add controller support at the same time

Primexz:
I can't test the controller support myself, but in theory it should work.

Yimura (approved):

Yimura:
You can maybe force the skydive task the moment the player jumps up from the ground? Could be a bit janky to implement though.

Primexz:
> You can maybe force the skydive task the moment the player jumps up from the ground? Could be a bit janky to implement though.

I also think that would be a bit difficult, I don't see how it could be implemented.
Theoretically users would just have to press S, I think they should be able to do that. :D

Yimura (approved):

PR #925, opened 2023-01-31
feat: add ragequit option + hotkey

Primexz:
I know this feature is a bit pointless, but we all love the classic rage quit.
I think the feature is quite a cool idea even if it is so simple ^^


Yimura (approved):

Seanghost117:
Just as a person looking on, I would just suggest that maybe you pull all your feature additions into one PR sort of like green did with his QOL and add all the changes as per type template 
(Feat) feature 1
(Feat) feature 2
(Fix) fix1
And so forth would just save some of the clutter and reviewers / yimura some time. I’m also not sure this really needs to be a commit u can easily just press alt + f4 or alt tab and close the console window. On top of that it makes more sense to just simple add a hotkey registry to unload the menu rather than just doing all this.

TheGreenBandit:
@Seanghost117 yes, this definitely couldve been added in one of the previous prs considering how little of a commit it is. The thing about my pr though is im still fixing it lmao

Peax:
> I know this feature is a bit pointless

I disagree. It's actually quite useful if you try Criminal Mastermind challenges where you're not allowed to die 😶

PR #928, opened 2023-02-01
Some very simple features

yasmasdas:
Some very simples features
- Disable Phone
- Request Ballistic equipment and Request Boat Pickup
- Keep Vehicle On Ground


src/backend/looped/self/disable_phone.cpp
```diff
@@ -0,0 +1,23 @@
+#include "natives.hpp"
+#include "backend/looped_command.hpp"
+#include "script_global.hpp"
+
+namespace big
+{
+	class keep_on_ground : looped_command
```
Yimura: command is called keep on ground same for the global `g_keep_on_ground`

Yimura (changes requested):

ghost:
Disable phone already exists? What about being able to use the phone but disabling incoming calls?

Aure7138:
I suggest not putting the globals in the command files, but putting it in the util or in one place for easy updating

yasmasdas:
> Disable phone already exists? What about being able to use the phone but disabling incoming calls?

You're right I did not see that and I'll put the globals in a diff file later today i'll draft this pr

PR #933, opened 2023-02-04
feat(protections): block receive pickup

Primexz:
This protection will prevent any pickups such as money drops from being collected.
Since all pickups are basically blocked, normal items can't be picked up either, so I have disabled the option by default.

### DIFF
--- a/src/core/globals.hpp
+++ b/src/core/globals.hpp
@@ -232,9 +232,10 @@ namespace big
 
 			bool desync_kick = false;
 			bool rid_join = false;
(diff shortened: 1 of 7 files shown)


Adler-D3R:
Can't this be blocked in ReceivedNetEvent Hook ? It would be better not to hook too much.

Yimura (approved):
Yeah I don't mind another hook, a few were removed in the in the 1.66 update and the performance penalty isn't noticeable to a human anyway.

PR #938, opened 2023-02-04
added some useful spam_texts in Chinese language

pzy2000:
Provide better defense against Chinese spamers, which should be handy for Asian users.
Also Added some new Chinese vocab in zh_CN.json, added popup info for FORCE_WANTED_LVL function.
Btw at first I forgot to delete some personal modification of codes, which has been corrected afterwards.


Yimura:
I'm going to wait 24h before merging this to ensure the newest version of the translations have been cached.

pzy2000:
> I'm going to wait 24h before merging this to ensure the newest version of the translations have been cached.

roger that, jsdelivr purge limit sucks

Yimura (approved):

PR #951, opened 2023-02-06
feat(world): blackhole

Primexz:
A small feature to have some fun with the entities of GTA

blackhole_1
blackhole_2
blackhole_3


Yimura (approved):
Will still need to test it in-game but looks like a neat feature.

PR #954, opened 2023-02-06
fix(session_browser): decrease session limit

Primexz:
It seems that Rockstar has simply limited the number of sessions we can find.
With the reduced limit, the session browser works perfectly again!

### DIFF
--- a/src/services/matchmaking/matchmaking_service.hpp
+++ b/src/services/matchmaking/matchmaking_service.hpp
@@ -5,7 +5,7 @@ namespace big
 	class matchmaking_service
 	{
 	public:
-		constexpr static int MAX_SESSIONS_TO_FIND = 1400;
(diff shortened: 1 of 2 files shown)


BugisoftRSG:
Still gettings 8447 results for the region europe, maybe you are on CIS Region which has per usual only a few lobbys?

Settings:
GAME_MODE=0
MMATTR_MM_GROUP_2=30
MMATTR_REGION=3

Primexz:
> 

image

BugisoftRSG:
There is only 1 native country in europe that speaks english and that is UK

Primexz:
> There is only 1 native country in europe that speaks english and that is UK

You can also test the whole thing without any filter....
According to my tests, you can not retrieve more than 1057 sessions, otherwise the whole thing will not work.

Yimura (approved):

PR #957, opened 2023-02-06
feat(translation_service): add support for custom gta labels

Primexz:
refs https://github.com/YimMenu/Translations/pull/9


Yimura (commented):
Services are meant to be individual building blocks of the menu and you should normally be able to just remove one of them without another service breaking or having to rely on it. (I get that there are several exceptions to this but still)
You're modifying the translation service to suit the needs for the custom text service but this is a bad practice as the next time you add something that may rely on translation service you need to include it again and modify the service accordingly.

Try maybe making use of the translation service as-is and not use nested objects to achieve what you wanted.
Maybe instead work on something that allows users to input translation hashstrings in a separate file for now?

Also adding these to translations seems like a bad idea, take for example the Off Radar hash it's just meant as a bad joke and leaving such a string up to each language to be translated seems awful.
There's some amount of subjective translating required when someone else is working on translating to a new language.

Initially my idea with it was to allow people to rename garages to more descriptive names instead of "Maze Bank Garage Floor 1" use something like "Muscle Cars".
However I've never gotten the time to implement such a thing.
It's also why I added a callback function for translation hashes that way you could hook externally to the custom text service with callbacks without directly modifying any of the files within the custom text service.

Primexz:
> Services are meant to be individual building blocks of the menu and you should normally be able to just remove one of them without another service breaking or having to rely on it. (I get that there are several exceptions to this but still) You're modifying the translation service to suit the needs for the custom text service but this is a bad practice as the next time you add something that may rely on translation service you need to include it again and modify the service accordingly.
> 
> Try maybe making use of the translation service as-is and not use nested objects to achieve what you wanted. Maybe instead work on something that allows users to input translation hashstrings in a separate file for now?
> 
> Also adding these to translations seems like a bad idea, take for example the Off Radar hash it's just meant as a bad joke and leaving such a string up to each language to be translated seems awful. There's some amount of subjective translating required when someone else is working on translating to a new language.
> 
> Initially my idea with it was to allow people to rename garages to more descriptive names instead of "Maze Bank Garage Floor 1" use something like "Muscle Cars". However I've never gotten the time to implement such a thing. It's also why I added a callback function for translation hashes that way you could hook externally to the custom text service with callbacks without directly modifying any of the files within the custom text service.

Okay thank you for your very very detailed feedback. I have thought about the whole topic and its implementation again. I would have the idea that the Custom Text Service is completely independent from the Translation Service again. My idea was to have a separate JSON where users can enter their custom GTA labels, regardless of the language. (As you already suggested) This would have the "disadvantage" of no more customizability to the language but still the advantage that users can add their own labels without having to compile the menu.

The only problem I see is the storage / management of the GTA labels. Do we want to manage the GTA labels via a repository or something, or do we leave the control over it completely to the user locally?

PR #961, opened 2023-02-08
Send Event Acknowledge pattern updated

sa413x:
Old pattern seems to refs to wrong function.


Yimura (approved):
You are right, I have been having some issues with this. Thanks!

PR #914, opened 2023-01-29
feat: CTaskGeneralSweep crash protection

Aure7138:
[_.log](https://github.com/YimMenu/YimMenu/files/10530512/_.log)

For some reason, passing invalid names to `TASK::TASK_SWEEP_AIM_POSITION` will trigger this crash (Thanks to unnamed204 for the code)

Not the best solution but I just put here

PR #969, opened 2023-02-09
feat: Improved exception handling

Aure7138:
image
image

In the crash of #914, the current exception handling has a probability of causing the process to enter an infinite EXCEPTION_ACCESS_VIOLATION (or whatever that is) function. 

Yes, this is my mistake. It didn’t happen when I tested it. Maybe it’s because I only triggered a few crashes, didn't spam it.

Well, after closing #914, I privately spam the crash and found that I entered an infinite loop function (0x14B51 in the second picture), this pr should fix this problem

### DIFF
--- a/src/logger/exception_handler.cpp
+++ b/src/logger/exception_handler.cpp
@@ -22,13 +22,17 @@ namespace big
             exception_code == DBG_PRINTEXCEPTION_WIDE_C)
             return EXCEPTION_CONTINUE_SEARCH;
 
-        LOG(FATAL) << stack_trace(exception_info);
+        auto stack_trace_instance = std::make_unique<stack_trace>(exception_info);
+        LOG(FATAL) << *stack_trace_instance.get();
 
         ZyanU64 opcode_address = exception_info->ContextRecord->Rip;
         ZydisDisassembledInstruction instruction;
         ZydisDisassembleIntel(ZYDIS_MACHINE_MODE_LONG_64, opcode_address, reinterpret_cast<void*>(opcode_address), 32, &instruction);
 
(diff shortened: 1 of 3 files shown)


src/logger/exception_handler.cpp
```diff
@@ -22,13 +22,17 @@ namespace big
             exception_code == DBG_PRINTEXCEPTION_WIDE_C)
             return EXCEPTION_CONTINUE_SEARCH;
 
-        LOG(FATAL) << stack_trace(exception_info);
+        auto stack_trace_instance = std::make_unique<stack_trace>(exception_info);
+        LOG(FATAL) << *stack_trace_instance.get();
```
Yimura: No need to dereference the pointer, you can just modify or add a new stream function that takes a pointer as an argument.

```cpp
inline std::ostream& operator<< (std::ostream& os, const stack_trace* st)
{
    os << st->str();

    return os;
}
```

Also curious why you changed the stack_trace instance to a pointer, is there a specific reason for this?

Yimura (changes requested):

src/logger/exception_handler.cpp
Aure7138: I am used to create any class using std::make_uniptr, that will make me feel safe

Yimura: Allocating on heap is more expensive, try to allocate on stack as that'll be better (aka. how it was before).
If it causes side effects (like infinitely crashing or something), then you can use this method.

Aure7138: sure

Yimura (approved):

PR #985, opened 2023-02-13
very small protection changes

yasmasdas:
very small protection changes


Yimura (dismissed):
Just wondering are these actual possible crashes or just some sanity checks?

yasmasdas:
I know one of these did cause a crash when set to a higher number. the other one is a bit of a guess. Do you know what that check is about?

Yimura (approved):

PR #993, opened 2023-02-14
chore: fewer sig scan

Aure7138:


Yimura (approved):
Have you cross-referenced old versions of the game to see if the `sub(37)` is the same for older versions of the game?

Aure7138:
yes
These are actually in a function, and I also compared the dumps of 1.63 and 1.66, and the offset is the same

PR #1002, opened 2023-02-16
chore(debug): some additional debug options

Aure7138:


yasmasdas:
Is there a point in litterally any of these features?

Aure7138:
> Is there a point in litterally any of these features?

Makes sense to me, the option to exclude script events when debugging crashes, and the other two allow you to go back to GTA story mode in any state

Yimura (approved):

PR #1019, opened 2023-02-22
fix(handling): fix crash with veto modern

### LINKED ISSUE #1017: [Bug]: YimMenu keeps crashing my game instantly after injecting
### Describe the bug

Once I inject the dll it loads up and after finishing (I belive) it throws an error that I can't take a look at nor copy because right after doing thta it crashes my game.

### Steps To Reproduce

1. Inject the dll

### Expected Behavior

Shouldn't crash

### Screenshots and/or Logs

```
<paste your logs here>
```

### Store

Rockstar

### Additional context

_No response_

### Build

Primexz:
Since we don't use this value anywhere in the UI I think we can overwrite it permanently. 🤷‍♂️
fixes #1017

### DIFF
--- a/src/services/vehicle/handling_profile.cpp
+++ b/src/services/vehicle/handling_profile.cpp
@@ -7,8 +7,7 @@ namespace big
 		m_gravity = vehicle->m_gravity;
 		m_handling_data = *vehicle->m_handling_data;
 
-		//this seems to be not set on vehicle: "Veto Modern"
-		if(&m_handling_data.m_traction_spring_delta_max_ratio == nullptr)
+		if(std::isinf(m_handling_data.m_traction_spring_delta_max_ratio))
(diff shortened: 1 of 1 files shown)


Yimura (changes requested):
This should not be the fix to this problem.

You're also overriding it for ALL vehicles even if they have valid values.
There should be a different underlying problem here, as the memory should be 0 at least if they don't use it. (it's not like this is an invalid pointer or something)

Did you verify in memory that value is correct?

Primexz:
> e a different underlying problem here, as the memory should be 0 at least if they don't use it. (it's not like this is an invalid pointer or something)
> 
> Did you verify in memory that value is correct?

Edit: Works now with the new commit ^^

image

Yimura (approved):

PR #1020, opened 2023-02-22
Add more spoofing options

maybegreat48:
* Added blip type spoofing
* Added rank spoofing
* Added K/D ratio spoofing
* Added badsport spoofing
* Added job points spoofing (buggy, may apply to other players)
* Added player model spoofing
* Fixed menu scaling

Fixes #1009 
Fixes #976 
Fixes #876 
Fixes #874 
Possibly _fixes_ #754 

Marking as draft due to the addition of a clang format file. Feel free to suggest possible improvements to it

PR #1029, opened 2023-02-27
feat: command executor (search bar)

Primexz:
Since the original Branch: #850 is apparently dead, I thought I'd take a look at the feature.

Here is my current approach:
https://user-images.githubusercontent.com/76482511/221678598-2e62de3f-5338-445c-9ac5-826c56cccd7a.mp4

## ToDo
- [x] proper rendering of the window
- [x] suggestions
- [x] execute incomplete commands (best suggestion)
- [x] code cleanup
- [x] translations
- [x] better management of hotkeys (when closing the executor the key in the input field is still received)
- [x] disable notifications when opening the cmd executor
- [x] wait for #1020

Feel free to suggest possible improvements to it, thx!


Yimura:
I like the design so far! Gonna be interesting to see how this will change the way I use the menu.

Primexz:
The PR is ready for now.

The closing of the search bar is still extremely bad and should be improved.
If you have suggestions on how I can make the closing of the command executor better, please write a review!

src/backend/command.cpp
```diff
@@ -99,17 +99,26 @@ namespace big
...
 
-			if (command->get_name().contains(search))
+			std::string cmd_name     = command->get_name();
+			std::string cmd_label    = command->get_label();
+			std::string lower_search = search;
+
+			//transform all strings to lower case
+			std::transform(cmd_name.begin(), cmd_name.end(), cmd_name.begin(), tolower);
+			std::transform(cmd_label.begin(), cmd_label.end(), cmd_label.begin(), tolower);
+			std::transform(lower_search.begin(), lower_search.end(), lower_search.begin(), tolower);
```
Yimura: Search can probably be made lower case outside of the for loop unless I'm missing something.

Yimura (changes requested):

src/backend/command.cpp
Primexz: Yep, that was my mistake.
I fixed it in the last commit

src/gui.cpp
```diff
@@ -161,6 +163,21 @@ namespace big
 			PAD::DISABLE_CONTROL_ACTION(2, 262, true);
 			PAD::DISABLE_CONTROL_ACTION(2, 331, true);
 		}
+
+		//wndproc will not work here. the timing here is very difficult. mayby can we hook the creation of the pause menu?
+		//this should be improved..
```
maybegreat48, line 168: Have you tried calling `DISABLE_FRONTEND_THIS_FRAME` every tick the cmd executor is active? `DISABLE_CONTROL_ACTION` could work too

Yimura: Yes this piece of code is really janky, ideally replace it with something else.

Primexz: I had wasted a lot of time on it. :/
Running the whole thing while the Command Executor is open didn't work.

Yimura (approved):

PR #1038, opened 2023-03-02
NPC´s ignores player

grztx:
NPC´s ignore player´s actions (only for nearby peds) and should work for online too


Primexz:
I think you have done a double job. :/

It seems that we already have this feature.
https://github.com/YimMenu/YimMenu/blob/master/src/backend/looped/world/nearby/ignore.cpp

If there is something wrong with the original feature it would be better to improve it. ;)

Seanghost117:
I've already added this, we also have a function to iterate through peds and vehicles there is no reason to use this method

PR #1039, opened 2023-03-02
feat(translations): add support for GUI tabs

Yimura:
This should add support for translations of GUI tabs, the GUI is terrible so this is a quick implementation.

Waiting for changes on https://github.com/YimMenu/Translations/pull/13 to propagate to the CDN.


maybegreat48:
You should be able to do this for commands too

Yimura:
> You should be able to do this for commands too

Yup will do that later on, don't want to do too large of a change and then not spend the time to finish it.

PR #1060, opened 2023-03-10
feat(vehicle): infinite ammo

JDMJay:


JDMJay:
confused

yasmasdas:
Can this not be done through the vehicles weapon info somehow? it 
CPed -> CPedWeaponManager *m_weapon_manager; //0x10B8 -> CWeaponInfo* m_vehicle_weapon_info; //0x0070
Saves the need for hashes

or as alternative you can even read CVehicleWeaponHandlingData atArray of Weapon hashes  (uWeaponHash 0x0008) to get the vehicles hashes

yasmasdas:
because there are way more and I mean WAY more vehicle weapon hashes then 12 there like a 100

binded-zz:
is there a way to add range to vehicle as well
there is a new range for weapons but it does not seem to work for vehicles

is there also away to turn back homing rockets when some one turned them off

Yimura (changes requested):
Try to see if you can make it work dynamically from the weapon list we parsed from the game.

Having a hardcoded array is a pain to maintain over time.

yasmasdas:
> Try to see if you can make it work dynamically from the weapon list we parsed from the game.

@Yimura I don't think using the weapon list is a good idea when you can read the vehicless weapon hashes directly https://alexguirre.github.io/rage-parser-dumps/dump.html?game=gta5&build=2824#CVehicleWeaponHandlingData

Yimura:
> @Yimura I don't think using the weapon list is a good idea when you can read the vehicless weapon hashes directly https://alexguirre.github.io/rage-parser-dumps/dump.html?game=gta5&build=2824#CVehicleWeaponHandlingData

You're correct, I was completely blinded by the fact that he used natives. 
Using classes is way better imo

JDMJay:
I'll use classes, I'll get started with updating my prs tommorow I've not had much time

Yimura:
Stale PR

PR #1090, opened 2023-03-14
Prevent exploit reports

birds1338:
Exploit reports for god mode and invisibility are sent in shop_controller, there is only one occurrence of the native being called in this script so it is safe to patch it this way.

### DIFF
--- a/src/native_hooks/native_hooks.cpp
+++ b/src/native_hooks/native_hooks.cpp
@@ -112,6 +112,8 @@ namespace big
 		add_native_detour(0xFE99B66D079CF6BC, all_scripts::DISABLE_CONTROL_ACTION);
 		add_native_detour(0xEB354E5376BC81A7, all_scripts::HUD_FORCE_WEAPON_WHEEL);
 		add_native_detour(0x158C16F5E4CF41F8, all_scripts::RETURN_TRUE); //bypass casino country restrictions
+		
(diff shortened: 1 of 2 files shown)


src/native_hooks/native_hooks.cpp
```diff
@@ -112,6 +112,8 @@ namespace big
 		add_native_detour(0xFE99B66D079CF6BC, all_scripts::DISABLE_CONTROL_ACTION);
 		add_native_detour(0xEB354E5376BC81A7, all_scripts::HUD_FORCE_WEAPON_WHEEL);
 		add_native_detour(0x158C16F5E4CF41F8, all_scripts::RETURN_TRUE); //bypass casino country restrictions
+		
+		add_native_detour(RAGE_JOAAT("shop_controller"), 0x34616828CD07F1A1, all_scripts::RETURN_FALSE); // prevent exploit reports
```
Aure7138, line 116: IS_PED_SHOOTING?
image

birds1338: shop_contoller...

Yimura:
This is already patched by YimMenu.

https://github.com/YimMenu/YimMenu/blob/d87caaa6d737dd9aacc1ec87220c37992236e2a3/src/backend/script_patches.hpp#L33

PR #1091, opened 2023-03-14
feat: prevent exploit reports

birds1338:
Your current patch is not working...

### DIFF
--- a/src/native_hooks/native_hooks.cpp
+++ b/src/native_hooks/native_hooks.cpp
@@ -112,6 +112,8 @@ namespace big
 		add_native_detour(0xFE99B66D079CF6BC, all_scripts::DISABLE_CONTROL_ACTION);
 		add_native_detour(0xEB354E5376BC81A7, all_scripts::HUD_FORCE_WEAPON_WHEEL);
 		add_native_detour(0x158C16F5E4CF41F8, all_scripts::RETURN_TRUE); //bypass casino country restrictions
+		
+		add_native_detour(RAGE_JOAAT("shop_controller"), 0x34616828CD07F1A1, all_scripts::RETURN_FALSE); // prevent exploit reports
 
 		add_native_detour(RAGE_JOAAT("carmod_shop"), 0x06843DA7060A026B, carmod_shop::SET_ENTITY_COORDS);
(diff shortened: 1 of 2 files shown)


src/native_hooks/native_hooks.cpp
```diff
@@ -112,6 +112,8 @@ namespace big
 		add_native_detour(0xFE99B66D079CF6BC, all_scripts::DISABLE_CONTROL_ACTION);
 		add_native_detour(0xEB354E5376BC81A7, all_scripts::HUD_FORCE_WEAPON_WHEEL);
 		add_native_detour(0x158C16F5E4CF41F8, all_scripts::RETURN_TRUE); //bypass casino country restrictions
+		
+		add_native_detour(RAGE_JOAAT("shop_controller"), 0x34616828CD07F1A1, all_scripts::RETURN_FALSE); // prevent exploit reports
```
Yimura, line 116: Any reason you're blocking [`PED::IS_PED_SHOOTING`](https://natives.gta.menu/natives/0x34616828CD07F1A1)?

Yimura: I'm guessing that is the condition before it checks for invisibility and invincibility.

birds1338: YES! Look at the script please. It is checking if you are invincible or invisible when shooting,
image

Yimura: How do you know it's not working?

birds1338: Because I am using it and getting the SERVICE_EARN_JBONUS_NO_DEATH transaction. Worst case scenario: you merge this and have an extra assurance that it is patched.

Yimura: Yes, I closed your last PR because I was under the assumption you weren't aware of the existing protection in the menu, neither did you mention this was a fix.

All I saw was just another protection for something that was already implemented, my bad.
I'll confirm it with others that the other project is no longer working and we can just remove it then.

maybegreat48: Can confirm that the patch is incorrect and does not work. This should work

Yimura (approved):

PR #1124, opened 2023-03-20
Hotfix outfit random component crash

### LINKED ISSUE #1118: [Bug]: random outfit crash
### Describe the bug

random outfit crash

### Steps To Reproduce

outfit editor
random

### Expected Behavior

no crash

### Screenshots and/or Logs

```
<paste your logs here>
```

### Store

Steam

### Additional context

_No response_

### Build

Aure7138:
Reimplement SET_PED_RANDOM_COMPONENT_VARIATION with SET_PED_COMPONENT_VARIATION

fixes #1118

The original random component crashes after about 20 clicks, this fix I tested works fine for me

### DIFF
--- a/src/views/settings/view_outfit_editor.cpp
+++ b/src/views/settings/view_outfit_editor.cpp
@@ -4,6 +4,12 @@
 
 namespace big
 {
+	static int range(int lower_bound, int upper_bound)
+	{
+		std::srand(std::chrono::system_clock::now().time_since_epoch().count());
+		return rand() % (upper_bound - lower_bound + 1) + lower_bound;
+	}
+
 	void view::outfit_editor()
 	{
 		struct outfit_t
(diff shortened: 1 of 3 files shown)


src/views/settings/view_outfit_editor.cpp
```diff
@@ -4,6 +4,12 @@
 
 namespace big
 {
+	static int range(int lower_bound, int upper_bound)
+	{
+		std::srand(std::chrono::system_clock::now().time_since_epoch().count());
```
Yimura: Setting the seed every time you call it?

image

src/views/settings/view_outfit_editor.cpp
```diff
@@ -4,6 +4,12 @@
...
+		return rand() % (upper_bound - lower_bound + 1) + lower_bound;
```
Yimura: You're settings the seed for the `std` version of rand and then here you're using the C version of `rand`.

Use `std::rand` instead of `::rand`.

Yimura (changes requested):

src/views/settings/view_outfit_editor.cpp
Aure7138: you're right, changed

src/views/settings/view_outfit_editor.cpp
Aure7138: Ah, yeah, I moved it into main

Yimura (approved):

PR #1139, opened 2023-03-24
Remove Recovery File

Rxann:
I noticed that there is a file for recovery that seems to not be used and violates the contribution policy.
so yeah this pr pretty much removes the file.


Seanghost117:
Although it may seem that this has no purpose, when someone in your lobby tried to use a recovery feature via the chat command it still explains in chat or to the user that it's not supported and provides an alternative.

Yimura:
Yes, it will stay in YimMenu to notify other people in the lobby to use other tools if they need recovery options.

PR #1148, opened 2023-03-26
Orbital Drone Feature

Tayipc38:
This feature is simply a version of the Orbital cannon in the Facility. Currently it fails to create an explosion the size of the Orbital cannon due to limitations put by the game. If anyone knows how to patch the check please don't be afraid to tell me. Some effort regarding the patch is already in the code yet I can't seem to get it to work.

### DIFF
--- a/src/backend/backend.cpp
+++ b/src/backend/backend.cpp
@@ -6,6 +6,7 @@
 #include "script_patches.hpp"
 #include "services/context_menu/context_menu_service.hpp"
 #include "thread_pool.hpp"
+#include "services/orbital_drone/orbital_drone.h"
 
 namespace big
 {
@@ -183,6 +184,30 @@ namespace big
 		while (g_running)
 		{
 			looped::world_spawn_ped();
+			script::get_current()->yield();
+		}
+	}
+
+	void backend::orbital_drone()
+	{
+		while (true)
+		{
+			if (orbitaldrone::orbital_drone && PAD::IS_CONTROL_JUST_PRESSED(2, (int)ControllerInputs::INPUT_VEH_LOOK_BEHIND))
+			{
+				g_fiber_pool->queue_job([=] {
+					if (!orbitaldrone::g_orbital_drone.initialized)
+						g_fiber_pool->queue_job([=] {
+							orbitaldrone::g_orbital_drone.initialize();
+						});
+					else
+						g_fiber_pool->queue_job([=] {
+							orbitaldrone::g_orbital_drone.~OrbitalDrone();
+						});
+				});
+			}
+
+			orbitaldrone::g_orbital_drone.tick();
+
 			script::get_current()->yield();
 		}
 	}
--- a/src/backend/backend.cpp
+++ b/src/backend/backend.cpp
@@ -6,7 +6,7 @@
 #include "script_patches.hpp"
 #include "services/context_menu/context_menu_service.hpp"
 #include "thread_pool.hpp"
-#include "services/orbital_drone/orbital_drone.h"
+#include "services/orbital_drone/orbital_drone.hpp"
 
 namespace big
 {
@@ -192,21 +192,15 @@ namespace big
 	{
 		while (true)
 		{
-			if (orbitaldrone::orbital_drone && PAD::IS_CONTROL_JUST_PRESSED(2, (int)ControllerInputs::INPUT_VEH_LOOK_BEHIND))
+			if (orbital_drone_t && PAD::IS_CONTROL_JUST_PRESSED(2, (int)ControllerInputs::INPUT_VEH_LOOK_BEHIND))
 			{
-				g_fiber_pool->queue_job([=] {
-					if (!orbitaldrone::g_orbital_drone.initialized)
-						g_fiber_pool->queue_job([=] {
-							orbitaldrone::g_orbital_drone.initialize();
-						});
-					else
-						g_fiber_pool->queue_job([=] {
-							orbitaldrone::g_orbital_drone.~OrbitalDrone();
-						});
-				});
+				if (!g_orbital_drone_service.initialized)
+					g_orbital_drone_service.initialize();
+				else
+					g_orbital_drone_service.destruct();
 			}
 
-			orbitaldrone::g_orbital_drone.tick();
+			g_orbital_drone_service.tick();
 
 			script::get_current()->yield();
 		}
--- a/src/backend/backend.cpp
+++ b/src/backend/backend.cpp
@@ -5,8 +5,9 @@
 #include "script.hpp"
 #include "script_patches.hpp"
 #include "services/context_menu/context_menu_service.hpp"
-#include "thread_pool.hpp"
 #include "services/orbital_drone/orbital_drone.hpp"
+#include "thread_pool.hpp"
+
 
 namespace big
 {
@@ -192,12 +193,12 @@ namespace big
 	{
 		while (true)
 		{
-			if (orbital_drone_t && PAD::IS_CONTROL_JUST_PRESSED(2, (int)ControllerInputs::INPUT_VEH_LOOK_BEHIND))
+			if (g.world.orbital_drone.enabled && PAD::IS_CONTROL_JUST_PRESSED(2, (int)ControllerInputs::INPUT_VEH_LOOK_BEHIND))
 			{
-				if (!g_orbital_drone_service.initialized)
-					g_orbital_drone_service.initialize();
+				if (!g_orbital_drone_service.initialized())
+					g_orbital_drone_service.init();
 				else
-					g_orbital_drone_service.destruct();
+					g_orbital_drone_service.destroy();
 			}
 
 			g_orbital_drone_service.tick();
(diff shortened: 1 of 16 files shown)


src/services/orbital_drone/orbital_drone.h
```diff
@@ -0,0 +1,123 @@
...
+#include "script.hpp"
+#include "gta/enums.hpp"
+#include "fiber_pool.hpp"
+
+
+namespace big::orbitaldrone
+{
+	inline bool orbital_drone;
+
+	class OrbitalDrone
```
maybegreat48: Class names should be in snake_case. The header should have `hpp` as the extension

src/util/toxic.hpp
```diff
@@ -20,8 +20,10 @@ namespace big::toxic
 	inline void blame_explode_coord(player_ptr to_blame, Vector3 pos, eExplosionTag explosion_type, float damage, bool is_audible, bool is_invisible, float camera_shake)
 	{
 		g_pointers->m_blame_explode->apply();
+		//g_pointers->m_explosion_patch->apply();
```
maybegreat48: Dead code. I assume it allows orbital cannon explosions from non OC scripts

src/views/self/view_self.cpp
```diff
@@ -61,6 +62,7 @@ namespace big
 		components::command_checkbox<"superman">();
 
 		ImGui::Checkbox("DANCE_MODE"_T.data(), &g.self.dance_mode);
+		ImGui::Checkbox("Orbital drone", &orbitaldrone::orbital_drone);
```
maybegreat48: I feel that this should be in a different category, like weapons or world

src/backend/backend.cpp
```diff
@@ -186,4 +187,28 @@ namespace big
...
+			if (orbitaldrone::orbital_drone && PAD::IS_CONTROL_JUST_PRESSED(2, (int)ControllerInputs::INPUT_VEH_LOOK_BEHIND))
+			{
+				g_fiber_pool->queue_job([=] {
+					if (!orbitaldrone::g_orbital_drone.initialized)
+						g_fiber_pool->queue_job([=] {
+							orbitaldrone::g_orbital_drone.initialize();
+						});
+					else
+						g_fiber_pool->queue_job([=] {
+							orbitaldrone::g_orbital_drone.~OrbitalDrone();
```
maybegreat48: Manually invoking the destructor is probably not a good idea

src/services/orbital_drone/orbital_drone.cpp
```diff
@@ -0,0 +1,380 @@
...
+#include "pointers.hpp"
+#include "network/CNetworkPlayerMgr.hpp"
+#include "orbital_drone.h"
+#include "vehicle/CVehicle.hpp"
+#include "services/players/player_service.hpp"
+#include "util/entity.hpp"
+#include "util/toxic.hpp"
+#include "gui.hpp"
+
+namespace big::orbitaldrone
```
maybegreat48: Services shouldn't be in their own namespace

src/services/orbital_drone/orbital_drone.cpp
```diff
@@ -0,0 +1,380 @@
...
+{
+	void orbital_cannon_esp()
```
maybegreat48: This should be inside the class (same goes for `detect_player`)

src/services/orbital_drone/orbital_drone.cpp
```diff
@@ -0,0 +1,380 @@
...
+	{
+		for (auto p : (*g_pointers->m_network_player_mgr)->m_player_list)
```
maybegreat48: You could use the player service here

src/services/orbital_drone/orbital_drone.h
```diff
@@ -0,0 +1,123 @@
...
+
+			tp = false;
+		}
+
+	private:
+		void cam_nav();
+		void orbital_cannon_explosion();
+	};
+
+	inline OrbitalDrone g_orbital_drone;
```
maybegreat48: Rename to `g_orbital_drone_service` to be consistent with the other services

maybegreat48 (commented):
You may want to set `eOrbitalBitset::kOrbitalCannonActive` in the local player's [`GlobalPlayerBDEntry`](https://github.com/Yimura/GTAV-Classes/blob/master/script/globals/GlobalPlayerBD.hpp#L640) so the players you kill will see the "OBLITERATED" message and get sent to the hospital

maybegreat48:
> Currently it fails to create an explosion the size of the Orbital cannon due to limitations put by the game

Try hooking `48 89 5C 24 10 57 48 83 EC 20 8B D9 E8` and always return true. It shouldn't be integrity checked but you can convert it into a patch (enable when required) if you are worried. `0x90C301B0` should work

src/services/orbital_drone/orbital_drone.cpp
Tayipc38: The esp will not be used, copied it alongside all the other old code I guess. So ill just remove it entirely.

src/util/toxic.hpp
Tayipc38: I have decided to stick with patching this particular if check.

Tayipc38:
> > Currently it fails to create an explosion the size of the Orbital cannon due to limitations put by the game
> 
> Try hooking `48 89 5C 24 10 57 48 83 EC 20 8B D9 E8` and always return true. It shouldn't be integrity checked but you can convert it into a patch (enable when required) if you are worried. `0x90C301B0` should work

Apparently (tnx to @Yimura) the pointer was instructed to cast uint8_t when it should have been uint16_t. Changing this fixed the byte patching and allowed the feature to spawn the orbital cannon explosion.

Tayipc38:
> You may want to set `eOrbitalBitset::kOrbitalCannonActive` in the local player's [`GlobalPlayerBDEntry`](https://github.com/Yimura/GTAV-Classes/blob/master/script/globals/GlobalPlayerBD.hpp#L640) so the players you kill will see the "OBLITERATED" message and get sent to the hospital

This does work, however the feature is equipped to blame other players in which case setting the global on the local player does little to nothing. I don't know whether this global can be set on other players.

src/services/orbital_drone/orbital_drone.cpp
```diff
@@ -0,0 +1,420 @@
...
+		if (m_initialized)
+		{
+			destroy();
+			return;
+		}
+
+		if (*g_pointers->m_is_session_started && gta_util::get_network_player_mgr()->m_local_net_player
+		    && gta_util::get_network_player_mgr()->m_local_net_player->m_player_info->m_ped)
+		{
+			scr_globals::globalplayer_bd.as<GlobalPlayerBD*>()->Entries[PLAYER::PLAYER_ID()].OrbitalBitset.Set(eOrbitalBitset::kOrbitalCannonActive);
```
maybegreat48: You should use `self::id` instead of the native

src/services/orbital_drone/orbital_drone.cpp
```diff
@@ -0,0 +1,420 @@
...
+		}
+	}
+
+	void orbital_drone::tick()
+	{
+		if (!m_initialized)
+			return;
+
+		PAD::ALLOW_ALTERNATIVE_SCRIPT_CONTROLS_LAYOUT(2);
+		PAD::DISABLE_CONTROL_ACTION(0, 24, true);
```
maybegreat48: The `ControllerInputs` enum should be used here too

src/services/orbital_drone/orbital_drone.cpp
```diff
@@ -0,0 +1,420 @@
...
+		HUD::HIDE_HUD_COMPONENT_THIS_FRAME(19);
+		HUD::HIDE_HUD_COMPONENT_THIS_FRAME(2);
+		HUD::HIDE_HUD_COMPONENT_THIS_FRAME(1);
+		HUD::HIDE_HUD_COMPONENT_THIS_FRAME(3);
+		HUD::HIDE_HUD_COMPONENT_THIS_FRAME(4);
+
+		Vector3 cam_pos = CAM::GET_CAM_COORD(m_cam);
+		Vector3 cam_rot = CAM::GET_CAM_ROT(m_cam, 0);
+
+		Entity self = PLAYER::PLAYER_PED_ID();
```
maybegreat48: You should use `self::ped` and `self::veh`

maybegreat48 (commented):
This is a pretty good feature
1. Appears to conflict with noclip if enabled
2. Have you tried using `SET_FAKE_PAUSEMAP_PLAYER_POSITION_THIS_FRAME`?

Tayipc38:
> This is a pretty good feature

Tnx! Means a lot coming from you.

> 1. Appears to conflict with noclip if enabled
> 2. Have you tried using `SET_FAKE_PAUSEMAP_PLAYER_POSITION_THIS_FRAME`?

We have also noticed navigating up and left interferes with no clip when enabled.
I will try the native you suggested.

Tayipc38:
Since NoClip seems to conflict with Orbital drone, ive added a check in the noclip.cpp file that will disable its function when the drone is initialized. This does mean that the Orbital drone takes precedence over NoClip.

Yimura (approved):

PR #1154, opened 2023-03-28
More context menu operations

Tayipc38:
The context menu seems to be somewhat troubled. Ranging from lack of options to outright buggy code. For now I simply added some more operations to its tables;

Vehicle table:
-Copy vehicle
-Boost
-Launch
-Eject

Ped table:
-Disarm
-Ragdoll
-Dance

Player table:
-Breakup kick
-Kick
-Disarm
-Ragdoll

Shared table:
-Enflame

These additions required some other tools and fixes such as:
-Remove weapons command fix.
-Ped animation player tool in ped.hpp
-get player from ped tool in ped.hpp.
-A fiber pool queue on execution of Context menu operations.

WEAPON::REMOVE_ALL_PED_WEAPONS seems to fail on Player peds. Instead ive opted to use REMOVE_WEAPON_FROM_PED which seems to apply even to Player peds.

Ive no clue how to use Commands so I just did what worked. Feel free to correct me.
image

### DIFF
--- a/src/util/ped.hpp
+++ b/src/util/ped.hpp
@@ -3,6 +3,7 @@
 #include "natives.hpp"
 #include "pointers.hpp"
 #include "outfit.hpp"
+#include "services/players/player_service.hpp"
 
 namespace big::ped
 {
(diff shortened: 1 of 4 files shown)


src/util/ped.hpp
```diff
@@ -110,4 +111,38 @@ namespace big::ped
...
+	{
+		for (auto& p : g_player_service->players())
+		{
+			if (p.second->get_ped())
+			{
+				if (p.second->get_ped() == g_pointers->m_handle_to_ptr(ped))
+					return p.second;
+			}
+		}
+		return NULL;
```
Yimura: return `nullptr` instead of `NULL`

Yimura (changes requested):

Yimura (approved):

PR #1156, opened 2023-03-29
feat(cache): expand ped

Aure7138:
The current two caching methods cannot read componentpeds_a_c.rpf, this pr adds a third caching method - cache on init and expand the number of peds on this basis

### DIFF
--- a/src/services/gta_data/gta_data_service.cpp
+++ b/src/services/gta_data/gta_data_service.cpp
@@ -9,6 +9,8 @@
 #include "thread_pool.hpp"
 #include "util/session.hpp"
 #include "yim_fipackfile.hpp"
+#include "util/vehicle.hpp"
+#include "util/misc.hpp"
 
 namespace big
 {
@@ -50,6 +52,11 @@ namespace big
 		return m_update_state;
 	}
 
+	void gta_data_service::set_state(eGtaDataUpdateState state)
+	{
+		m_update_state = state;
+	}
+
 	void gta_data_service::update_in_online()
 	{
 		m_update_state = eGtaDataUpdateState::WAITING_FOR_SINGLE_PLAYER;
@@ -66,7 +73,7 @@ namespace big
 			{
 				script::get_current()->yield(100ms);
 			}
-
+			m_update_state = eGtaDataUpdateState::UPDATING;
 			rebuild_cache();
 		});
 	}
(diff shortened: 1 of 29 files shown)


Aure7138:
Currently finished adding a third caching method, everything is fine, the next step is to expand ped

image
image

If anyone has any suggestions please let me know

xiaoxiao921:
long time since I wrote this but isnt hooking GetDevice better when injecting before game load?

xiaoxiao921:
The whole point is that you don't have to use any yim packfile related functions and just utilize the fact that game is already reading all the files from the rpf as the game load, then you just profit off that

oh yeah another issue: the game doesnt load all dlc rpfs when loading in SP, meaning that they have to go to MP directly

Aure7138:
> long time since I wrote this but isnt hooking GetDevice better when injecting before game load?

will try that one later

> The whole point is that you don't have to use any yim packfile related functions and just utilize the fact that game is already reading all the files from the rpf as the game load, then you just profit off that
> 
> oh yeah another issue: the game doesnt load all dlc rpfs when loading in SP, meaning that they have to go to MP directly

Ah, I see, the hook currently still need yim packfile, to do this probably need hook other functions, I'll try
By the way, I tried SP and MP, the amount seems to be the same, because the hook is special?

image
Get 1019 peds, need a method to get ped type from CPedModelInfo

src/services/gta_data/gta_data_service.cpp
```diff
@@ -459,10 +456,16 @@ namespace big
 			}
 			for (auto it = peds.begin(); it != peds.end();)
 			{
-				if (!model_info::does_model_exist(it->m_hash))
-					peds.erase(it);
-				else
+				if (CPedModelInfo* info = model_info::get_model<CPedModelInfo*>(it->m_hash))
+				{
+					static std::array<std::string, 30> ped_types = {"PLAYER_0", "PLAYER_1", "NETWORK_PLAYER", "PLAYER_2", "CIVMALE", "CIVFEMALE", "COP", "GANG_ALBANIAN", "GANG_BIKER_1", "GANG_BIKER_2", "GANG_BIKER_2", "GANG_RUSSIAN", "GANG_RUSSIAN_2", "GANG_RUSSIAN_2", "GANG_JAMAICAN", "GANG_AFRICAN_AMERICAN", "GANG_KOREAN", "GANG_CHINESE_JAPANESE", "GANG_PUERTO_RICAN", "DEALER", "MEDIC", "FIREMAN", "CRIMINAL", "BUM", "PROSTITUTE", "SPECIAL", "MISSION", "SWAT", "ANIMAL", "ARMY"};
+					std::strncpy(it->m_ped_type, ped_types[*(uint32_t*)((uintptr_t)info + 0x138)].c_str(), sizeof(it->m_ped_type));
```
Yimura: Please don't use a static offset, modify the class on GTAV-Classes repo.

Yimura (changes requested):

Yimura (approved):

PR #1159, opened 2023-03-29
style: GUI Improvements

T14D3:
Multiple GUI Improvements, mainly offering more color customization options

### DIFF
--- a/src/core/globals.hpp
+++ b/src/core/globals.hpp
@@ -395,8 +395,9 @@ namespace big
 				int fast_quit            = 0;
 				int cmd_excecutor        = 0x55;
 				int repairpv             = 0;
+				int context              = 0;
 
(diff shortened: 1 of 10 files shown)


T14D3:
(Just as a note, am still very much new to github stuffs, so if I messed anything up please tell me lol)

Seanghost117:
I think you have the right idea but wrong approach. Firstly and foremost I would really advise you not use the same color variable for a wide section of elements, for example changing all the background colors for items to the same thing or the hovered color, active, etc. all you are doing is meshing colors, at that point you are best loading imgui's default color editor that gives you the option of changing the color of every single items separately and individually and simply just saving and loading the color selection.

T14D3:
I know using the same color for different things isn't ideal, but firstly I wanted to not clutter the settings GUI with too many options and secondly I thrw most of the colors that already were the same into one "group"

Will have to look into the ImGui color editor though

PR #1175, opened 2023-04-02
Vehicle control & Summon

### LINKED ISSUE #1114: [Request]: Remote Vehicle Control (not what you think)
### Problem

1) GTA:O offers some basic remote vehicle control functions over personal vehicle such as opening doors, turning on and off lights, engine and radio.
2) Some not so online options, that I know of, add to this with abilities to open/close windows, toggle roof for applicable vehicles, toggle high beams, car alarm, indicator lights, etc.

Sadly we do not have this control over spawned vehicles.

### Solution

Introduce mentioned features to the menu.
Would be nice if we could have check for last entered vehicle, which would then be target of the controls and pull up "standalone" window with features via keybind.

### Reason

Personal vehicle controls are limited.
You have to purchase the vehicle.
Non toxic, fun feature for anyone who likes to play around with cars.

### Additional context

_No response_

### LINKED ISSUE #1168: [Request]: Summon Vehicle
### Problem

Not a problem, just a feature

### Solution

A feature that operates similar to Tesla's summon feature. You press a button and your car drives to you.

### Reason

This would be very fun to have

### Additional context

_No response_

Tayipc38:
I made this feature at the request of the following issues; #1114 and #1168. The feature is simple and basic as of now. The main premise is to imitate Rockstars remote vehicle control menu for all cars (Not only personal vehicles). Vehicle related operations can be executed both within and outside the car. Such operations consists of but are not limited to; locking doors, opening/closing doors, activating/deactivating lights, operating convertible roofs and several more. Per request the window is standalone and not part of the main UI. It can be activated by a hotkey set in the Hotkeys settings or just in the Vehicle UI. I have imitated the scripts as much as I could with features such as the animations and sounds that play.

Some pictures:
image
image

Current problems:

~~-Since the menu is made with ImGui, the cursor is required to make us of it. However, the cursor is only activated when the main UI is on limiting the users screen when using the feature.
-For some reason the summon feature won't work on Personal Vehicles. The Invisible driver just wont execute its task. Any help on this one is appreciated.~~


Tayipc38:
After some troubleshooting I have discovered that either the scripts or some other loop is making the invisible driver exit the vehicle on the summon feature if vehicle access is set to 'Passengers' in the interaction menu. I have looked through the scripts to find if the access is synced with a global to make a little work-around but was unsuccessful.

src/gta/enums.hpp
```diff
@@ -1447,3 +1447,873 @@ enum class PedBones : std::uint32_t
 	FB_R_Lip_Bot_001                 = 0xC33B,
 	FB_Tongue_001                    = 0xB987
 };
+
```
maybegreat48, line 1450: Adding leaked source code to YimMenu is not a good idea. The menu is under constant observation by Rockstar and P2Cs that may want to be rid of us

You can obviously use the leaks to develop stuff, but try not to make it obvious

src/views/view_vehicle_control.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+		if (ImGui::Checkbox(g_vehicle_control_service.m_controlled_vehicle.engine ? "Stop" : "Start", &g_vehicle_control_service.m_controlled_vehicle.engine))
+		{
+			g_fiber_pool->queue_job([=] {
+				if (g.window.vehicle_control.operation_animation)
+					g_vehicle_control_service.animated_vehicle_operation(self::ped);
+
+				if (entity::take_control_of(g_vehicle_control_service.m_controlled_vehicle.handle))
+					VEHICLE::SET_VEHICLE_ENGINE_ON(g_vehicle_control_service.m_controlled_vehicle.handle,
+				    g_vehicle_control_service.m_controlled_vehicle.engine ? false : true,
```
maybegreat48: `!g_vehicle_control_service.m_controlled_vehicle.engine`

src/views/debug/view_debug_misc.cpp
```diff
@@ -65,6 +67,16 @@ namespace big
 				NETWORK::SHUTDOWN_AND_LOAD_MOST_RECENT_SAVE();
 			});
 
+			components::button("Tp to safe pos", [] {
```
maybegreat48: All interactive UI elements should use Title Case for consistency. So it should be `TP To Safe Pos`

src/services/vehicle/vehicle_control_service.hpp
```diff
@@ -0,0 +1,74 @@
...
+		void driver_tick();
+		bool ensure_driver();
+		void render_distance_on_vehicle();
+		Vector3 m_destination;
+		Ped m_driver;
+
+
+	public:
+		controlled_vehicle m_controlled_vehicle;
+		std::map<int, std::string_view> radiostations;
```
maybegreat48: This isn't used anywhere

Tayipc38: It was meant for a future feature which I laughably failed to implement

src/views/view_vehicle_control.cpp
Tayipc38: I have had minor issues regarding similar code. Common sense would dictate that; 
If (action) neon_lights( !vehicle.neonlights).
However this seemed to fail almost always, and inserting the value without the negative operand somehow worked.

src/gta/enums.hpp
Tayipc38: Should I remove those enums or just the comments alongside them?

Tayipc38:
> After some troubleshooting I have discovered that either the scripts or some other loop is making the invisible driver exit the vehicle on the summon feature if vehicle access is set to 'Passengers' in the interaction menu. I have looked through the scripts to find if the access is synced with a global to make a little work-around but was unsuccessful.

This has been bypassed now.

src/gta/enums.hpp
Yimura: We can recreate the enums from reverse engineering but we can't use leaks directly.

It's best you remove the whole enum.

src/services/vehicle/vehicle_control_service.hpp
```diff
@@ -0,0 +1,74 @@
+#pragma once
+#include "gta/enums.hpp"
+
+#define MAX_VEHICLE_DOORS 6
+#define MAX_VEHICLE_LOCK_STATES 11
+#define VEH_OP_ANIM_DICT "ANIM@MP_PLAYER_INTMENU@KEY_FOB@"
+#define VEH_OP_ANIM "FOB_CLICK"
```
Yimura: Prefer `constexpr auto` over macro's.

Yimura (changes requested):

src/views/view_vehicle_control.cpp
xiaoxiao921: `g_vehicle_control_service.m_controlled_vehicle.engine ? false : true`

is identical to `!g_vehicle_control_service.m_controlled_vehicle.engine` though

PR #1198, opened 2023-04-05
Admin Check Feature Updates

Rxann:
This pr adds some upgrades to the Admin Check feature.

Updates:

1. Uses the stand.gg public api for r* admin ids
2. Settings option to turn off admin checks (enabled by default)
3. Player list highlighting for admins


Rxann:
this is the branch that was merged for the original admin check feature so that is why there are so many commits

PR #1211, opened 2023-04-07
Locals editor

Tayipc38:
The debug window already had a tab called locals, but it wasn't implemented yet. I took the initiative and did my take on a locals editor. It is capable of fetching game threads and checking for their state. Locals are exclusive to their script threads, hence why the local editor needs some extra consideration when compared to globals. All locals are loaded and saved manually, so the load button will need to be used to get saved locals. The save button will overwrite the save file so load before you save if you don't want to lose your locals. This feature is meant for developers so regular users can ignore this feature.

Some pictures:
image
image
image

It is very experimental as of yet. Any improvements and corrections are more than welcome.


src/services/locals/locals_service.cpp
```diff
@@ -0,0 +1,133 @@
...
+				return thread;
+			}
+		}
+
+		return nullptr;
+	}
+
+	bool locals_service::load()
+	{
+		std::string path = std::getenv("appdata");
```
maybegreat48: You can use `g_file_manager->get_project_file("./locals.json").get_path()`

src/views/debug/view_debug_locals.cpp
```diff
@@ -1,11 +1,175 @@
...
+
+				ImGui::Text("%s : %s", local.m_name, local.m_script_thread_name);
+				//Find the thread among the script threads
+				local.m_script_thread = locals_service::find_script_thread(rage::joaat(local.m_script_thread_name));
+				//Check whether the specified script is running
+				if (local.m_script_thread && locals_service::is_script_thread_running(local.m_script_thread))
+				{
+					script_local actual_local = script_local(local.m_script_thread, local.m_base_address);
+
+					//Apply offsets to base address
```
maybegreat48: This should be cached somewhere instead of being calculated every frame

src/services/locals/locals_service.cpp
```diff
@@ -0,0 +1,133 @@
...
+		std::ifstream file(path);
+
+		if (!file.is_open())
+			return false;
+
+		try
+		{
+			nlohmann::json j;
+			file >> j;
+			this->m_locals.clear();
```
maybegreat48: `this->` is implicit inside class methods

src/views/debug/view_debug_locals.cpp
```diff
@@ -1,11 +1,175 @@
...
 namespace big
 {
+
+	void render_local_creator_popup_content()
+	{
+		static int base_address            = 0;
+		static bool freeze                 = false;
+		static char name[32]               = "";
+		static char script_thread_name[32] = "";
+		static int(*offsets)[2]            = nullptr;
```
maybegreat48: You can just use a statically allocated array of size 10 instead of all that dynamic allocation code

src/views/debug/view_debug_locals.cpp
```diff
@@ -1,11 +1,175 @@
...
+				else
+				{
+					ImGui::PushStyleColor(ImGuiCol_Text, ImVec4(1.0f, 0.0f, 0.0f, 1.0f));
+					ImGui::Text("%s isn't running", local.m_script_thread_name);
+					ImGui::PopStyleColor();
+				}
+
+				if (components::button("Delete"))
+				{
+					for (int i = 0; i < g_locals_service.m_locals.size(); i++)
```
maybegreat48: This looks like a good candidate for [`std::erase_if`](https://en.cppreference.com/w/cpp/container/vector/erase2)

src/services/locals/locals_service.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+	bool locals_service::does_script_exist(std::string script_name)
+	{
+		for (auto s : all_script_names)
+			if (script_name == s)
+				return true;
+		return false;
+	}
+
+	GtaThread* locals_service::find_script_thread(Hash hash)
```
maybegreat48: A thread find function already exists in `gta_util.hpp`

src/views/debug/view_debug_locals.cpp
```diff
@@ -1,11 +1,175 @@
...
 		{
+			if (components::button("LOAD"_T))
+				g_locals_service.load();
+			ImGui::SameLine();
+			if (components::button("SAVE"_T))
+				g_locals_service.save();
+
+			if (components::button("Add local"))
+			{
+				ImGui::OpenPopup("Addlocal");
```
maybegreat48: Renaming the popup to `##addlocal` should prevent it from showing a title

maybegreat48 (commented):
Not sure if this is intended, but the locals are only frozen when the GUI is open and the locals tab is shown

src/services/locals/locals_service.cpp
Tayipc38: > You can use `g_file_manager->get_project_file("./locals.json").get_path()`

 `g_file_manager->get_project_file("locals.json").get_path()` would be more like it I think.

Tayipc38:
> Not sure if this is intended, but the locals are only frozen when the GUI is open and the locals tab is shown

The loop is run within the UI as of now.

src/views/debug/view_debug_locals.cpp
Tayipc38: I regrettably admit that this eludes me.

Yimura (approved):
Going to merge it in as-is, because it works good enough.

Any improvements can be made later on.

PR #1214, opened 2023-04-09
feat(world): custom outfit for spawn ped

### LINKED ISSUE #1179: [Request]: Can't edit ped's outfit
### Problem

This is more of a feature rather than a problem. You can add the ability to change the appearance of the ped before spawning and not spawn it with random clothes

### Solution

I think that the spawn with random clothes can be taken out as a separate item in the drop-down list. As for me, it will be fun if you spawn Trevor without panties, for example

### Reason

Can't edit ped's outfit

### Additional context

_No response_

Aure7138:
Closes #1179 

NOTE: 
1. Export to clipboard rules have changed in this PR
2. There is no range check when setting the outfit for ped, different models have different outfit ranges, out of range may cause unknown errors


Yimura (dismissed):

Yimura:
Stale PR

PR #1227, opened 2023-04-12
Feat ptr cache continued

xiaoxiao921:


Yimura (dismissed):

xiaoxiao921:
Not really expecting contributors to be happy about that one: The MSVC compiler properly support NTTP and this code (it compiles just fine and subsequent launch of YimMenu are ultra fast)
The issue is that for some reason Intellisense is lagging behind the feature set of the compiler and is not capable of parsing the pointers.cpp file correctly, I have already made a feedback report to them but i'm just expecting it to be ignored https://developercommunity.visualstudio.com/t/Visual-Studio-incorrectly-parse-C-code/10333957

src/pointers.cpp
```diff
@@ -1,980 +1,1535 @@
...
-					LOG(FATAL) << "Just tried to load from cache a pointer supposedly within the gta 5 module range but wasnt!!! Offset from start of pointers instance: " << (reinterpret_cast<uintptr_t>(field_ptr) - reinterpret_cast<uintptr_t>(this));
-				}
+        // Uncomment this to see the current hash in the output console
+        //compile_time_helper<main_batch_and_hash.m_hash>::print_hash;
 
-				field_ptr++;
-			}
-		}
+        // Never comment this. Used for versioning the pointers cache
+        static_assert(main_batch_and_hash.m_hash == 3110229078);
```
maybegreat48: Why do we need this? Shouldn't the below statement update it automatically?

Yimura: Yeah we can most likely just remove the static_assert.

xiaoxiao921: this is embarassing, I totally don't have brain rot btw

Yimura (dismissed):

Yimura (approved):

PR #1230, opened 2023-04-13
custom gun refactor

Rxann:
This pr pretty much moves the variable definition outside of the if statement, just like the other custom gun files.

I know this is kind of a very small change, but I saw an opportunity to fix something.


Yimura (changes requested):
This is a downgrade, I'd rather all other custom guns switch to the bool in the if statement.

Rxann:
Okay, I will change them to this new solution

Yimura (approved):

PR #1231, opened 2023-04-13
Custom Weapon Firing Change

### LINKED ISSUE #1224: [Bug]: Custom Weapon fires even when weapon isn't out
### Describe the bug

When using a custom weapon, like the vehicle one, the buses for example will still spawn when you don't have a gun out.

### Steps To Reproduce

1. Enable custom weapon (I use the vehicle one and launch buses)
2. Put weapon away
3. Interact with something that uses the same key that you would to fire a gun.

### Expected Behavior

The vehicles only spawn when your weapon is out.

### Screenshots and/or Logs

_No response_

### Store

Steam

### Additional context

_No response_

### Build

Rxann:
This implements a check to see if the player has a weapon out. This pretty much makes it so custom weapons will only fire when, you know an actual gun is out. I tested this, and it worked. I am aware this may not be the most optimal solution, so please tell me  if there is anything I can do. fixes #1224


TheGreenBandit:
might want to make it so users can toggle the custom gun being disabled on weapon switch as im sure some would like to switch weapons.

Rxann:
Okay, I will work on it!

Rxann:
> might want to make it so users can toggle the custom gun being disabled on weapon switch as im sure some would like to switch weapons.

I mean the weapon is only disabled when the user doesn't have a gun out. But I will still add functionality for a setting.

Rxann:
So I have implemented @TheGreenBandit 's idea(not very well). Nevermind, It seems that I have fixed the conflicts

Rxann:
I believe these changes are ready to be merged(assuming its in the most optimal condition and is going to be put into the menu)

src/views/self/view_weapons.cpp
```diff
@@ -111,6 +111,9 @@ namespace big
 
 		components::sub_title("CUSTOM_WEAPONS"_T);
 
+		ImGui::Checkbox("Custom Gun Restriction", &g.self.custom_weapon_stop);
```
Yimura: This option is misleading and doesn't do what it says it does, it fully disables all the custom weapons regardless of if they're out or not. (aka. the tooltip description isn't right)

Yimura (changes requested):

Rxann:
Okay I will edit this

src/backend/looped/weapons/cage_gun.cpp
```diff
@@ -7,23 +7,49 @@ namespace big
 {
 	void looped::weapons_cage_gun()
 	{
-		if (const bool bCageGun = g.weapons.custom_weapon == CustomWeapon::CAGE_GUN; bCageGun)
+		if (g.self.custom_weapon_stop)
 		{
-			if (PAD::IS_DISABLED_CONTROL_PRESSED(0, (int)ControllerInputs::INPUT_AIM))
+			if (const bool bCageGun = g.weapons.custom_weapon == CustomWeapon::CAGE_GUN; bCageGun && WEAPON::IS_PED_ARMED(self::ped, 4 | 2))
```
Yimura: I hadn't really taken a look but this code can be written differently, there's a lot of code duplication for no real reason.

If you write this if statement differently you don't need to add the else statement with a whole copy of the original code.

```cpp
if (const bool bCageGun = g.weapons.custom_weapon == CustomWeapon::CAGE_GUN; bCageGun && (!g.self.custom_weapon_stop || WEAPON::IS_PED_ARMED(self::ped, 4 | 2)))
```

Writing the if statement like this will skip the WEAPON::IS_PED_ARMED call if `g.self.custom_weapon_stop` is false and if `g.self.custom_weapon_stop` is true then it will call WEAPON::IS_PED_ARMED function.

Please implement it like that for all custom weapons.

Yimura (changes requested):

Yimura (approved):
Much better code wise like this.

Rxann:
Yes, It probably runs so much better also.

PR #1250, opened 2023-04-16
Squad spawner

Tayipc38:
A literal squad spawner that spawns attack squads on unsuspecting victims. Much like the 'Spawn Ped' feature already in YimMenu, but just a little more advanced. It is definitely not finished yet, but it works adequately well. Im PR'ing this for the reviews and improvements. And of course any bug reports that come with a feature that has a lot of moving parts. For the sake of a summary ill list some features:

-Spawning squads up to a size of 8 peds.
-Providing the squad with a vehicle of choice.
-Providing the squad with a weapon of choice.
-In built 'Spawn distance' mode that makes use of pathfinding.
-Premade Templates for the normies.
-Separate victim selection.
-Advanced customization such as ped proofs, vehicle catchup etc.
-In built vehicle type detection to account for attack behavior (Heli, plane, boat).
-A description to summarize the squad.
-Load and save.

Some pictures:

image
image
image
image
image
image


DayibBaba:
Needs adjustment of classes, so dependent on another PR on the classes rep.

src/services/gui/gui_service.hpp
```diff
@@ -116,6 +117,7 @@ namespace big
 		            view::world,
 		            {
 		                {tabs::SPAWN_PED, {"GUI_TAB_SPAWN_PED", view::spawn_ped}},
+						{tabs::SQUAD_SPAWNER, {"Squad spawner", view::squad_spawner}},
```
Yimura: Weird indent

src/services/squad_spawner/squad_spawner_save_files.cpp
```diff
@@ -0,0 +1,171 @@
...
+		new_squad.m_spawn_distance             = j.value("spawndistance", 0);
+		new_squad.m_squad_size                 = j.value("squadsize", 1);
+		new_squad.m_spawn_distance_mode        = (eSquadSpawnDistance)j.value("spawndistancemode", 1);
+		new_squad.m_combat_ability_level       = (eCombatAbilityLevel)j.value("combatabilitylevel", 2);
+		new_squad.m_stay_in_veh                = j.value("stayinveh", false);
+		new_squad.m_spawn_behind_same_velocity = j.value("spawnbehindsamevelocity", true);
+		new_squad.m_disperse                   = j.value("disperse", false);
+
+		return new_squad;
+	}
```
Yimura: Replace this code with `NLOHMANN_DEFINE_TYPE_INTRUSIVE`.

You can then easily parse the json to the struct and go from the struct to json.
```cpp
// go from squad struct to JSON
nlohmann::json j = sqd;

// go from JSON to squad struct
auto sqd = j.get<squad>();
```

If you want to get an idea of how it works take a look at the `src/services/vehicle/handling_profile.hpp` file as well as the `src/services/vehicle/handling_service.cpp` file.

src/views/world/view_squad_spawner.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+		static float ped_health;
+		static float ped_armor;
+
+		static float custom_spawn_distance;
+		static int squad_size = 1;
+		static bool ped_proofs[5];
+		static bool stay_in_veh                = false;
+		static bool spawn_behind_same_velocity = false;
+		static bool disperse                   = false;
```
Yimura: Move these values to the squad service/make a struct of these so you can easily reset them.

Yimura (changes requested):

src/services/gui/gui_service.hpp
Tayipc38: Ye idk why this keeps happening even tho I press the format thingy on each edited file

src/services/squad_spawner/squad_spawner_save_files.cpp
Tayipc38: Understood

src/views/world/view_squad_spawner.cpp
Tayipc38: Consider it done

src/services/squad_spawner/squad_spawner.cpp
```diff
@@ -0,0 +1,301 @@
...
+			return true;
+		}
+		return false;
+	}
+
+	void squad_spawner::tick()
+	{
+		for (auto& s : m_active_squads)
+		{
+			if (s.is_squad_alive())
```
maybegreat48, line 243: You may want to negate the condition to remove the empty if clause

src/services/squad_spawner/squad_spawner.hpp
```diff
@@ -0,0 +1,222 @@
...
+		{
+			return m_members.size() > 0;
+		}
+
+		bool is_squad_alive()
+		{
+			bool alive = false;
+			for (auto& p : m_members)
+				if (!ENTITY::IS_ENTITY_DEAD(p.handle, false))
+					alive = true;
```
maybegreat48: You can return true right there and stop iterating more members

src/views/world/view_squad_spawner.cpp
maybegreat48: I think you should be able to use a static `squad` instance here

maybegreat48 (commented):
* The spawner does not currently work in singleplayer
* You may or may not have issues with storing pointers that may be invalidated after the ped or vehicle is destroyed
* UI elements should use Title Case
 
But LGTM otherwise

src/services/squad_spawner/squad_spawner.cpp
Tayipc38: Left that there to figure out a way how to check whether the attackers have their targets after the player dies. Ive tried about 4 methods including the SET_SEQUENCE_TO_REPEAT repeat. Im afraid that doesn't always work due to their previous task never finishing. Any idea how I could check if a player ped has respawned and make sure the attackers are tasked accordingly?

maybegreat48: Have you tried polling `GET_SCRIPT_TASK_STATUS`? Check the "documentation" for the return value enum. I assume it changes when the task stops. You can try logging it every second to get an idea of what it changes to

Tayipc38: Thats exactly the method I use for the Vehicle controller. However, it seems ped attacking tasks are varied to the point  that checking the required ones overlap certain cases id like to avoid. Since the feature has vehicle capabilities, the tasks the peds issue for the required order range from the simplest DoNothing to ExitVehicle. I have tried scrutinizing the squad vehicle with the native is mission active but that failed as soon as the peds had to leave the vehicle, since that is paired with the end of said mission. For on-foot attackers, checking with is ped in combat native is enough. I just have to consider all scenarios.

src/views/world/view_squad_spawner.cpp
Tayipc38: Wouldn't a static instance keep its ID and not increment since it would run the constructor only once? Right now I have switched to an instance but do spawn by creating another squad object to make sure the ID's are unique.

yourmartinfischer:
I am impressed by your skills and i would like to have you working for me. If you would want to i would love to have you working on exactly these kind of things(unique trolling options). I am willing to pay you for them very good. For things like this one i would be willing to pay like 150-200€ to you in Bitcoin or ETH. If you agree just leave your discord name here or make a new dc account and send to me so I can contact you there.

src/views/world/view_squad_spawner.cpp
maybegreat48: The copy constructor will copy the ID too so you have to increment it yourself

Yimura (approved):

PR #1253, opened 2023-04-16
Hotfix VM patch

maybegreat48:
YimMenu breaking the game without this patch is not a good sign


Yimura (approved):

xiaoxiao921:
Not really important but are the patches different than before? Interesting

PR #1265, opened 2023-04-18
Improved does_script_exist() and local editor

Tayipc38:
Removed the old way of checking for validity of scripts, using a game function will probably make it work for scripts that get added with updates. Also added ability to adjust a frozen value and fetch the internal address as either int, float, Bitfield or even a Vector3.

### DIFF
--- a/src/services/locals/locals_service.cpp
+++ b/src/services/locals/locals_service.cpp
@@ -19,13 +19,7 @@ namespace big
 
 	bool locals_service::does_script_exist(std::string script_name)
 	{
-		// TODO: find a better way to check script validity
-
-		for (auto s : all_script_names)
-			if (script_name == s)
-				return true;
-
-		return false;
+		return g_pointers->m_gta.m_does_script_exist(rage::joaat(script_name));
 	}
 
 	std::filesystem::path locals_service::get_path()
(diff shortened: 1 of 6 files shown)


src/services/locals/locals_service.hpp
```diff
@@ -80,6 +89,52 @@ namespace big
 			return nullptr;
 		}
 
+		float* fetch_local_pointer_float()
```
maybegreat48: You can just cast the underlying pointer instead of creating near-duplicate functions for each one

src/services/locals/locals_service.cpp
```diff
@@ -19,13 +19,7 @@ namespace big
...
 	bool locals_service::does_script_exist(std::string script_name)
 	{
-		// TODO: find a better way to check script validity
-
-		for (auto s : all_script_names)
-			if (script_name == s)
-				return true;
-
-		return false;
+		return g_pointers->m_gta.m_does_script_exist(rage::joaat(script_name));
```
maybegreat48: You can just use `DOES_SCRIPT_WITH_NAME_HASH_EXIST` (the sig and the native are the same) which works outside game threads

Tayipc38: Keeping it a native will make it easier to maintain, I agree on this one. I didn't know the native could be run on the UI and I had no idea if the invoker made any difference when executing the same function.

Yimura (approved):

PR #1283, opened 2023-04-23
feat(world): ipl

### LINKED ISSUE #1271: [Request]: Load cayo perico and north yankton in online
### Problem

Right now, there is no way to load the Cayo Perico and North Yankton maps in online

### Solution

Add an option which lets the user load in the cayo perico or north yankton map in gta online

### Reason

for fun

### Additional context

TP to cayo perico works, but if u tp outside of the party thing you get kicked out of cayo perico back to los santos

Primexz:
fixes #1271
refs https://github.com/YimMenu/Translations/pull/32


maybegreat48:
1. Active IPLs should be removed when a new one is loaded
2. This does not need a new tab (teleport should work)

Yimura (changes requested):
Is the comment from @maybegreat48 implemented? I've merged the translations on the Translation repo.

Primexz:
> Is the comment from @maybegreat48 implemented? I've merged the translations on the Translation repo.

Not yet. will do that today

Yimura (approved):

PR #1299, opened 2023-04-27
code cleanup(c-style casting -> dynamic casting)

Rxann:
This just changes c style casts to dynamic casts, which of what I know, are better


Rxann:
This should cover most of the casting, granted the codebase is huge and i can't get all of them

Yimura (approved):

PR #1308, opened 2023-04-30
Player Kick Method Choice

### LINKED ISSUE #1281: [Request]: Choose Kick in Reactions
### Problem

_No response_

### Solution

Add a List of Kick's after checking the box "Kick Player", 
Make it that you can select only one.
List : 
Null Function Kick.
End Session Kick.
Desync Kick.
OOM Kick.
Bail Kick.

### Reason

I sometimes wanna use others kicks in reactions

### Additional context

Oh and If a kick Doesn't Work Out then Use the other one

Rxann:
I need help with this. The menu items for end and null function kick do not work. I don't even know why in the hell they don't. This seems to be what I need help with.

closes #1281

### DIFF
--- a/src/backend/reactions/reaction.cpp
+++ b/src/backend/reactions/reaction.cpp
@@ -55,11 +55,16 @@ namespace big
 				else if (g.settings.end_reaction_kick)
 				{
 					dynamic_cast<player_command*>(command::get(RAGE_JOAAT("endkick")))->call(player, {});
+					script::get_current()->yield(700ms);
+					if (player->is_valid())
+					{
+						g_notification_service->push_error("End Kick Failure", "End kicks are blocked by most menus");
+					}
 				}
 				else if (g.settings.nf_reaction_kick)
 				{
 					dynamic_cast<player_command*>(command::get(RAGE_JOAAT("nfkick")))->call(player, {});
-					script::get_current()->yield(700ms);
+					script::get_current()->yield(15s);
 					if (player->is_valid())
 					{
 						g_notification_service->push_error("Null Function Kick Failed", "Attempting Try 2. Breakup kick will be used if kick fails again...");
--- a/src/backend/reactions/reaction.cpp
+++ b/src/backend/reactions/reaction.cpp
@@ -1,12 +1,14 @@
 #include "reaction.hpp"
 
 #include "backend/player_command.hpp"
+#include "core/globals.hpp"
 #include "fiber_pool.hpp"
 #include "hooking.hpp"
 #include "pointers.hpp"
 #include "script.hpp"
 #include "services/player_database/player_database_service.hpp"
 #include "util/notify.hpp"
+
 
 namespace big
 {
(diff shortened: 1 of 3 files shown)


src/views/settings/view_reaction_settings.cpp
```diff
@@ -102,6 +104,13 @@ namespace big
 		draw_reaction(g.reactions.gamer_instruction_kick);
 		draw_interloper_reaction(g.reactions.breakup_others);
 
+		components::title("Reaction Kick Method");
+		components::sub_title("Kick Option:");
+		ImGui::Checkbox("Default", &g.settings.default_reaction_kick);
+		ImGui::Checkbox("Breakup Kick", &g.settings.breakup_reaction_kick);
+		ImGui::Checkbox("End Session Kick", &g.settings.end_reaction_kick);
```
maybegreat48: Try `"End Session Kick##kick_method"` and `"Null Function Kick##kick_method"`

src/backend/reactions/reaction.cpp
```diff
@@ -39,10 +41,37 @@ namespace big
...
+				}
+				else if (g.settings.end_reaction_kick)
+				{
+					dynamic_cast<player_command*>(command::get(RAGE_JOAAT("endkick")))->call(player, {});
+				}
+				else if (g.settings.nf_reaction_kick)
+				{
+					dynamic_cast<player_command*>(command::get(RAGE_JOAAT("nfkick")))->call(player, {});
+					script::get_current()->yield(700ms);
+					if (player->is_valid())
```
maybegreat48, line 68: Null function kick should take about 15 seconds to work if they haven't blocked it so you would use the breakup kick anyway. End kick is also blocked by most menus

src/views/settings/view_reaction_settings.cpp
Rxann: Should Default and Breakup Kick have these also?

src/backend/reactions/reaction.cpp
Rxann: I have changed the script yield time to 15seconds, and do you know another kick I could use as a replacement for end kick? or if I even should have another kick?

Rxann:
I think there may be an issue with the global or checkbox because when the menu is opened none of the option checkboxes are checked.

src/backend/reactions/reaction.cpp
maybegreat48: You can try the bail kick, but literally everything has protections for that

Rxann: Okay, I think we can just leave it with end kick

src/views/settings/view_reaction_settings.cpp
maybegreat48: You don't need it if it works

Rxann: @maybegreat48 the default option isn't checked when I open it

Rxann:
Could anyone review this code because I believe that the default option isn't checked by default when opening the reactions.

Yimura:
> Could anyone review this code because I believe that the default option isn't checked by default when opening the reactions.

As I said [here](https://github.com/YimMenu/YimMenu/discussions/1286) I will not be spending time on "enhancement" PR's/PR's that add new features to the menu.
I'm completely focussing on improving the code base and fixing bugs.

Rxann:
> > Could anyone review this code because I believe that the default option isn't checked by default when opening the reactions.
> 
> As I said [here](https://github.com/YimMenu/YimMenu/discussions/1286) I will not be spending time on "enhancement" PR's/PR's that add new features to the menu. I'm completely focussing on improving the code base and fixing bugs.

Ah okay I see. I will try to debug it myself

Rxann:
Alright, this feature works, I just needed to delete my settings file 😆

Rxann:
It should be good to merge assuming no major bugs have been found

Rxann:
this is pretty stale, can be closed

gir489returns:
So then close it.

PR #1318, opened 2023-05-02
code cleanup(reorg script loading)

Rxann:
this pr moves scripts into better categories based on what they do


Yimura:
I don't think there's a benefit to merging this, there's a reason why we split those "scripts".

The reason being that `script::get_current()->yield();` would pause all other scripts executed in the same call.
By splitting those features into multiple scripts they won't be interfered by the `script::get_current()->yield();` call.

Rxann:
Okay, I see. Should I close this?

Yimura:
> Okay, I see. Should I close this?

I’ll close for you.

PR #1325, opened 2023-05-03
Refactor commands so that hard reference to variable are used instead of hardcoded strings

xiaoxiao921:
string control flow is evil:

- code can break while still compiling just fine
- it's difficult to read and understand
- it can lead to runtime errors that are hard to trace

By using hard references to variables instead of hardcoded strings, we can improve the reliability and readability of the codebase. This pull request refactors the commands to use hard references to variables wherever possible. This change will make the code more robust and easier to maintain in the long run.

~~For example, several buttons did absolutely nothing because the strings used in the `view_session.cpp` for the "all" variants of some of these commands were never implemented. I have removed them for the time being, but of course they can be reinstated when someone wants to implement them.~~
Missed the system entirely, oops, added support for it through exposing the m_all_component field

Note: The hotkey system and the command executor system still use string control flow.

With the way the hotkeys are currently written, we could move them to variable references, but I don't think we really want that because of the current system which is a kind of crappy hard-coded system: we could have a gui or even a command executor `bind` system where we could have any command to shortcut instead of a hard-coded list in the code as we currently have.


ghost:
Agree! Obviously not a code reviewer but looks great

src/backend/commands/player/kick/null_function_kick.hpp
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include "backend/player_command.hpp"
+#include "core/scr_globals.hpp"
+#include "natives.hpp"
+#include "pointers.hpp"
```
Yimura: These includes can be moved to the CPP file.

src/backend/commands/player/kick/script_host_kick.hpp
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include "backend/player_command.hpp"
+#include "core/scr_globals.hpp"
+#include "natives.hpp"
+#include "pointers.hpp"
+#include "util/scripts.hpp"
```
Yimura: These includes can be moved to the CPP file.

src/backend/commands/player/misc/clear_wanted_level.hpp
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include "backend/player_command.hpp"
+#include "natives.hpp"
+#include "pointers.hpp"
+#include "util/globals.hpp"
```
Yimura: These includes can be moved to the CPP file.

src/backend/looped/self/no_water_collision.hpp
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include "backend/looped/looped.hpp"
```
Yimura: looped.hpp?

src/backend/looped/self/part_water.hpp
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include "backend/looped/looped.hpp"
```
Yimura: looped.hpp?

src/backend/looped/vehicle/speedo_meter.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include "backend/bool_command.hpp"
+#include "backend/looped/looped.hpp"
```
Yimura: looped.hpp?

Yimura (changes requested):

Yimura (approved):
@maybegreat48, any insights on the change?

xiaoxiao921:
Will be merged after most PRs so that its not a burden to others

Yimura:
Stale PR

PR #1335, opened 2023-05-06
Context menu: Fixed irregular distribution of operations and added some more

Tayipc38:
Some operations that needed to be in the Shared pool were exclusive and there were some trivial operations missing.

Added:
-Burst tires (Vehicle)
-Halt vehicle (Vehicle)
-Recruit Ped (Ped)
-Set selected player (Player)
-Tp on top (Shared)
-Delete (Shared)
-Enflame (Shared), changed bullet spawning to explosion spawning.

Currently, recruited peds act scared and cower when entering the player vehicle. No idea why and how to fix that.

### DIFF
--- a/src/services/context_menu/context_menu_service.hpp
+++ b/src/services/context_menu/context_menu_service.hpp
@@ -71,6 +71,29 @@ namespace big
 			         else
 				         g_notification_service->push_warning("Toxic", "Failed to take control of vehicle.");
 		         }},
+		        {"BURST TIRES",
+		            [this] {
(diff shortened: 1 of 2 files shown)


Yimura (approved):
Except for the large piece of code in the recruit option the code appears to be fine.

We can move to a function later on to avoid code duplication.

Tayipc38:
> Except for the large piece of code in the recruit option the code appears to be fine.
> 
> We can move to a function later on to avoid code duplication.

Had no patience to rebuild ped.hpp

PR #1336, opened 2023-05-06
feat(player online notification)

### LINKED ISSUE #1298: [Request]: Display an alert when a player is online
### Problem

I think it's problematic to constantly click on the "Reload Player Online States" button.

### Solution

I think it should be a loop that will gradually check players online, instead of constantly clicking on the appropriate button. For saved players, you can add the corresponding checkbox "Notify when online". You can add the same to the reaction, but I think it will be superfluous.

### Reason

Makes it easier to chase down pesky griefers and other scum who get in the way of others. Well, just for convenience. Let's hope that constantly updating the status in the network of players will not hurt performance much.

### Additional context

_No response_

Rxann:
This uses the automatic player status checking to notify if the player is online.

I will make translations for these and implement them if this gets merged.

closes #1298

### DIFF
--- a/src/services/player_database/persistent_player.hpp
+++ b/src/services/player_database/persistent_player.hpp
@@ -51,6 +51,7 @@ namespace big
 		bool block_join           = false;
 		int block_join_reason     = 1;
 		bool is_modder            = false;
+		bool notify_online        = false;
 		std::unordered_set<int> infractions;
 		std::optional<CommandAccessLevel> command_access_level = std::nullopt;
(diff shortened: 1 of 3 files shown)


Yimura (changes requested):
This notifies every 5 minutes about all the players that are online, change it so it only notifies if the player went from offline to online state.

Rxann:
@Yimura is this what you mentioned?

Yimura (changes requested):
Have you tested this? This does not work as you expect it to work, check the code again.

It might not even work at all.

Yimura:
Implemented by #1386

PR #1279, opened 2023-04-23
Kick player if Block Join fails

Rxann:
This PR edits the assign_physical_index function to kick the player if block join fails (maybe).
The code compiled by I could not get the chance to test if it really works. I will try.

closes #1278

PR #1342, opened 2023-05-08
Lock Lobby

Rxann:
This pr adds a feature that allows the user to block everyone from joining.

### DIFF
--- a/src/core/globals.hpp
+++ b/src/core/globals.hpp
@@ -352,6 +352,7 @@ namespace big
 			int player_magnet_count    = 32;
 			bool is_team               = false;
 			bool join_in_sctv_slots    = false;
+			bool lock_session          = false;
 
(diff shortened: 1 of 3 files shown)


Yimura (changes requested):
Since you're adding a bunch of features around the assign_player_physical_index.cpp hook could you look into optimising the code?

For example you're fetching the player from the player service twice in this code, you can modify the code to make only one call to the player service.

Every time you add a feature or touch code somewhere you'll also be in part responsible to see if what you added is still optimised in regards to the rest of the code. Try rewriting it partly as [I suggested](https://github.com/YimMenu/YimMenu/pull/1279#issuecomment-1526207424) in the previous feature you added for `Block Join Fallback`.

Yimura (approved):

PR #1343, opened 2023-05-09
update for PDB filters notify and status 

binded-zz:
adds colors to player list player DB
and friends to PDB only with online status
and online notify from rxan 
with notice in circle
red blocked 
blue friends
orange modders

filters in PDB
friends
modders
blocked
command level access
notify online
and people that are unclassed but in the PDB
same colors apply 
friends, modders, blocked
no colors for CLA or NO
theres no onhover or selected player or player colors the status is in the overlayed circle

also adds player icons for friends you select as in PDB as well as R* friends 
also added icon for non-friends ie. blocked in PDB for those that get by and still manage to join will have a head with a x in it and be red

this should about cover it 

if they select to many as notify you will get a flood of notices when it updates every idk 5mins i guess i didnt change 
any thing for that

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,0 +1,37 @@
+name: CMake
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+
+env:
+  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
+  BUILD_TYPE: Release
+
+jobs:
+  build:
+    # The CMake configure and build commands are platform agnostic and should work equally well on Windows or Mac.
+    # You can convert this to a matrix build if you need cross-platform coverage.
+    # See: https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Configure CMake
+      # Configure CMake in a 'build' subdirectory. `CMAKE_BUILD_TYPE` is only required if you are using a single-configuration generator such as make.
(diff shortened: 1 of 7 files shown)


src/hooks/player_management/assign_physical_index.cpp
```diff
@@ -91,6 +91,9 @@ namespace big
 							plyr->is_modder         = entry->is_modder;
 							plyr->block_join        = entry->block_join;
 							plyr->block_join_reason = entry->block_join_reason;
+							plyr->is_friends        = entry->is_friends;
+            	plyr->notify_online     = entry->notify_online;
```
Rxann: indent?

src/services/player_database/player_database_service.cpp
```diff
@@ -200,7 +200,14 @@ namespace big
 					{
 						it->second->online_state = PlayerOnlineStatus::OFFLINE;
 						if (online[i] == 1)
+						{
 							it->second->online_state = PlayerOnlineStatus::ONLINE;
+							if (it->second->notify_online)
```
Rxann, line 205: This seems to be from my code, I would recommend waiting until my pr gets merged(maybe) then merge master.

Rxann:
I know you have most likely have heard this before, but I would recommend strongly to use a local machine and dev environment.

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,37 @@
+name: CMake
```
Rxann: Is this new workflow really needed? I believe there is already a CI for this...

binded-zz: i didnt know i messed it up but it will looks like its messed up

src/services/player_database/player_database_service.cpp
binded-zz: yes i commented it out left the button

.github/workflows/cmake.yml
Rxann: I would just remove the file

PR #1344, opened 2023-05-09
IP Blocking

### LINKED ISSUE #847: [Request]: Firewall or IP Filter
### Problem

Players can spoof RID and player name and bypass block joins. Even if IP is blocked through computer firewall rules, the offending IP address is still able to connect through GTA since it's using their matchmaking server. Sometimes this is enough to pass along attacks remotely. It would be beneficial if their traffic was dumped entirely. 

### Solution

A firewall for GTAO. IP filter or block join based off of IP address and or IP address range. 

### Reason

Another tool to prevent connections that are not desired. 

### Additional context

For example: I have a stalker that randomizes their RID and player name but their IP has been very consistent.

Rxann:
it implements a pretty basic IP Firewall or Blocker.

closes #847

### DIFF
--- a/src/core/globals.hpp
+++ b/src/core/globals.hpp
@@ -9,8 +9,10 @@
 #include <bitset>
 #include <imgui.h>
 #include <rage/rlSessionInfo.hpp>
+#include <vector>
 #include <weapon/CAmmoInfo.hpp>
 #include <weapon/CWeaponInfo.hpp>
+
 
 class CNetGamePlayer;
 
@@ -248,13 +250,14 @@ namespace big
 				NLOHMANN_DEFINE_TYPE_INTRUSIVE(script_events, bounty, ceo_money, clear_wanted_level, fake_deposit, force_mission, force_teleport, gta_banner, mc_teleport, personal_vehicle_destroyed, remote_off_radar, rotate_cam, send_to_cutscene, send_to_location, sound_spam, spectate, give_collectible, vehicle_kick, teleport_to_warehouse, start_activity, send_sms)
 			} script_events{};
 
-			bool desync_kick     = false;
-			bool rid_join        = false;
-			bool lessen_breakups = false; // disabled by default due to anticheat concerns
-			bool receive_pickup  = false;
-			bool admin_check     = true;
+			bool desync_kick                       = false;
+			bool rid_join                          = false;
+			bool lessen_breakups                   = false; // disabled by default due to anticheat concerns
+			bool receive_pickup                    = false;
+			bool admin_check                       = true;
+			std::vector<std::string> ip_block_list = {};
 
-			NLOHMANN_DEFINE_TYPE_INTRUSIVE(protections, script_events, rid_join, lessen_breakups, desync_kick, receive_pickup, admin_check)
+			NLOHMANN_DEFINE_TYPE_INTRUSIVE(protections, script_events, rid_join, lessen_breakups, desync_kick, receive_pickup, admin_check, ip_block_list)
 		} protections{};
 
 		struct self
--- a/src/hooks/player_management/assign_physical_index.cpp
+++ b/src/hooks/player_management/assign_physical_index.cpp
@@ -85,6 +85,21 @@ namespace big
 				{
 					if (plyr->get_net_data()->m_gamer_handle.m_rockstar_id != 0)
 					{
+						auto ip   = plyr->get_ip_address();
+						auto text = std::format("{}.{}.{}.{}", ip.m_field1, ip.m_field2, ip.m_field3, ip.m_field4);
+						auto it = std::find(g.protections.ip_block_list.begin(), g.protections.ip_block_list.end(), text);
+						if (it != g.protections.ip_block_list.end())
+						{
+							dynamic_cast<player_command*>(command::get(RAGE_JOAAT("breakup")))->call(plyr, {});
+							g_notification_service->push_warning("IP Block",
+							    std::format("Blocked a user with the IP of {}.{}.{}.{} and the name of {}",
+							        ip.m_field1,
+							        ip.m_field2,
+							        ip.m_field3,
+							        ip.m_field4,
+							        plyr->get_net_data()->m_name));
+						}
+
 						if (auto entry = g_player_database_service->get_player_by_rockstar_id(
 						        plyr->get_net_data()->m_gamer_handle.m_rockstar_id))
 						{
--- a/src/views/settings/view_protection_settings.cpp
+++ b/src/views/settings/view_protection_settings.cpp
@@ -50,6 +50,29 @@ namespace big
 			ImGui::SetTooltip("This prevents any pickup from the ground such as unwanted money drops.\nAttention: Normal pickups are also no longer possible.");
 		ImGui::Checkbox("Admin Check", &g.protections.admin_check);
 		ImGui::EndGroup();
+
+		ImGui::BeginGroup();
+		components::sub_title("Blocked IPs");
+		for (auto& element : g.protections.ip_block_list)
+		{
+			char* text = const_cast<char*>(element.c_str());
+			ImGui::Text(text);
+			if (ImGui::Button("Remove"))
+			{
+				auto result = remove(g.protections.ip_block_list.begin(), g.protections.ip_block_list.end(), element);
+				g.protections.ip_block_list.erase(result, g.protections.ip_block_list.end());
+				g_notification_service->push_warning("IP Blocking", "New IP has been removed to the Block List!");
+			}
+		}
+		ImGui::NewLine();
+		static int64_t ip;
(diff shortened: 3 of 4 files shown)


src/hooks/player_management/assign_physical_index.cpp
```diff
@@ -85,6 +85,21 @@ namespace big
 				{
 					if (plyr->get_net_data()->m_gamer_handle.m_rockstar_id != 0)
 					{
+						auto ip   = plyr->get_ip_address();
+						auto text = std::format("{}.{}.{}.{}", ip.m_field1, ip.m_field2, ip.m_field3, ip.m_field4);
```
Tayipc38, line 89: Good that you predefine the variables, makes it readable. This one could be used on line 95 as well.

src/core/globals.hpp
```diff
@@ -9,9 +9,11 @@
 #include <bitset>
 #include <imgui.h>
 #include <rage/rlSessionInfo.hpp>
+#include <vector>
 #include <weapon/CAmmoInfo.hpp>
 #include <weapon/CWeaponInfo.hpp>
 
+
```
Tayipc38, line 16: Don't see the need for a double empty line.

src/views/settings/view_protection_settings.cpp
```diff
@@ -50,6 +50,29 @@ namespace big
...
 		ImGui::Checkbox("Admin Check", &g.protections.admin_check);
 		ImGui::EndGroup();
+
+		ImGui::BeginGroup();
+		components::sub_title("Blocked IPs");
+		for (auto& element : g.protections.ip_block_list)
+		{
+			char* text = const_cast<char*>(element.c_str());
+			ImGui::Text(text);
+			if (ImGui::Button("Remove"))
```
Tayipc38, line 60: Instead of making two separate UI elements, combine them in a single one. Preferably one that employs some sort of label in which the IP could be displayed and its operation connected to removal. Such as a ```ImGui::Selectable``` with the flag ```ImGuiSelectableFlags_AllowDoubleClick``` (This method would benefit from a listbox). The button itself could even contain the blocked Ip (with some imgui magic to make it look half decent).

src/views/settings/view_protection_settings.cpp
```diff
@@ -50,6 +50,29 @@ namespace big
...
+			{
+				auto result = remove(g.protections.ip_block_list.begin(), g.protections.ip_block_list.end(), element);
+				g.protections.ip_block_list.erase(result, g.protections.ip_block_list.end());
+				g_notification_service->push_warning("IP Blocking", "New IP has been removed to the Block List!");
```
Tayipc38, line 64: I would specify the ip instead of calling it New IP.

src/views/settings/view_protection_settings.cpp
Tayipc38: I sense element id complications here as well. PushId and PopId would suffice.

src/core/globals.hpp
```diff
@@ -248,13 +250,14 @@ namespace big
...
-			bool rid_join        = false;
-			bool lessen_breakups = false; // disabled by default due to anticheat concerns
-			bool receive_pickup  = false;
-			bool admin_check     = true;
+			bool desync_kick                       = false;
+			bool rid_join                          = false;
+			bool lessen_breakups                   = false; // disabled by default due to anticheat concerns
+			bool receive_pickup                    = false;
+			bool admin_check                       = true;
+			std::vector<std::string> ip_block_list = {};
```
maybegreat48, line 258: I recommend storing IP addresses in a structure (`rage::netAddress`?) rather than a string. This will make iterating through the vector more efficient

src/views/settings/view_protection_settings.cpp
```diff
@@ -50,6 +50,29 @@ namespace big
...
+			}
+		}
+		ImGui::NewLine();
+		static int64_t ip;
+		ImGui::InputScalar("Address", ImGuiDataType_S64, &ip);
```
maybegreat48, line 69: This isn't going to work. You can create four input boxes for each IP field instead (and don't forget to use `ImGui::SameLine()`)

Tayipc38: InputInt4 is an ImGui element. Would fit right in.

src/hooks/player_management/assign_physical_index.cpp
Rxann: Thank you, I will implement this later

src/views/settings/view_protection_settings.cpp
Rxann: Okay, I will do this

src/views/settings/view_protection_settings.cpp
Rxann: So I should make a new function or is that a function of the vector?

src/core/globals.hpp
Rxann: Thank you, so do you think I should just store every plyr->get_ip_address() right into the vector?

src/views/settings/view_protection_settings.cpp
Rxann: Thank you both, I will implement this!

src/views/settings/view_protection_settings.cpp
Tayipc38: Iirc anything other than ImGui::Text, requires an unique id (the labels). Otherwise the UI elements won't behave as expected. In such scopes as loops, creating an UI element is like creating multiple. In which case you either handle the Id's yourself using the ## identifier or let the library handle it by using ImGui::PushId and ImGui::PopId respectively. Besides the id problem, the method you have employed won't be pleasing to the eye. Id advise you look at similar features in the code base and adapt.

Rxann: So I add an ImGui::PushId before calling the button in the loop?

Tayipc38: Yes. Keep in mind it requires an unique identifier param. You could use the ip itself.

src/core/globals.hpp
Rxann: Personally,  I see the IP in string form more coherent, but you are much more intelligent in this, so I will go with your opinion

Tayipc38: > Personally, I see the IP in string form more coherent, but you are much more intelligent in this, so I will go with your opinion

His intelligence is besides the point. What he means by efficient is the fact that the list will be more versatile if and when it's services are required in a different event. Imagine a scenario in which you require blocked Ip's but only their last field, dissecting the string would be hell of a lot more work than simply accessing the fourth field of rage::netAddress.

Rxann: Okay. So you guys suggest that I change the vector type to a network address and then when an ip is blocked just split the Input4 into the struct and add that?

Yimura (changes requested):
This does not cover the whole requirements of #847, because of this we could still be vulnerable from remote packets getting sent over the low levels of the network stack.

Please implement, support for CIDR ranges.
Also if possible could you implement 2 modes of blocking people?
1. Dropping their packets (blocking the incoming net message, they might still be able to join the session but not be able to communicate with you client)
2. Kicking them if their IP matches with what's in the block list (normal kick on join)

Rxann:
Alright, I will give it my best try

src/core/globals.hpp
Rxann: How do you guys suggest I implement this due to the fact you cannot do `std::vector<netAddress>`?

Rxann:
Yimura, if I may ask, why is CIDR implementation needed? From what I have seen all the player IP's are in one format(CIDR Class A maybe)

ghost:
> Yimura, if I may ask, why is CIDR implementation needed? From what I have seen all the player IP's are in one format(CIDR Class A maybe)

Because it's more user friendly to enter 10.0.0.0/24 or 10.0.0.0 - 10.0.0.255 versus 10.0.0.0, 10.0.0.1, 10.0.0.2, 10.0.0.3 ... 10.0.0.255. And that's just one subnet. Once you advance the next subnet over it's another 256 entries each. It snowballs from there. Not every isp has a huge range however, but you can come to appreciate the issue this presents. 

If your attacker has the ability to request a new ip fairly quickly, say through a vpn service, then blocking the vpn's entire ip address range is more effective than playing whack-a-mole with whatever address they get within that range. 

And not everyone has a static ip from their isp.

src/core/globals.hpp
Tayipc38: > How do you guys suggest I implement this due to the fact you cannot do `std::vector<netAddress>`?

I do not see why you wouldn't be able to make vector of netAdress.

Rxann:
Alright, This will definitely be a whole project to try to make but I will try

Rxann:
I may just close this PR. It is taking me a lot of time and I feel I could be doing better things.

PR #1346, opened 2023-05-10
Add filters to player db

binded-zz:
adds filters to player DB
all, friends, modders ,blocked, command level access, notify, and unclassed
color legend
circle in circle details 
green/ grey  = online/offline top/bottom halfs

red is always first for blocked also 
blue is friends  as long as block is not selected even if modder is selected 
orange is for modders as long as friend or block is not selected

adds icon "notfriend" to player list and red color if they bypass block
adds icon "friend" for people in the DB and blue color
adds orange color for modders no icons

 theres no function yet for the notify but the mark works


Rxann:
This pr depends on #1336 being merged?

binded-zz:
yes it did i removed that part of the code the  box only works and the back end parts to save the infos 
i made a new branch with out any of that and just posted it

PR #1354, opened 2023-05-14
Hunt the beast support for the mission viewer

Tayipc38:
Shows the current player that has been chosen as beast and lists the land marks with the closest to the beast highlighted. The land marks can also be teleported to by simply clicking the UI element.


Yimura (dismissed):

Tayipc38:
@maybegreat48 I suspect the mission UI is yours. Could you provide some insight or feedback?

src/views/network/missions/hunt_the_beast.hpp
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		return closest_index;
+	}
+
+	inline void render_hunt_the_beast_ui()
+	{
+		if (auto hunt_the_beast_script_thread = gta_util::find_script_thread(RAGE_JOAAT("am_hunt_the_beast")))
+		{
+			auto beast_player_index = *script_local(hunt_the_beast_script_thread, 599).at(1).at(6).as<uint32_t*>();
```
maybegreat48: You might want to use `scr_locals::am_hunt_the_beast::broadcast_idx` from [scr_globals.hpp](https://github.com/YimMenu/YimMenu/blob/master/src/core/scr_globals.hpp#L48)

src/views/network/missions/hunt_the_beast.hpp
```diff
@@ -0,0 +1,75 @@
...
+			if (g_player_service->get_by_id(beast_player_index))
```
maybegreat48: You can store the player here and use it inside the if block

src/views/network/missions/hunt_the_beast.hpp
```diff
@@ -0,0 +1,75 @@
...
+					{
+						auto script_local_land_mark = *beast_land_mark_list.at(i, 3).as<Vector3*>();
+						std::string label           = std::format("Tp to landmark {} at {}  {}  {}",
+                            i,
+                            script_local_land_mark.x,
+                            script_local_land_mark.y,
+                            script_local_land_mark.z);
+
+						if (ImGui::Selectable(label.data(), i == get_land_mark_beast_is_closest_to(g_player_service->get_by_id(beast_player_index), beast_land_mark_list)))
+							teleport::to_coords(script_local_land_mark);
```
maybegreat48: This probably shouldn't run in the DX thread

src/views/network/missions/hunt_the_beast.hpp
```diff
@@ -0,0 +1,75 @@
...
+			{
+				ImGui::Text("%s is the beast", g_player_service->get_by_id(beast_player_index).get()->get_name());
+				ImGui::SameLine();
+				components::button("Set as selected", [beast_player_index] {
+					g_player_service->set_selected(g_player_service->get_by_id(beast_player_index));
+				});
+
+				ImGui::Spacing();
+
+				auto beast_land_marks = *script_global(262145).at(11711).as<int*>();
```
maybegreat48, line 51: Haven't tested this yet, but you should be able to use `*beast_land_mark_list.as<int*>()` to get the array size

src/views/network/missions/hunt_the_beast.hpp
maybegreat48: You should also be able to use `teleport::teleport_player_to_coords` to teleport remote players to their objective, but you will need to add code to teleport the local player directly here or in that function

maybegreat48 (commented):
While you are at it, you might as well split up all the mission viewer scripts into separate files

Tayipc38:
> While you are at it, you might as well split up all the mission viewer scripts into separate files

Kinda what I was suggesting with the separate file on this one.

src/views/network/missions/hunt_the_beast.hpp
Tayipc38: A quick test revealed the array size to be 19 at first index, which does not work the way the feature is built. So ill stick with this particular global that is frequently used in the original script.

src/views/network/missions/hunt_the_beast.hpp
Tayipc38: I've limited the teleporting of other players to spectating only. That way the user can both teleport themselves and teleport the beast to the landmarks. Spectating results in teleporting the selected player.

Yimura (approved):

PR #1366, opened 2023-05-21
Context menu: Fixed janky entity detection

Tayipc38:
The feature was prioritizing entity types due to the nature of its iteration. It stuck to certain entities while others were closer to the center of the screen. I have decided to stick with its way of scanning the pools, but added a different (small) algorithm to fix the said issue.

### DIFF
--- a/src/services/context_menu/context_menu_service.cpp
+++ b/src/services/context_menu/context_menu_service.cpp
@@ -198,8 +198,7 @@ namespace big
 		{
 			double distance = 1;
 
-			const auto get_closest_to_center = [this, &distance](auto entity_getter_func) -> auto
-			{
+			const auto get_closest_to_center = [this, &distance](auto entity_getter_func) -> auto {
 				rage::fvector2 screen_pos{};
 				bool got_an_entity = false;
 				for (const auto entity : entity_getter_func())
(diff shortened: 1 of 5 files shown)


Tayipc38:
The entity selection might suffer from issues unrelated to the entity detection method. There are times at which it refuses to select a player ped even though they maybe located right in front.

src/services/context_menu/context_menu_service.cpp
```diff
@@ -198,8 +198,7 @@ namespace big
 		{
 			double distance = 1;
 
-			const auto get_closest_to_center = [this, &distance](auto entity_getter_func) -> auto
-			{
+			const auto get_closest_to_center = [this, &distance](auto entity_getter_func) -> auto {
```
Yimura, line 201: You can modify the `get_closest_to_center` function to return an entity pointer if it finds one close to the middle of the screen.
Then simplify the code inside the `get_closest_to_center` to not assign to m_veh_pointer, m_ped_pointer, m_prop_pointer but instead do it where you return the entity.

src/services/context_menu/context_menu_service.cpp
```diff
@@ -229,6 +233,65 @@ namespace big
 			got_an_entity |= get_closest_to_center(pools::get_all_peds);
 			got_an_entity |= get_closest_to_center(pools::get_all_props);
```
Yimura, line 234: These calls to `get_closest_to_center` can have their return values modified to m_veh_pointer, m_ped_pointer, m_prop_pointer.
You can then also move those member values to normal stack values within the function (aka not part of the class).

src/services/context_menu/context_menu_service.cpp
```diff
@@ -229,6 +233,65 @@ namespace big
...
+			if (m_prop_pointer && m_prop_pointer->m_navigation)
+			{
+				const auto prop_pos = m_prop_pointer->m_navigation->get_position();
+				HUD::GET_HUD_SCREEN_POSITION_FROM_WORLD_POSITION(prop_pos->x,
+				    prop_pos->y,
+				    prop_pos->z,
+				    &prop_screen_pos.x,
+				    &prop_screen_pos.y);
+				prop_distance = distance_to_middle_of_screen(prop_screen_pos);
+			}
```
Yimura, line 272: Code duplication, can be made into a lambda like `get_closest_to_center`.

src/services/context_menu/context_menu_service.hpp
```diff
@@ -56,6 +56,10 @@ namespace big
 
 		Entity m_handle;
 		rage::fwEntity* m_pointer{};
+		rage::fwEntity* m_veh_pointer{};
+		rage::fwEntity* m_ped_pointer{};
+		rage::fwEntity* m_prop_pointer{};
```
Yimura, line 61: Can be removed after the proposed changes to `get_closest_to_center` have been implemented.

Yimura (changes requested):

Tayipc38:
This method seems unreliable. I will try and rewrite the entity scanner in a different manner.

PR #1371, opened 2023-05-24
fix(gui): Use correct imgui style for hotkey background

ghost:
Probably not intended, feel free to disapprove if it was.


Yimura (approved):
The ImGui hotkey widget was copy pasted from some old CS:GO cheat codebase so that most likely why it doesn't use that variable.

PR #1407, opened 2023-06-13
Update to 1.67

maybegreat48:
Fixes #1406

PR #1408, opened 2023-06-13
Update 1.67 signatures

Tayipc38:
I believe to have updated the sigs and adjusted some offsets successfully. Small step for updating to 1.67.

P.s: Removed constraint attachment crash due to rockstar applying the same themselves.


DayibBaba:
Redundant #1407

Tayipc38:
Redundant due to #1407

PR #1417, opened 2023-06-15
fix(view_settings): Deal with carriage returns properly when importing stats from clipboard

denoflionsx:
The carriage return character was not being parsed out of the import string and thus the import would fail because the value would not get passed correctly.

This fix might be somewhat blunt as I'm not a C++ guy, but it at least illustrates the problem I was having when importing stats from the clipboard.


Yimura:
Ok great CI/CD is being stupid, will need to investigate why it's failing.

Yimura:
Will take a look at this later, definitely something that should be fixed.

Yimura:
PR invalidated by #2021

PR #1424, opened 2023-06-15
feat(Vehicle): Added keep engine running functionality when you leave…

Democles85:
Adding the requested functionality of keeping the engine running for the current vehicle.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,13 +12,6 @@ jobs:
       - name: Check CMake version
         run: cmake --version
 
-      - name: Cache build/
-        id: cache-build
-        uses: actions/cache@v3
-        with:
-          path: build
-          key: ${{ runner.os }}-build
-
       - name: Setup MSVC environment
         uses: ilammy/msvc-dev-cmd@v1
         with:


Yimura (changes requested):
Could you make this into a looped command? We're trying to step away from adding to looped scripts.

Democles85:
> 

Could I not, you mean?

Democles85:
> Could you make this into a looped command? We're trying to step away from adding to looped scripts.

Looped command... Can you please give me like an example of which function is a looped command? I only have a little experience when it comes to game modding and very little to YimMenu. :)

Yimura:
Take a look at the disable_siren looped command for example:
https://github.com/YimMenu/YimMenu/blob/6aa713c831b65a044bbc4561356dfc57674bea9b/src/backend/looped/vehicle/disable_siren.cpp#L1-L29

There's an `on_tick()` function which is basically where you put your code that you would otherwise loop.
Then there's also `on_disable()` which gets ran when the feature gets disabled so you can do the proper cleanup if needed.
If you need to do some one time setup in a looped command you can also use `on_enable()` for that.

Democles85:
> Take a look at the disable_siren looped command for example:
> 
> https://github.com/YimMenu/YimMenu/blob/6aa713c831b65a044bbc4561356dfc57674bea9b/src/backend/looped/vehicle/disable_siren.cpp#L1-L29
> 
> There's an `on_tick()` function which is basically where you put your code that you would otherwise loop. Then there's also `on_disable()` which gets ran when the feature gets disabled so you can do the proper cleanup if needed. If you need to do some one time setup in a looped command you can also use `on_enable()` for that.

Okay thank you! 🤗

Yimura:
No need to close the PR, you can keep pushing changes here.

Edit:
Any changes you made won't be visible either as we squash everything before it's merged into master.

Democles85:
It closed since I merged mine with yours here, for the CI... oops

PR #1428, opened 2023-06-15
fix(protection) : invaild train track index crash protection

takoyakiwhite:
It doesn't work when testing the protection, so with this.


Yimura (commented):
Does this fix the issue? Have you tested this?

MarkEcza:
Track ID is unsigned

takoyakiwhite:
> Does this fix the issue? Have you tested this?

Before I modified it, I LOG m_track_id result is 口, it is not recognized, after modification the normal m_track_id is -10

MarkEcza:
> > Does this fix the issue? Have you tested this?
> 
> Before I modified it, I LOG m_track_id result is 口, it is not recognized, after modification the normal m_track_id is -10

Track ID value is unsigned , It already checks if the track id is below 0 anyways.

MarkEcza:
image
image

First picture is the gamestate node, second is a vfunc in CTrain. + 0x150F is the track id.

xiaoxiao921:
Change the pr to 
```cpp
if (train_node->m_track_id >= 27)
```

Change https://github.com/Yimura/GTAV-Classes/blob/master/netsync/nodes/train/CTrainGameStateDataNode.hpp#L20
to 
```cpp
uint8_t
```

though i'm not really sure why the current code doesnt work? should be just fine?

takoyakiwhite:
> Change the pr to
> 
> ```c++
> if (train_node->m_track_id >= 27)
> ```
> 
> Change https://github.com/Yimura/GTAV-Classes/blob/master/netsync/nodes/train/CTrainGameStateDataNode.hpp#L20 to
> 
> ```c++
> uint8_t
> ```
> 
> though i'm not really sure why the current code doesnt work? should be just fine?

it's log result
image

Yimura:
Look the problem lies with how you're logging it, the data is still there indicating the value.
The CPU will still see it as a number and do the comparison correctly, there's no issue with the current implementation.

Make a simple C++ program where you do the following:
```cpp
#include <iostream>

int main()
{
    unsigned char m_track_id = 5;

    std::cout << m_track_id;

    return 0;
}
```

You should see the exact same output, this is because std::cout tries to print it out as a character from the ASCII table and not as a number because it doesn't see it as a number.
If you cast it to an int you will see that std::cout understands it as a number and prints out the correct value.

```cpp
#include <iostream>

int main()
{
    unsigned char m_track_id = 5;

    std::cout << (int)m_track_id;

    return 0;
}
```

xiaoxiao921:
```cpp
LOG(INFO) << HEX_TO_UPPER(train_node->m_track_id);
```
will work just fine too

takoyakiwhite:
> Look the problem lies with how you're logging it, the data is still there indicating the value. The CPU will still see it as a number and do the comparison correctly, there's no issue with the current implementation.
> 
> Make a simple C++ program where you do the following:
> 
> ```c++
> #include <iostream>
> 
> int main()
> {
>     unsigned char m_track_id = 5;
> 
>     std::cout << m_track_id;
> 
>     return 0;
> }
> ```
> 
> You should see the exact same output, this is because std::cout tries to print it out as a character from the ASCII table and not as a number because it doesn't see it as a number. If you cast it to an int you will see that std::cout understands it as a number and prints out the correct value.
> 
> ```c++
> #include <iostream>
> 
> int main()
> {
>     unsigned char m_track_id = 5;
> 
>     std::cout << (int)m_track_id;
> 
>     return 0;
> }
> ```

PR #1434, opened 2023-06-16
Fix train track id read in can apply data.

MarkEcza:
The track id is a unsigned char value. Here i casted it as that then i read it as a unsigned int to compare to a number. 
The track id is already checked if its below 0 in the game code.

It's not tested yet. But it works in my own CTrain vfunc hook that i use for this crash.


xiaoxiao921:
Just change the field type instead of doing stuff like this...

Yimura:
Yes you can change the type to uint8_t in the GTAV-Classes repo, no need to dirty cast like this.

Also, did you test this?

PR #1437, opened 2023-06-16
Basic hostile ped disarm

### LINKED ISSUE #1403: [Request]: Add auto disarm weapon from peds.
### Problem

In GTA Online heists, when peds shoot your character, or those from your teamates, you loose money.
It would be very useful to have an option that automatically remove the weapons from the peds that are around.

### Solution

.

### Reason

.

### Additional context

There is already an option to make peds ignore you but that probably won't work (i didn't actually tried that option) if you are playing with your teammates, they will still get shooten from the peds.

Tayipc38:
Closes #1403

### DIFF
--- a/src/backend/looped/world/nearby/auto_disarm.cpp
+++ b/src/backend/looped/world/nearby/auto_disarm.cpp
@@ -36,10 +36,6 @@ namespace big
 				}
 			}
 		}
-
-		virtual void on_disable() override
-		{
-		}
 	};
 
 	auto_disarm g_auto_disarm("autodisarm", "Auto Disarm", "Disarm nearby pedestrians", g.world.nearby.auto_disarm.enable);
--- a/src/backend/looped/world/nearby/auto_disarm.cpp
+++ b/src/backend/looped/world/nearby/auto_disarm.cpp
@@ -1,0 +1,46 @@
+#include "backend/looped_command.hpp"
+#include "natives.hpp"
+#include "util/entity.hpp"
+#include "util/pools.hpp"
+
+namespace big
+{
+	class auto_disarm : looped_command
+	{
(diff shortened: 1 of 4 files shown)


src/views/world/view_nearby.cpp
```diff
@@ -42,6 +42,10 @@ namespace big
 		components::command_checkbox<"highalert">();
 		ImGui::SameLine(140.f);
 		components::command_checkbox<"pedrush">();
+		components::command_checkbox<"autodisarm">();
```
Rxann, line 45: Personally I would put this inside the modal so the modal contains all the items related to the feature. You could have a button that opens the modal. just a suggestion either works.

src/backend/looped/world/nearby/auto_disarm.cpp
```diff
@@ -0,0 +1,46 @@
...
+						if (g.world.nearby.auto_disarm.neutralize)
+							TASK::TASK_STAND_STILL(handle, -1);
+					}
+				}
+			}
+		}
+
+		virtual void on_disable() override
+		{
+		}
```
Yimura: Does this have to exist? Can't you just leave it away?

Tayipc38: Oversight after copy pasting another command. Makes you think how many have such redundancies.

src/views/world/view_nearby.cpp
Tayipc38: The rest of similar elements are designed in this way. Where the global enable variable is readily available, and its subordinates within a modal. Although this was my idea, I am open to suggestions since it is not the prettiest look.

Yimura (approved):

PR #1476, opened 2023-06-20
Nearby entity deletion

### LINKED ISSUE #1455: [Request]: Nearby options for entity deletion
### Problem

Car delete.

### Solution

A button can be added where I can delete the vehicle I am in and the vehicles around me.

### Reason

When I spawn a vehicle, when I want to delete that vehicle, I have to open noclip and throw it away from the map :D

### Additional context

_No response_

Tayipc38:
Closes #1455 

~~Currently takes several tries to clear the whole area. No clue why it is skipping entities.~~

### DIFF
--- a/src/gta/pools.hpp
+++ b/src/gta/pools.hpp
@@ -56,7 +56,8 @@ public:
 		std::vector<Entity> arr;
 		for (auto entity : *static_cast<T*>(this))
 		{
-			arr.push_back(big::g_pointers->m_gta.m_ptr_to_handle(entity));
+			if(entity)
+				arr.push_back(big::g_pointers->m_gta.m_ptr_to_handle(entity));
 		}
 
 		return arr;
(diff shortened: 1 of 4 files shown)


maybegreat48:
Perhaps you don't have control of some of the nearby entities? You need control to delete entities (but not to set them as no longer needed)

Tayipc38:
> Perhaps you don't have control of some of the nearby entities? You need control to delete entities (but not to set them as no longer needed)

I have tried to control both peds and objects. On objects the control seems to bug out even though it works without. Peds don't bug out but it makes practically no difference. The reason i kept requests on vehicle is due to players often being in them. ~~I suspect the query of the pools being faulty.~~

ghost:
You should probably fiber tick after deleting an entity

Tayipc38:
> You should probably fiber tick after deleting an entity

Do you mean yield before checking deletion?

Tayipc38:
Well it seems I had made an embarrassing mistake regarding the for loop. Erasing vector elements during its iteration results in an obvious iteration mishap, how on earth I could not see that is beyond me.

Yimura (approved):

PR #1479, opened 2023-06-20
view_session.cpp UI tweaks

Tayipc38:
It was bugging me that the Network tab had no default UI. Instead the Session tab contained all relevant features. I suggest we move this and redesign the UI as following (included in this PR).

image

### DIFF
--- a/src/services/gui/gui_service.hpp
+++ b/src/services/gui/gui_service.hpp
@@ -36,7 +36,6 @@ namespace big
 		NEARBY,
 
 		NETWORK,
-		SESSION,
 		MISSIONS,
 		SPOOFING,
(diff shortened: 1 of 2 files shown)


Yimura (approved):

tupoy-ya:
Looks good as a concept(for YimMenu UI), no idea how usable it will be in practice.

The only problem i see now is that it is inconsistent because it is not used anywhere else in the menu, and I don't know where else it could be used.

<hr>

###### NOTE: I didn't test this or look at the code.

Yimura:
It's a definite improvement and we can always switch back or modify it over time.

It was just getting too crowded and disorganised, this will improve on those points.

PR #1510, opened 2023-06-23
fix: Bool CMD hotkey toggle message showing translation key

Yimura:
Issue introduce when we added support for translations in commands.

### DIFF
--- a/src/backend/bool_command.cpp
+++ b/src/backend/bool_command.cpp
@@ -1,4 +1,5 @@
 #include "bool_command.hpp"
+#include "services/translation_service/translation_service.hpp"
 
 namespace big
 {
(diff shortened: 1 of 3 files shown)


tupoy-ya (approved):
Should work, didn't test.

Commands also check if translation is not empty, no idea if it is needed.
```cpp
std::string key;
if (key = g_translation_service.get_translation(m_label_hash); key.empty())
	key = m_label;
```

Yimura:
The translation service just returns the key passed to it if it doesn't exist in the translation service.

Edit:
Never mind, only the string literal does that.
I'll instead just introduce the behaviour of returning the key passed to it.

Edit edit:
Not possible with how the translation service is coded, I will implement your code instead.

Yimura:
I instead modified the translation service and added fallback key to the joaat version of the `translation_service::get_translation` function.

The string_view version of get_translation then passes the translation_key as a fallback key to the joaat version of get_translation, this way original behaviour is also preserved.

tupoy-ya (approved):
Looks good to me.

Didn't test.

PR #1517, opened 2023-06-23
Add vehicles from LS Mercenaries

tyackman:


tupoy-ya:
Is this file even used anywhere in YimMenu?

Yimura:
> Is this file even used anywhere in YimMenu?

Nope, we might just remove it.

PR #1521, opened 2023-06-24
Vehicle tabs UI tweak

Tayipc38:
Instead of having 3 separate tabs we could have them in one. This way any new method of spawning vehicles can just be added to the same UI. Also adjusted some small elements to make it look nicer.

image
image


src/views/vehicle/view_spawn_vehicle.cpp
```diff
@@ -196,4 +196,24 @@ namespace big
 			ImGui::EndListBox();
 		}
 	}
+
+	void view::spawn_vehicle()
+	{
+		static int spawn_type = 0;
+		if (ImGui::RadioButton("New", spawn_type == 0))
```
tupoy-ya: Why does nobody look at the demo?

```cpp
IMGUI_DEMO_MARKER("Widgets/Basic/RadioButton");
static int e = 0;
ImGui::RadioButton("radio a", &e, 0); ImGui::SameLine();
ImGui::RadioButton("radio b", &e, 1); ImGui::SameLine();
ImGui::RadioButton("radio c", &e, 2);
```

Tayipc38: Ohw, didnt know lol

Yimura (approved):

PR #1552, opened 2023-06-28
Lua doc

xiaoxiao921:
Doc isn't complete but we got an ok doc converter I think, feel free to give feedback

can be merged as is tbf as it's better than nothing and extra doc comments can be added later


Aboba7778:
How to create new menu tab and elements (Adding new items from your documentation doesn't work (maybe I'm doing something wrong))

Tayipc38:
> How to create new menu tab and elements (Adding new items from your documentation doesn't work (maybe I'm doing something wrong))

I don't think thats possible in its current state. How would you think of drawing a new lua tab? Like where would it ideally be located?

Aboba7778:
> > How to create new menu tab and elements (Adding new items from your documentation doesn't work (maybe I'm doing something wrong))
> 
> I don't think thats possible in its current state. How would you think of drawing a new lua tab? Like where would it ideally be located?

I had an idea to make it possible to create separate imgui windows and put elements there via lua code.

I took the code for creating a checkbox from this documentation and I don’t understand where it is placed and where it appears in the menu

FIying-Scotsman:
> > How to create new menu tab and elements (Adding new items from your documentation doesn't work (maybe I'm doing something wrong))
> 
> I don't think thats possible in its current state. How would you think of drawing a new lua tab? Like where would it ideally be located?

As a reference- [CyberEngine Tweaks (CET)](https://github.com/maximegmd/CyberEngineTweaks/tree/v1.25.2)  (think SHV but for Cyberpunk) extends its Imgui implementation used for the console to lua scripts so they can draw their own menus using the “core” Imgui implementation in CET. 

Each Lua script has its own floating window and can control the style, layout etc of its own window. 

It was designed around that however so I’m not sure if that could be translated into Yim as easily.

Yimura:
> It was designed around that however so I’m not sure if that could be translated into Yim as easily.

It’s possible to inject into existing UI tabs currently, not sure if it’s possible to create new tabs though and then write UI elements to there.

xiaoxiao921:
> > It was designed around that however so I’m not sure if that could be translated into Yim as easily.
> 
> It’s possible to inject into existing UI tabs currently, not sure if it’s possible to create new tabs though and then write UI elements to there.

Not possible to make new ones currently, only hijack existing ones, making new windows is probably the best way to go

Tayipc38:
> > > It was designed around that however so I’m not sure if that could be translated into Yim as easily.
> > 
> > 
> > It’s possible to inject into existing UI tabs currently, not sure if it’s possible to create new tabs though and then write UI elements to there.
> 
> Not possible to make new ones currently, only hijack existing ones, making new windows is probably the best way to go

A single window to host the scripts and separate them somehow, maybe with a listview or even tabs (tabs are ugly) would be ideal.

Aboba7778:
> > > It was designed around that however so I’m not sure if that could be translated into Yim as easily.
> > 
> > 
> > It’s possible to inject into existing UI tabs currently, not sure if it’s possible to create new tabs though and then write UI elements to there.
> 
> Not possible to make new ones currently, only hijack existing ones, making new windows is probably the best way to go

How can I add my element to a specific tab using lua code?

xiaoxiao921:
Pr is ready.

i'll most likely make other prs related to the lua documentation, most likely two, one centered around tabs and the other around yim menu commands

yim menu commands are difficult to parse due to the way cpp is written, i'm thinking using the yimmenu project cpp code itself as a way to generate it, of course this piece of code won't be at all compiled for release builds for not bloating it, if anyone has any other proper idea / library suggestion for this please tell me.

Basically the easiest would be to simply log g_command from the cpp code itself and format that into markdown like the rest of the doc is instead of using an external parser like we are doing for the rest of the doc

Yimura:
Documentation is a bit hidden, if we're having it in the Github repository I'd put them at the top level in a `docs/lua/`.
Otherwise I'd also like to take a look at generating and putting them automatically on the wiki.
@GeopJr any insights on this?

Yimura (approved):
Gonna approve they're only MD files so should be fine.

We can look to other solutions in the future.

PR #1557, opened 2023-06-29
More bug fixes

Tayipc38:
Fixed nearby features affecting players and local ped.
Removed faulty vehicle kick protection from `can_apply`.

### DIFF
--- a/src/hooks/protections/can_apply_data.cpp
+++ b/src/hooks/protections/can_apply_data.cpp
@@ -773,10 +773,6 @@ namespace big
 				{
 					if (object && g_local_player && g_local_player->m_net_object)
 					{
-						// if (is_in_vehicle(g_local_player, g_local_player->m_vehicle) && g_local_player->m_vehicle->m_net_object
-						//     && g_local_player->m_vehicle->m_net_object->m_object_id == object->m_object_id)
-						// 	return true; // vehicle kick?
-
 						if (!g_local_player->m_vehicle || !g_local_player->m_vehicle->m_net_object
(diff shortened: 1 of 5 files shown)


Yimura (dismissed):

src/hooks/protections/can_apply_data.cpp
```diff
@@ -773,9 +773,9 @@ namespace big
 				{
 					if (object && g_local_player && g_local_player->m_net_object)
 					{
-						if (is_in_vehicle(g_local_player, g_local_player->m_vehicle) && g_local_player->m_vehicle->m_net_object
-						    && g_local_player->m_vehicle->m_net_object->m_object_id == object->m_object_id)
-							return true; // vehicle kick?
+						// if (is_in_vehicle(g_local_player, g_local_player->m_vehicle) && g_local_player->m_vehicle->m_net_object
+						//     && g_local_player->m_vehicle->m_net_object->m_object_id == object->m_object_id)
+						// 	return true; // vehicle kick?
```
Yimura: We could've just removed it since we've got version control.

Though it might not take too long until this gets fixed so it can stay for a bit.

Yimura (dismissed):

src/hooks/protections/can_apply_data.cpp
Tayipc38: I hesitated cuz it is not mine and it might be an easy fix.

maybegreat48: You can remove it for now, it never worked anyway

Yimura (approved):

PR #1567, opened 2023-06-30
Custom location system

Tayipc38:
A simple json formatted location saving feature:

image

Needs a little testing.

Also put fetching squads and these new locations to the backend.cpp to be loaded on inject.

### DIFF
--- a/src/backend/backend.cpp
+++ b/src/backend/backend.cpp
@@ -11,6 +11,8 @@
 #include "services/tunables/tunables_service.hpp"
 #include "services/vehicle/vehicle_control_service.hpp"
 #include "thread_pool.hpp"
+#include "util/teleport.hpp"
+#include "services/squad_spawner/squad_spawner.hpp"
 
 
 namespace big
@@ -21,6 +23,8 @@ namespace big
 			command->refresh();
 
 		register_script_patches();
+		teleport::fetch_saved_locations();
+		g_squad_spawner_service.fetch_squads();
 
 		while (g_running)
 		{
--- a/src/util/teleport.hpp
+++ b/src/util/teleport.hpp
@@ -8,6 +8,160 @@
 
 namespace big::teleport
 {
+	struct telelocation
+	{
+		std::string name;
+		float x, y, z;
+		std::string category = "Default";
+	};
+
+	NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(telelocation, name, x, y, z, category);
+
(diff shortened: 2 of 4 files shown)


Rxann:
Should these be inlined when they are decently big functions?

Yimura:
> Should these be inlined when they are decently big functions?

They need to be declared with inline in .hpp files if you're going to include the .hpp file in more than one place.

Tayipc38:
> Should these be inlined when they are decently big functions?

Like @Yimura mentioned, I have to inline them due to the nature of the file structure. It was not big enough to have its own service yet does seem too large for it to be made like this. I believe I can think of a couple more additions that might make use of `teleport::all_saved_locations`.

Yimura (changes requested):
I would probably move these functions to a service since you've got data members like the std::vector with telelocations in it.
Some functions don't need to be public like the get_telelocations_file_path function, only these specific functions are making use of that path anyways.

In regards to categories you could make another struct like follows:
```cpp
struct telelocation_category
{
    std::vector<telelocation> locations;
    std::string category.
};
// key will be the category for easier lookup when adding a new telelocation to a category
using telelocation_map = std::map<std::string, telelocation_category>;
```
```json
{
    "Default": {
        "category": "Default",
        "locations": [
            { ... }
        ]
    }
}
```
Structuring the data like this will have less duplication of categories in each telelocation struct (aka remove the member that saves the category).


When loading in the file nlohmann::json will do the conversion from JSON array to std::vector properly so no need to do anything extra.

Same for saving a new telelocation, no need to load the file from disk again, check against if it's actually an array or not and then push a JSON object onto that array to then write it to disk. I'd just push it on the all_saved_locations and save that.

You'd then only need to add a check to the reader and if there's a failure just don't read the data from the JSON file. Next time a location is saved it will just wipe the file with valid data.

And if you want to give the user an option to load a new location file from disk just add a reload button in the UI.

Tayipc38:
>A whole lot of structure related mumbo jumbo...

Saving categories with their own locations does seem like a good idea. But id like to keep the way im maintaining the actual life-cycles of the entries. The way it works now is entirely dependent on the save file and won't cause any differences in the run-time instance of the structure object. In short, saving, clearing and fetching should be the way to go imo.

This will avoid any unwanted loss of data due to writing errors, in which case the local instance will have the new entries and the save file wont. Resulting in a bad surprise upon a reload of the menu.

Yimura (approved):

PR #1575, opened 2023-07-01
Fix lua native bindings and lua script reload

### LINKED ISSUE #1565: [Bug]: Crashing the game when reloading Lua
### Describe the bug

Using script. register_ When looped, there will be no problem when adding code and reloading it normally, but if I delete it and write other code before reloading, it will cause the game to crash without any logs

### Steps To Reproduce

First, run this string of code.
`script.register_looped("test",function()
	local mypid = PLAYER.PLAYER_ID()
	log.info("test register_looped")
	script.yield()
end)`
After running, modify it like this before reloading
`log.info("test")`
After reloading, the game crashed without an error log

### Expected Behavior

It should be able to reload normally.

### Screenshots and/or Logs

```
<paste your logs here>
```

### Store

Steam

### Additional context

_No response_

### Build

xiaoxiao921:
This isnt the cleanest at all but this fix a lot of issues with the current system

Closes #1565 : Reloading scripts should not crash anymore

Fix pointers parameters for lua natives:
- for out parameters: they are returned as a tuple, sol handle the rest (fix natives that are defined like NETWORK_GET_GLOBAL_MULTIPLAYER_CLOCK)
- for in parameters: they are just passed as value instead, still returned though sometimes it's not useful, fix natives like ENTITY.DELETE_ENTITY(ent)

This is not perfect and some natives won't work correctly, but it's relatively low now.


Yimura (approved):
No way I'm checking the whole 36k lines of code from the bindings file, it managed to compile.

PR #1578, opened 2023-07-01
refactor!: General changes

Yimura:
- Removed try catch from main.cpp for better error logging
- Added migration code to the YimMenu folder
- Renamed globals to settings


src/util/migrate.hpp
```diff
@@ -0,0 +1,17 @@
...
+{
+    void do_migration(const std::filesystem::path& current_dir)
+    {
+        std::filesystem::path old_dir = std::getenv("appdata");
+        old_dir /= "BigBaseV2";
+
+        if (!exists(current_dir))
+        {
+            LOG(INFO) << "YimMenu folder does not exist, migrating to new folder.";
+            std::filesystem::copy(old_dir, current_dir, std::filesystem::copy_options::recursive);
```
tupoy-ya, line 14: It shouldn't try to do this if neither BigBaseV2 or YimMenu folders exist.

Just a nitpick as it seems the only "problem" it would cause is an extra log line for new users.

tupoy-ya (changes requested):
In what way is this a `feat:`?
This is a breaking `refactor!:`.

tupoy-ya (approved):
The game launched with this didn't test further.
Tested on the eaab129b1960f85633a1be94d7753f7ef84fae47 commit.

Didn't know we could log before logger init.

F to disc space of people with 4GB exception logs.

PR #1579, opened 2023-07-02
feat(PlayerDB): Added search button

Yimura:
This adds a "Search" button on the Player DB feature for people that don't necessarily have the RID of a user but remember the name.


tupoy-ya:
Search is probably a bad name for this, maybe "Find by Name" or something like this.
Also we already have an item with ID "Search" in this view, ImGui does not like that.

Yimura:
> Search is probably a bad name for this, maybe "Find by Name" or something like this. Also we already have an item with ID "Search" in this view, ImGui does not like that.

That `SEARCH` isn't used by ImGui as the key for the other widget however I do agree that "Search" does not make it obvious what it does exactly.

PR #1594, opened 2023-07-04
Xml vehicles (Menyoo)

### LINKED ISSUE #864: [Request]: INI and XML
### Problem

INI and XML

### Solution

Request the addition of INI and XML support

### Reason

practicability

### Additional context

_No response_

Tayipc38:
Initial menyoo xml reader, bound to be updated since I can't be bothered to test it with more files than I did. I've left out what I think was unnecessary.

image
image

Closes #864


src/services/vehicle/xml_vehicles_service.cpp
```diff
@@ -0,0 +1,249 @@
...
+#include "util/vehicle.hpp"
+#include "util/world_model.hpp"
+
+namespace big
+{
+
+	xml_vehicles_service::xml_vehicles_service()
+	{
+		g_xml_vehicles_service = this;
+	}
```
Yimura, line 15: No deconstructor present to set the `g_xml_vehicles_service` ptr to `nullptr`.

iDominiqe:
How does it all work? And how do you add yours?

iDominiqe:
works great, thank you very much, waiting for the update! if you can also add support for maps 😍

iDominiqe:
if someone uses a modified xml transport, then errors pop up, they say the user is trying to crash you. Basically works great
image
image
image
image

Tayipc38:
There are missing parts regarding xml nodes, but I believe the essentials are covered. Any incompatible xml files I'd like to see personally so I can add or fix its inclusion.

Yimura (approved):

PR #1596, opened 2023-07-04
feat(Overlay): Added Invisibility indicator and is able to save in unload.

Democles85:
1. Added the Invisibility Indicator as requested
2. Is not able to save on Unload because I am stupid and forgor that I only did that in my own version... lel developer btw.
3. Structured overlay indicator settings a little better I guess :d

image


src/views/settings/view_gui_settings.cpp
```diff
@@ -63,18 +63,24 @@ namespace big
 		{
 			if (ImGui::TreeNode("Overlay Indicators"))
 			{
+				ImGui::Separator();
 				ImGui::BeginGroup();
+				ImGui::Text("True by default");
```
Yimura: Seems unnecessary

src/views/settings/view_gui_settings.cpp
```diff
@@ -63,18 +63,24 @@ namespace big
...
 				ImGui::Checkbox("Show Player Godmode", &g.window.ingame_overlay_indicators.show_player_godmode);
 				ImGui::Checkbox("Show Off Radar", &g.window.ingame_overlay_indicators.show_off_radar);
 				ImGui::Checkbox("Show Vehicle Godmode", &g.window.ingame_overlay_indicators.show_vehicle_godmode);
 				ImGui::Checkbox("Show Never Wanted", &g.window.ingame_overlay_indicators.show_never_wanted);
 				ImGui::EndGroup();
+
 				ImGui::SameLine();
+
 				ImGui::BeginGroup();
+				ImGui::Text("False by default");
```
Yimura: Seems unnecessary 

Up for debate obviously however I don't think this information matters after a user has modified it.

Yimura (changes requested):

src/views/settings/view_gui_settings.cpp
Yimura: Up for debate obviously however I don't think this information matters after a user has modified it.

Issue #1610, opened 2023-07-05
[Request]: Infinite Ammo causes weapon status HUD to disappear, offer Always Full Ammo instead


gir489returns:
### Problem

When Infinite Ammo is used, the native SET_PED_INFINITE_AMMO is called, which disables showing the weapon's stats.

![image](https://github.com/YimMenu/YimMenu/assets/100792176/b4e437db-787e-44f0-a085-5ad418a8886a)
![image](https://github.com/YimMenu/YimMenu/assets/100792176/818fac7c-4758-4ae1-84e2-76904371317d)

This is problematic as I can't see how much ammo is left in the magazine.

### Solution

Offer a second function that just constantly refills the player's ammo like so:

`int max_ammo;
if (WEAPON::IS_WEAPON_VALID(g_local_player.weapon) && WEAPON::GET_MAX_AMMO(g_local_player.player_ped, g_local_player.weapon, &max_ammo))
	WEAPON::SET_PED_AMMO(g_local_player.player_ped, g_local_player.weapon, max_ammo, 0);`

### Reason

Knowledge is power.

gir489returns:
Closed as #1616 resolved this issue.

PR #1616, opened 2023-07-05
Added Always Full Ammo.

gir489returns:
This addresses issue #1610

### DIFF
--- a/src/views/self/view_weapons.cpp
+++ b/src/views/self/view_weapons.cpp
@@ -17,6 +17,7 @@ namespace big
 		ImGui::BeginGroup();
 
 		components::command_checkbox<"infammo">();
+		components::command_checkbox<"alwaysfullammo">();
 		components::command_checkbox<"infclip">();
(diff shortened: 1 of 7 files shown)


Yimura (approved):

src/views/self/view_weapons.cpp
```diff
@@ -17,6 +17,7 @@ namespace big
 		ImGui::BeginGroup();
 
 		components::command_checkbox<"infammo">();
+		components::command_checkbox<"alwaysfullammo">();
```
TheGreenBandit, line 20: might want to emphasize difference between infinite ammo

gir489returns: What? This is just the internal command name. The user will see "Always Full Ammo" and the description of "Refills your ammo every tick." What would you have me change it to?

Yimura: Nah it's fine as-is.

PR #1645, opened 2023-07-08
fix(cache): change how the game folder is retrieved

xiaoxiao921:
Use GetModuleFileNameW instead of std::filesystem::current_path()

### DIFF
--- a/src/services/gta_data/yim_fipackfile.cpp
+++ b/src/services/gta_data/yim_fipackfile.cpp
@@ -84,9 +84,49 @@ namespace big
 		return output;
 	}
 
+	static std::filesystem::path get_game_folder_path()
+	{
+		std::wstring game_module_path(MAX_PATH, '\0');
+
+		if (GetModuleFileNameW(nullptr, game_module_path.data(), game_module_path.size()))
+		{
+			const auto game_folder = std::filesystem::path(game_module_path).parent_path();
+
+			if (std::filesystem::is_directory(game_folder))
+			{
+				return game_folder;
+			}
+			else
+			{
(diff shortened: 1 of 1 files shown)


src/services/gta_data/yim_fipackfile.cpp
```diff
@@ -84,16 +84,58 @@ namespace big
 		return output;
 	}
 
+	static std::filesystem::path get_game_folder_path()
+	{
+		std::wstring game_module_path(MAX_PATH, '\0');
+
+		if (GetModuleFileNameW(nullptr, game_module_path.data(), game_module_path.size()))
```
gir489returns, line 91: When I used to manually map my cheat, I had problems with using the WinAPI to retrieve information about GTA's process. Has this been tested with MMing? I don't know if the project only supports LLA injection or not, but that might be something to consider.

xiaoxiao921: I can't even MM with current nightly (without this PR merged in) so ¯\_(ツ)_/¯

Yimura: Was the user having problems injecting with manual mapping?

xiaoxiao921: No, i'm just suspecting std::filesystem::current_path being able to return bogus result on some systems, so i'm just removing one extra layer of potential non controlled errors

Yimura (approved):
Is this ready to merge or should we wait for more user input before merging?

xiaoxiao921:
You can merge, i'm confident this is much safer than having to rely on current_path

PR #1646, opened 2023-07-08
(feat) alternative compact interface

T14D3:
Will be draft for a while, working on implementing an alternative more "traditional"-styled menu for keyboard-only navigation


gir489returns:
This commit seems like nonsense. Why are you basically creating a 2nd duplicate menu with another key? I don't see how this is anymore useful than just going through the traditional INS menu.

T14D3:
The point is having a 2nd menu that is easily usable just using keyboard input, without blocking mouse input to the game - this would allow you to change things on the fly, while driving etc.
Again, very much still a draft, just very rough outline so people can give feedback

Tayipc38:
> The point is having a 2nd menu that is easily usable just using keyboard input, without blocking mouse input to the game - this would allow you to change things on the fly, while driving etc. Again, very much still a draft, just very rough outline so people can give feedback

I think @tupoy-ya did this without an extra interface.

T14D3:
> I think @tupoy-ya did this without an extra interface.

I know he did, but afaik his implementation is more of a replacement/modification of the original interface - I'm aiming for more of an "additional" feature that should not replace the original gui, but instead just be a way to quickly access some of the functions of the menu - but not all of them, by far. Might be somewhat similar to the "Self>Mobile" thingie, maybe even replace it at some point, since I'm pretty sure most if not all of the features of "Mobile" exist somewhere else as well

tupoy-ya:
> I think @tupoy-ya did this without an extra interface.

I forgot it existed because i almost never actually used it.
Plus it was removed a long time ago.

ghost:
怎么下载？

T14D3:
Very much still work in progress, you can compile from source if you want or wait 'till I set up actions on my fork

> how to download?

Yimura:
We currently have no need for extra overlays or different additions to the already confusing and bloated UI.

Can be re-explored whenever the UI will be redone...

PR #1652, opened 2023-07-08
Custom teleport sub tab

### LINKED ISSUE #1623: [Request]: Teleportation menu
### Problem

The teleportation menu is really small when we have to go through a big list. On a 1440 p resolution it is ok, but on a FHD (1920 p) there isn't a lot of space for the list and is inconvenient, we have to scroll a lot
I am still working on the list, will post it soon on UC

NB: same problem with the saved cars list : too small  (width and lenght) when u want to spawn a vehicle u saved (not the personal ones)

### Solution

Maybe putting the menu in another part ?

### Reason

Usage enhancement

### Additional context
teleport

Tayipc38:
Moved custom teleport feature to its own service and respective sub window. Added a filter option and highlight on closest location.

Needs translation label.

Fixes #1623 

image


src/services/gui/gui_service.hpp
```diff
@@ -156,26 +160,27 @@ namespace big
...
+void set_nav_size(int);
+void increment_nav_size();
+void reset_nav_size();
+std::map<tabs, navigation_struct>& get_navigation();
+void remove_from_nav(tabs existing_tab_id);
+}
+;
 
-	inline gui_service* g_gui_service{};
+inline gui_service* g_gui_service{};
```
Yimura: The hell happened to indentation here?

Yimura (changes requested):

src/services/gui/gui_service.hpp
Tayipc38: Just used clang format thingy. Seems it did something weird.

Yimura (approved):

PR #1689, opened 2023-07-11
fix(mobile): PV summon timer

Yimura:
- Modified the `script_local` class to support instantiation without `rage::scrThread` or `stack`.
- Updated PV summon offset
- Moved freemode timer offset to `scr_globals.hpp`

- [x] Test if it works as intended

### DIFF
--- a/src/core/scr_globals.hpp
+++ b/src/core/scr_globals.hpp
@@ -1,5 +1,6 @@
 #pragma once
 #include "script_global.hpp"
+#include "script_local.hpp"
 
 namespace big::scr_globals
 {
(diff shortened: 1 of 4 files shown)


gir489returns:
This should be moved into scr_globals.hpp so that it's easier to maintain what needs to be updated each update.

Yimura:
> This should be moved into scr_globals.hpp so that it's easier to maintain what needs to be updated each update.

Implemented.

gir489returns:
> > This should be moved into scr_globals.hpp so that it's easier to maintain what needs to be updated each update.
> 
> Implemented.

Looks great. 👍🏻

PR #1696, opened 2023-07-12
Redesigned Debug Globals.

gir489returns:
This commit requires code review and compartmentalization designations from the team, as I couldn't be fucked to do it myself.


Yimura (approved):
Looks good to me, no complaints.

PR #1714, opened 2023-07-14
feat(Vehicle): Added vehicle dirt control

Democles85:
image


gir489returns:
Why isn't there a checkbox for enabling keep_vehicle_clean? Also, if you have Always Vehicle Repaired and keep_vehicle_clean on, they will both set the dirt level. Check for this collision.

Democles85:
> Why isn't there a checkbox for enabling keep_vehicle_clean? Also, if you have Always Vehicle Repaired and keep_vehicle_clean on, they will both set the dirt level. Check for this collision.

I do have a checkbox for keep vehicle clean. I will add the check for that collision, thanks

Edit: The checkbox is in the main view for the Vehicle Controls, thought that would be better to have it there, and if you want to control the dirt level, you just use Fun Features.

Yimura (dismissed):
Waiting for a PR on the translation repo before I add it.

Democles85:
> Waiting for a PR on the translation repo before I add it.

Added it!

Yimura (approved):

PR #1717, opened 2023-07-14
feat(Protections): Added buttons for controlling the states of protections

### LINKED ISSUE #1058: [Request]: Buttons for quickly enabling / disabling default protections
### Problem

Due to how aggressive some protections can be it may be a good bandaid to have them all disabled / enabled with a single user action through disable all / enable default enabled protections when playing only with friends / doing missions 

### Solution

A better way would be to make sure protections never interfer when playing missions with friends but that sounds unrealistic

### Reason

Minor qol

### Additional context

_No response_

Democles85:
## Closes #1058 
image


xiaoxiao921:
image

that's how i'd do it, though idk if everyone like that kind of hacks

Yimura:
> image
> 
> that's how i'd do it, though idk if everyone like that kind of hacks

I would've probably done the same, and it's pretty safe to do.

Democles85:
> image
> 
> that's how i'd do it, though idk if everyone like that kind of hacks

Thanks, was trying to do something with loops, but didn't know this was possible

Yimura (approved):

PR #1732, opened 2023-07-16
Fixed issue #1727 DIRT_LEVEL is calling natives inside the presentati…

gir489returns:
…on layer.

### DIFF
--- a/src/views/vehicle/view_fun_vehicle.cpp
+++ b/src/views/vehicle/view_fun_vehicle.cpp
@@ -136,28 +136,22 @@ namespace big
 
 		ImGui::SeparatorText("DIRT_LEVEL"_T.data());
 		{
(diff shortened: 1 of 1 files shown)


src/views/vehicle/view_fun_vehicle.cpp
```diff
@@ -136,28 +136,22 @@ namespace big
 
 		ImGui::SeparatorText("DIRT_LEVEL"_T.data());
 		{
-			if (!ENTITY::DOES_ENTITY_EXIST(self::veh))
+			auto vehicle = (CVehicle*)g_pointers->m_gta.m_handle_to_ptr(self::veh);
```
Yimura: You can probably switch this to `g_local_player->m_vehicle`.

Yimura (changes requested):

Yimura (approved):

PR #1745, opened 2023-07-17
fix(Protections): Fixed how the reset protections is handled

Democles85:


src/views/settings/view_protection_settings.cpp
```diff
@@ -8,10 +8,13 @@ namespace big
 			*(bool*)i = state;
 	}
 
-	void view::protection_settings()
+	static inline void reset_protections()
 	{
-		auto initial_protections = g.protections;
+		g.protections = {};
```
gir489returns, line 13: Won't this just zero out the protections struct?

Yimura: Nope that should default construct the object, as long as every field has a value it'll set it correctly.

xiaoxiao921: https://godbolt.org/z/49oqszh63

xiaoxiao921 (approved):

Yimura (approved):

PR #1750, opened 2023-07-17
Update cpr.cmake

BugisoftRSG:
Disables the cpr/curl tests

No more 10 min tests


BugisoftRSG:
Not possible without changing the curl source and forking both cpr and curl. Will close this for now.

PR #1751, opened 2023-07-17
feat(CI): swap to self-hosted

Yimura:
Draft as a last resort if CI keeps failing with Github provided runners.


Yimura:
People that don't have self-hosted runners available on their account should not merge this if they're using Github action runners.

It's unlikely that Github action runners will work anyways...

PR #1759, opened 2023-07-18
feat(IssueTemplates): Update issue templates and added 2 more

Yimura:


xiaoxiao921:
I think dragging `%appdata%/YimMenu/cout.log` should be required for the issue to be valid.

xiaoxiao921:
https://github.com/YimMenu/YimMenu/pull/1759/files#diff-0c9d0c3b2b7670c04f5e728c478a154bfdd1f7105288e692b366df8015249f35R24

Label: Store seems wrong, also maybe instead a checkbox that would say i'm using latest nightly instead?

xiaoxiao921 (approved):

PR #1772, opened 2023-07-19
Closes #1676 Improve modder detection reaction

### LINKED ISSUE #1676: [Request]: Improve modder detection reaction
### Problem

You recently added player detections such as desync protection, spoofed host token, attacked a player using a mod year, and so on.

### Solution

It would be appropriate to combine this with the "Modder detected by antivirus" reaction so that they can be automatically kicked and blocked from joining and that there would be an option not to add them to the player database, since the table quickly becomes clogged to no avail.
screenshot

### Reason

 I think this is much easier to do than filtering and sorting the player database

### Additional context

_No response_

ghost:


xiaoxiao921:
Just put `g.reactions.modder_detection.process(player);` inside `session::add_infraction`, no?

ghost:
I don't understand how to do it but this pull request works fine for me.  I'm just not very versed in all the code

xiaoxiao921:
Well you are basically repeating `g.reactions.modder_detection.process(player);` 7 times when you could just put `g.reactions.modder_detection.process(player);` inside the `session::add_infraction` function

I don't know how else to put it

ghost:
Can you confirm this?

xiaoxiao921 (dismissed):

Yimura:
This gonna be a problem with false positives, @maybegreat48 do we add infractions for false positives? I don't think we do but want to verify it.

xiaoxiao921:
> This gonna be a problem with false positives, @maybegreat48 do we add infractions for false positives? I don't think we do but want to verify it.

Then just don't have false positive?

Yimura (changes requested):
The `modder_detection` object doesn't use the appropriate texts as defined in `settings.hpp:L212`.
```cpp
reaction modder_detection{"Modder Detection", "%s is detected as a modder by the anticheat", "%s is detected as a modder by the anticheat"};
```

It'll probably be best for you to rename the `modder_detection` reaction object to something like `anti_cheat_detection`. Create a new object for this single purpose as a catch-all kick.

Something like:
```cpp
reaction general_detection{"General Modder Detection", "%s is detected as a modder", "%s is detected as a modder"};
```

xiaoxiao921:
@kbabrmover1995 i'll go ahead and finish the pr

Yimura (approved):

PR #1776, opened 2023-07-20
Don't allow friends to be blocked from joining(maybe)

Rxann:
also update the assign_player_index block join to use the translation instead of pure text.

### DIFF
--- a/src/hooks/protections/handle_join_request.cpp
+++ b/src/hooks/protections/handle_join_request.cpp
@@ -10,8 +10,9 @@ namespace big
 {
 	bool hooks::handle_join_request(Network* network, rage::snSession* session, rage::rlGamerInfo* player_info, CJoinRequestContext* ctx, BOOL is_transition_session)
 	{
-		if (auto player = g_player_database_service->get_player_by_rockstar_id(player_info->m_gamer_handle.m_rockstar_id);
-		    player && player->block_join)
+		if (auto player = g_player_database_service->get_player_by_rockstar_id(player_info->m_gamer_handle.m_rockstar_id); player
+		    && player->block_join && !g.session.allow_friends_into_locked_session
+		    && !g_player_service->get_by_host_token(player_info->m_host_token)->is_friend())
 		{
 			CMsgJoinResponse response{};
 			response.m_status_code = player->block_join_reason;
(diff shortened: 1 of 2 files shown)


src/hooks/protections/handle_join_request.cpp
```diff
@@ -10,8 +10,9 @@ namespace big
 {
 	bool hooks::handle_join_request(Network* network, rage::snSession* session, rage::rlGamerInfo* player_info, CJoinRequestContext* ctx, BOOL is_transition_session)
 	{
-		if (auto player = g_player_database_service->get_player_by_rockstar_id(player_info->m_gamer_handle.m_rockstar_id);
```
BugisoftRSG, line 13: The problem is not that it can't fetch the player via rockstar id, instead there are multiple protections against blocked joins not just in handle_join_request and assign_physical_index. Also in [send_non_physical_player_data.cpp](https://github.com/YimMenu/YimMenu/blob/7e50d5377a29206d8f20b1d740230f9214456809/src/hooks/protections/send_non_physical_player_data.cpp#L15)

Rxann: Okay, in this file specifically I just added this to prevent anything from blocking them in this file. Do you believe adding something like this in the other file would prevent friends also?

gir489returns: Can you not read? He's saying there are multiple places that we protect against people joining (and even with that spider's nest of crap the project does, they can still come through). All you did was change this one file and submit it. What the fuck are you doing? Did you even test this? I hate to ask, and I probably know the answer, do you even have friends to use this with? Test your fucking code, have them join in a closed session and see if they can't join. If they can't, then you know this doesn't work. What's your end goal here? You just want a contributor gold star on this repository or something for class? What are you even doing with your life?

Rxann: 1: Yes I can read. 2: I know what the hell he was saying 3: I submit it because I believed it worked 4: Yes actually I have friends to play with and I play with friends all the time 5: I just want to have some fun and try to collaborate to a project I like. And also, you should probably be a little less rude because it's a dick move.

PR #1775, opened 2023-07-19
lua: sub folders support, check for .lua file extension, refactor and simplify internals.

xiaoxiao921:

PR #1779, opened 2023-07-20
Custom Sky Color and unify some stuff

BugisoftRSG:
- Added Custom Sky Color (currently Azimuth East/West/Transition + Zenith + Stars Intensity)
- Unified the use of std:: for integer types (sometimes it was std::uint32_t, sometimes uint32_t)
- Unified pointers.hpp since initializing something to nullptr would never prevent any error or helps in any other way
- Readded settings for Rainbow controller (maybe change that?)
- Allow the usage of .luac (Compiled lua scripts), but only if explictly activated. (Maybe discuss this a bit, we all know they dont care even if "Use at your own risk" is warned, #1775 may also allow that then) [EDIT: Removed]

- Waiting for GTAV-Classes merge so I can change gtav-classes.cmake again


gir489returns:
Luac should not be allowed at all. The only reason for it is to hide code. Unless we implement encryption; which would be useless on an open source project, the overhead for compiling a script at runtime does not outweigh the potential risk of malicious scripts. When I used to hack PAYDAY 2, there was some menu called HoxHud I think and they thought they were slick too hiding malicious stuff in LuaC, while a dedicated team of reverse engineers can break the compiled code and put it back to Lua, that should not be necessary as the typical user isn’t going to do that, and having a checkbox to stop potential abuse isn’t going to help.

scripts/gtav-classes.cmake
```diff
@@ -2,8 +2,8 @@ include(FetchContent)
 
 FetchContent_Declare(
     gtav_classes
-    GIT_REPOSITORY https://github.com/Yimura/GTAV-Classes.git
-    GIT_TAG        5ef3545feb947a1bd143a84d15fdc624e1366c18
+    GIT_REPOSITORY https://github.com/YimMenu/GTAV-Classes.git
```
Yimura: Gonna be a bad time if you use that one.

src/views/debug/view_debug_misc.cpp
```diff
@@ -10,13 +10,16 @@
...
 #include "view_debug.hpp"
+#include "gta/color.hpp"
 
 #include "hooking.hpp"
 
 namespace big
 {
 	void debug::misc()
 	{
+		rage::rgbaColor a;
```
Yimura: Unused local var

src/views/debug/view_debug_misc.cpp
```diff
@@ -10,13 +10,16 @@
 #include "util/ped.hpp"
 #include "util/system.hpp"
 #include "view_debug.hpp"
+#include "gta/color.hpp"
```
Yimura: Unused include

Yimura (changes requested):

src/views/debug/view_debug_misc.cpp
BugisoftRSG: true, used them to test the classes, its not used anywhere maybe delete since there is also ImU32 or just float[4]?

Yimura (approved):

PR #1785, opened 2023-07-21
Fix REQUEST_CONTROL_EVENT for missions.

gir489returns:


Yimura (commented):
Are all of these hooks in required to have the check for if the camera is faded out?

gir489returns:
> Are all of these hooks in required to have the check for if the camera is faded out?

No, I wanted to see if that fixed the issue, then it would move it one by one until it comes back, then I’ll know which function is the problem child.

gir489returns:
> You can actually use natives outside of game threads as long as they don't use the TLS context or run in the GUI thread

Thanks. I resorted to using the pointer method as I was just copying it from BBv2. If this does end up fixing the problem and isolate which hook is causing it, I'll see if we can use the natives.

gir489returns:
This failed to address #1788. More solution ideas might be required.

gir489returns:
Still failed to address https://github.com/YimMenu/YimMenu/issues/1788. I'm out of ideas.

gir489returns:
@maybegreat48 can you look at https://github.com/YimMenu/YimMenu/pull/1785/commits/2ff899a3724d449eb3a94dadef594e2cdbdeb5ce and see if any of the functions I put the fade/cutscene check in could be potentially problematic? Tomorrow I'm going to add some logging to see which are actually called during the loading of the mission, but I was able to play the Casino Heist from start to finish no problem, although that could mean nothing.

src/util/teleport.hpp
```diff
@@ -54,15 +54,16 @@ namespace big::teleport
...
 			ENTITY::SET_ENTITY_VISIBLE(hnd, false, false);
 			ENTITY::SET_ENTITY_COLLISION(hnd, false, false);
 			ENTITY::FREEZE_ENTITY_POSITION(hnd, true);
 
 			auto obj_id                      = player->get_ped()->m_net_object->m_object_id;
+			auto veh_id                      = g_pointers->m_gta.m_handle_to_ptr(hnd)->m_net_object->m_object_id;
 			remote_player_teleport remote_tp = {obj_id, {coords.x, coords.y, coords.z}};
 
-			g.m_remote_player_teleports.emplace(g_pointers->m_gta.m_handle_to_ptr(hnd)->m_net_object->m_object_id, remote_tp);
+			g.m_remote_player_teleports.emplace(veh_id, remote_tp);
```
maybegreat48, line 66: This breaks on-foot teleport. See https://github.com/YimMenu/YimMenu/blob/master/src/hooks/toxic/write_vehicle_proximity_migration_data_node.cpp

gir489returns:
> You're disabling all protections during a cutscene which is a really bad idea. Let me try to figure out what's preventing the heist from working. The clone create and sync hooks can be safely ruled out, so the received event hook must be broken

I changed it to only do it on missions.

src/util/teleport.hpp
gir489returns: I tested it and it works. This fixes the bug from #1787. https://github.com/YimMenu/YimMenu/blob/master/src/util/teleport.hpp#L65C4-L65C118 You are emplacing the vehicle's network ID but at the bottom of the function https://github.com/YimMenu/YimMenu/blob/master/src/util/teleport.hpp#L93 are checking if the vehicle's network ID is obj_id which was set to player->get_ped()->m_net_object->m_object_id, so it never erases the entry from the list, causing the bug to appear.

maybegreat48: Didn't notice that, thanks for fixing

gir489returns:
@maybegreat48 You seemed to be correct, received_event is causing this, and that would make sense, since I was doing the same thing you are doing in BBv2 and I had the same bug. You can reference my hook https://bitbucket.org/gir489/bigbasev2/src/master/BigBaseV2/src/hooks/NetEventHook.cpp to see what I was doing and compare it to yours and maybe rule some stuff out.

gir489returns:
@maybegreat48 I have deduced the bug is coming from GIVE_CONTROL_EVENT. It is the only network event called that is being rejected by the menu during the cutscenes or when my camera is faded out.

Yimura (approved):

PR #1790, opened 2023-07-21
feat(debug): log msvc c++ exception properly.

xiaoxiao921:


gir489returns (approved):
This is a very good and elegant solution to catch C++ exceptions without having to resort to nonsense like RTTI. Although we really don't concern ourselves with speed when catching exceptions and more so with accuracy, the two might be more linked than you expect, as you may not have enough time to catch the exception and dump it to file before the program explodes completely.

xiaoxiao921:
Well the unhandledexceptionfilter is supposed to be called no matter and we flush the logger so it shouldnt be possible to have 0 exception logged in the console or the file but yeah i have seen this happen still for some reason

Yimura (approved):

PR #1791, opened 2023-07-21
feat(noclip): fix aim nullifying noclip speed.

### LINKED ISSUE #1719: [Bug]: Noclip does not work in first person.
### Describe the bug

When noclip is enabled, the player is stuck in place while in first person.

### Steps To Reproduce

1. Enable Noclip.
2. Use first person camera.
3. Bug appears.

### Expected Behavior

The user should be able to move while in first person.

### Screenshots and/or Logs

_No response_

### Store

Rockstar

### Additional context

_No response_

### Build

xiaoxiao921:
closes #1719


gir489returns (approved):
Tested and working.

src/backend/looped/self/noclip.cpp
```diff
@@ -76,11 +76,10 @@ namespace big
...
 				ENTITY::FREEZE_ENTITY_POSITION(ent, false);
 
-				const auto offset = ENTITY::GET_OFFSET_FROM_ENTITY_IN_WORLD_COORDS(ent, vel.x, vel.y, 0.f);
-				vel.x             = offset.x - location.x;
-				vel.y             = offset.y - location.y;
+				const auto offset = ENTITY::GET_OFFSET_FROM_ENTITY_IN_WORLD_COORDS(ent, vel.x * m_speed_multiplier, vel.y * m_speed_multiplier, vel.z * m_speed_multiplier);
 
-				ENTITY::SET_ENTITY_VELOCITY(ent, vel.x * m_speed_multiplier, vel.y * m_speed_multiplier, vel.z * m_speed_multiplier);
+				ENTITY::SET_ENTITY_VELOCITY(ent, 0, 0, 0);
+				ENTITY::SET_ENTITY_COORDS_NO_OFFSET(ent, offset.x, offset.y, offset.z, true, true, true);
```
Yimura: This will completely downgrade noclip for other players in the session.

No clip will become very choppy for people that are in your vehicle or from other player's point of view when noclipping. This would be a downgrade on the current noclip implementation.

I'd rather find the cause on why velocity is externally influenced or limited.

Yimura (dismissed):

src/backend/looped/self/noclip.cpp
xiaoxiao921: Not happening unless someone reverse the aiming code, it can most likely be done by just getting the writers to the player velocity vector but I can't be bothered to do that, I can make a simple if statement that check if the player is aiming and fallback to SET_ENTITY_COORDS_NO_OFFSET and otherwise set the entity veocity but that's about it

gir489returns: Noclip already is very choppy and buggy, because you’re rapidly changing direction, the interpolation doesn’t have time to fill in the gaps between frames. You’re moving way too fast for that anyway.

Yimura: I've already nopped all code that writes to the player velocity addresses, no change in that.
There's something else doing funky stuff.

Yimura: Perhaps we can go with a middle ground of both worlds, have vehicle noclip use calls to the velocity native. And for noclipping the ped we can use the normal noclip because the game isn't setup to frequently update and lerp our position properly for that.

Yimura (approved):

Yimura:
Waiting for translations before I merge it.

PR #1797, opened 2023-07-21
No Max Distance For Vehicle Summon

Rxann:
Tested. Drove all the way around the map autonomously.


src/views/view_vehicle_control.cpp
```diff
@@ -331,13 +331,17 @@ namespace big
...
 		ImGui::SliderFloat("Max summon distance", &g.window.vehicle_control.max_summon_range, 10.f, 250.f);
 		if (ImGui::IsItemHovered())
 		{
 			ImGui::BeginTooltip();
 			ImGui::Text("At what range the vehicle will drive towards the summoned location as oposed to being teleported");
 			ImGui::EndTooltip();
 		}
+		ImGui::EndDisabled();
+
+		ImGui::Checkbox("No Max Summon Distance", &g.window.vehicle_control.no_max_summon);
```
Yimura, line 344: Can be made into a bool command.

Yimura (changes requested):

src/views/view_vehicle_control.cpp
GTMoraes: This is great. I love using the vehicle summon for RP purposes (like a getaway vehicle being driven to me to get out of a hairy situation), but the 250M limit sometimes breaks my immersion.

Hope this gets into the main build soon!

Rxann: > Can be made into a bool command.

Which part of it? the whole thing(summoning)?

Yimura: The line I highlighted, it can be changed into a `bool_command`.

Yimura:
```cpp
// register command
bool_command g_ignore_max_summon_distance("vehcontrolignoremaxsummondist", "Ignore Max Summon Distance", "Enabling this will allow summoning your last vehicle from any distance", g.window.vehicle_control.no_max_summon);

// Show checkbox in UI
components::command_checkbox<"vehcontrolignoremaxsummondist">();
```

It's not that hard, here is another example from in the codebase:
https://github.com/YimMenu/YimMenu/blob/be7bf8ffa6f969c1850bd44b5c1b8d4c5ba7bb66/src/views/view_vehicle_control.cpp#L318-L330

Rxann: Okay, thanks. I have been a bit busy but I will get to it.

Yimura:
PR has gone completely stale.

PR #1813, opened 2023-07-23
Force relay connections

maybegreat48:
* Added Force Relay Connections, which will prevent direct connections to hide your IP address. This has the side effect of hiding other players' IP addresses


src/hooks/protections/can_apply_data.cpp
```diff
@@ -651,7 +651,7 @@ namespace big
 		if (!model_info::get_model(model))
 			return false;
 
-		if (!model_info::is_model_of_type(model, eModelType::Object, eModelType::Time, eModelType::Weapon, eModelType::Destructable, eModelType::WorldObject, eModelType::Sprinkler, eModelType::Unk65, eModelType::Plant, eModelType::LOD, eModelType::Unk132, eModelType::Building))
```
gir489returns, line 654: Was this removed intentionally?

maybegreat48: Yeah, but I'll add it back if I see false positives

Yimura (approved):

PR #1824, opened 2023-07-23
refactor: make window hook into a toggle command

Yimura:
I think this should fix the issue with AHK script users that had their scripts suddenly stop working.

### DIFF
--- a/src/backend/commands/system/window_hook.cpp
+++ b/src/backend/commands/system/window_hook.cpp
@@ -1,33 +1,31 @@
-#include "backend/looped_command.hpp"
+#include "backend/bool_command.hpp"
 #include "memory/byte_patch.hpp"
 #include "pointers.hpp"
 
 namespace big
 {
-	class window_hook : looped_command
+	class window_hook : bool_command
 	{
-		using looped_command::looped_command;
+		using bool_command::bool_command;
 
 		memory::byte_patch* m_window_hook_patch;
 
-		virtual void on_enable() override
+		virtual void refresh() override
 		{
 			static auto patch = (m_window_hook_patch = memory::byte_patch::make(g_pointers->m_gta.m_window_hook.as<void*>(), std::to_array({0xC3, 0x90, 0x90, 0x90}))
 			                                               .get(),
 			    true);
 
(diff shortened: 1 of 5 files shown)


xiaoxiao921:
formatting looks fucked, other than that looks good to me

src/backend/commands/system/window_hook.cpp
```diff
@@ -0,0 +1,32 @@
+#include "backend/looped_command.hpp"
+#include "memory/byte_patch.hpp"
+#include "pointers.hpp"
+
+namespace big
+{
+	class window_hook : looped_command
```
xiaoxiao921: nevermind, why is this a looped_command?

Yimura: Bool command doesn't have an on_enable/on_disable overrides.

image

xiaoxiao921: https://github.com/YimMenu/YimMenu/blob/ddb12d6e2bc9ec0768550f6760b4459a171c0cd4/src/backend/bool_command.cpp#L13-L38

Please read the impl of bool_command::execute, this will make sense once you do

you need to impl refresh and then check the flag accordingly

gir489returns:
I think this should only be included for dev/debug build, this has no purpose for normal users.

src/backend/commands/system/window_hook.cpp
Yimura: Well I'll be damned.

Yimura:
> I think this should only be included for dev/debug build, this has no purpose for normal users.

It's under the Debug window, in general this has nothing to do with normal users.

src/backend/commands/system/window_hook.cpp
Yimura: There are a bunch of commands implemented in this way. I think maybegreat said something along the lines of implementing it like this but my brain my be failing me.

xiaoxiao921: that doesnt look like correct usage, i'd even argue that on_enable and on_disable should be moved to bool_command,

xiaoxiao921 (approved):

PR #1826, opened 2023-07-23
refactor!: Modify command argument handling and move `on_` method to bool_command

Yimura:
## Changes
 - Added a `command_arguments` class to help with argument parsing
 - Moved `on_enable/on_disable` methods from `looped_command` to `bool_command`

## Fixes
 - Don't enable context menu while the cmd executor is active
 - Disable freecam and noclip movement while the other is active (as well as disable them when the cmd executor is active)

## Todo
 - [ ] ~~Re-implement arguments with `std::any`~~
 - [x] Rename pop to shift to reflect the internal functionality
 - [x] Take a large enough sample size of commands to test if they behave as expected
 - [x] Fix `bool_command::on_enable` doesn't get called on menu init

### DIFF
--- a/src/backend/backend.cpp
+++ b/src/backend/backend.cpp
@@ -19,6 +19,8 @@ namespace big
 {
 	void backend::loop()
 	{
+		for (auto& command : g_bool_commands)
+			command->refresh();
 		for (auto& command : g_looped_commands)
 			command->refresh();
 
--- a/src/backend/backend.cpp
+++ b/src/backend/backend.cpp
@@ -21,8 +21,6 @@ namespace big
 	{
 		for (auto& command : g_bool_commands)
 			command->refresh();
-		for (auto& command : g_looped_commands)
-			command->refresh();
 
 		register_script_patches();
 
--- a/src/backend/command_arguments.hpp
+++ b/src/backend/command_arguments.hpp
@@ -13,7 +13,7 @@ namespace big
     private:
         const std::size_t m_argument_count;
         std::vector<uint64_t> m_argument_data;
-        std::size_t m_idx;
+        mutable std::size_t m_idx;
 
     public:
         command_arguments(std::size_t argument_count = 0) :
@@ -52,7 +52,7 @@ namespace big
 
         template<typename T = uint64_t>
             requires ArgumentLimit<T>
-        T pop()
+        T shift() const
         {
             if (m_idx >= m_argument_count)
             {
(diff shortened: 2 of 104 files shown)


src/backend/command_arguments.hpp
```diff
@@ -0,0 +1,113 @@
...
+        template<typename T = uint64_t>
+            requires ArgumentLimit<T>
+        std::enable_if_t<std::is_pointer_v<T>, const T> pop() const
+        {
+            if (m_idx >= m_argument_count)
+            {
+                throw std::runtime_error("Attempted to pop argument beyond allocated argument size.");
+            }
+
+            return static_cast<T>(m_argument_data[m_idx++]);
```
xiaoxiao921: `m_idx++`

is this correct?

src/backend/command_arguments.hpp
```diff
@@ -0,0 +1,113 @@
...
+        template<typename T = uint64_t>
+            requires ArgumentLimit<T>
+        T pop()
+        {
+            if (m_idx >= m_argument_count)
+            {
+                throw std::runtime_error("Attempted to pop argument beyond allocated argument size.");
+            }
+
+            return reinterpret_cast<const T&>(m_argument_data[m_idx++]);
```
xiaoxiao921, line 62: `m_idx++`

is this correct?

src/backend/command_arguments.hpp
Yimura: Will be renamed to `shift`.

src/backend/command_arguments.hpp
Yimura: Will be renamed to `shift`.

src/backend/backend.cpp
```diff
@@ -19,6 +19,8 @@ namespace big
 {
 	void backend::loop()
 	{
+		for (auto& command : g_bool_commands)
+			command->refresh();
 		for (auto& command : g_looped_commands)
 			command->refresh();
```
xiaoxiao921: double refresh?

Yimura: One for bool commands and one for looped commands.

src/backend/command_arguments.hpp
```diff
@@ -0,0 +1,114 @@
...
+            return static_cast<T>(m_argument_data[idx]);
+        }
+
+        template<typename T = uint64_t>
+            requires ArgumentLimit<T>
+        T shift() const
+        {
+            if (m_idx >= m_argument_count)
+            {
+                throw std::runtime_error("Attempted to pop argument beyond allocated argument size.");
```
xiaoxiao921: the exception says `pop` still

xiaoxiao921 (approved):

PR #1834, opened 2023-07-24
fix(hooks): return types

ghost:
fixes a bunch of return types that were seemingly copied from IDA without manual verification

### DIFF
--- a/src/backend/commands/session/wipe_session.cpp
+++ b/src/backend/commands/session/wipe_session.cpp
@@ -1,0 +1,21 @@
+#include "backend/player_command.hpp"
+#include "network/CNetworkPlayerMgr.hpp"
+#include "pointers.hpp"
+#include "services/players/player_service.hpp"
+
+namespace big
+{
+	class empty_session : command
+	{
+		using command::command;
+		virtual void execute(const std::vector<uint64_t>&, const std::shared_ptr<command_context> ctx)
+		{
+			g_player_service->iterate([](const player_entry& player) {
+				auto mgr = *g_pointers->m_gta.m_network_player_mgr;
+				mgr->RemovePlayer(player.second->get_net_game_player());
+			});
+			g_notification_service->push_success("Empty Session", "Completed");
+		}
+	};
+	empty_session g_empty_session("emptysession", "Remove All Players From Session", "Removes everyone from the session", 0);
+}
--- a/src/hooking.hpp
+++ b/src/hooking.hpp
@@ -70,7 +70,7 @@ namespace big
 		static bool init_native_tables(rage::scrProgram* program);
 		static rage::eThreadState script_vm(uint64_t* start_stack, uint64_t** scr_globals, rage::scrProgram* program, rage::scrThreadContext* ctx);
 
-		static void network_player_mgr_init(CNetworkPlayerMgr* _this, uint64_t a2, uint32_t a3, uint32_t a4[4]);
+		static bool network_player_mgr_init(CNetworkPlayerMgr* _this, uint64_t a2, uint32_t a3, uint32_t a4[4]);
 		static void network_player_mgr_shutdown(CNetworkPlayerMgr* _this);
 
 		static bool fragment_physics_crash_2(float* a1, float* a2);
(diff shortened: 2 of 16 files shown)


Yimura (approved):
Thanks for your contribution and fixing these things.

xiaoxiao921:
The change for `RECEIVED_CLONE_CREATE` is clearly wrong, as the return function clearly returns 0 or 1 and an if statement depends on that. Now you could be right that the RECEIVED_CLONE_CREATE caller does indeed not check that return value then, but it feels terribly wrong to change the whole function definition just because the caller decided to discard the return value.

maybegreat48:
The child function that the return function calls clobbers `rax`, and IDA thinks that it's returning that value. I would say that the evidence is inconclusive, but the fact that the return value isn't used by the caller makes me lean to a no

xiaoxiao921:
My point was that in this specific case there is absolutly no harm on letting the function as is. (like every other changes in that PR), it's just for the sake of being correct but here the correct doesn't even matter?

maybegreat48:
> the correct

There's no way to know for sure with the GTA dump alone

ghost:
> The change for `RECEIVED_CLONE_CREATE` is clearly wrong, as the return function clearly returns 0 or 1 and an if statement depends on that. Now you could be right that the RECEIVED_CLONE_CREATE caller does indeed not check that return value then, but it feels terribly wrong to change the whole function definition just because the caller decided to discard the return value.

Blindly trusting IDA pseudocode is exactly why there were so many wrong types. Although in this specific instance you can make the argument that it is impossible to be 100% certain with just a game dump.
Even if it were to return a value, it isn't used, so why do you care?


> My point was that in this specific case there is absolutly no harm on letting the function as is. (like every other changes in that PR), it's just for the sake of being correct but here the correct doesn't even matter?

Why do you focus so hard on this specific instance where it doesn't make a difference instead of being happy that I've fixed much more than that where it can indeed break shit if R* uses the return value in a future update or changes the logic?

xiaoxiao921:
@maybegreat48 Is that original function now being called fine here? https://github.com/YimMenu/YimMenu/pull/1834/files#diff-c40e2cdd806d2c4034c85d8fb885194da4d7c5f928d0da1e207010da625e1d2dR50

@MegaMemeMoment i've checked all functions you changed and you did make a potential future compat change for `network_player_mgr_init` and `send_non_physical_player_data`, but you did the opposite for `write_player_camera_data_node` and `write_player_game_state_data_node`, so i'm not sure really?

gir489returns:
I guess if you want to be super paranoid you can always return __int64 from every function which would just copy RAX from the original call and return it to the original calle.

ghost:
> @MegaMemeMoment i've checked all functions you changed and you did make a potential future compat change for `network_player_mgr_init` and `send_non_physical_player_data`, but you did the opposite for `write_player_camera_data_node` and `write_player_game_state_data_node`, so i'm not sure really?

Well those data node functions definitely do not return a value.

xiaoxiao921:
> I've fixed much more than that where it can indeed break shit if R* uses the return value in a future update or changes the logic?

If the goal of this pr is to make the code more reliable for future updates, can you explain how doing the opposite for these data node functions achieves that?

Yimura:
> If the goal of this pr is to make the code more reliable for future updates, can you explain how doing the opposite for these data node functions achieves that?

Well then we fix it by the time it comes around, R* has updated functions in the past. If it breaks it breaks.

xiaoxiao921:
ooooh right right, so exactly what was the purpose of all this? 🤣

ghost:
@xiaoxiao921 It's not only about making it more reliable, I simply like to have these be as accurate/correct as possible. And as I said before, these data node functions definitely do not return a value and most likely won't, so it really shouldn't matter.
I dislike just slamming random stuff in and calling it a day, and wanted to contribute my fixes/research to this project.

maybegreat48:
> @maybegreat48 Is that original function now being called fine here? https://github.com/YimMenu/YimMenu/pull/1834/files#diff-c40e2cdd806d2c4034c85d8fb885194da4d7c5f928d0da1e207010da625e1d2dR50
> 
> @MegaMemeMoment i've checked all functions you changed and you did make a potential future compat change for `network_player_mgr_init` and `send_non_physical_player_data`, but you did the opposite for `write_player_camera_data_node` and `write_player_game_state_data_node`, so i'm not sure really?

Didn't see that. It's probably a good idea to revert that change. Some metrics can flag your account

PR #1887, opened 2023-07-29
fix(session): sctv spectator not being enabled

Londiuh:
The global index to enable spectator mode was outdated.


gir489returns:
Can you instead move this to https://github.com/YimMenu/YimMenu/blob/c54be57469fbb62ad6bb3f3cd7ad4d0d8f9cf510/src/core/scr_globals.hpp#L7?

This will help us know which globals need to be updated. Include any notes you think the person updating it would need to know, such as RegExs, the script it belongs to, potential false-positive search results, etc.

Yimura (approved):
Thanks for implementing the requested changes.

PR #1893, opened 2023-07-30
Faster session change through yim menu

NjuuOwO:
Being at +=751 z height makes no these character switch jumps into sky so its little faster and its a neat feature ig


NjuuOwO:
these commits sorry first time with these pull request etc

PR #1894, opened 2023-07-30
Faster change session through menu

NjuuOwO:
+750 z makes no jumpy jumps into sky when changing sessions, neat feature ig

### DIFF
--- a/src/util/session.hpp
+++ b/src/util/session.hpp
@@ -37,6 +37,8 @@ namespace big::session
 
 	inline bool join_type(eSessionType session)
 	{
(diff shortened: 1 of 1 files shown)


Yimura:
Isn't it possible to remove the swoop up camera movement? No need for a jank implementation like this.

gir489returns:
> Isn't it possible to remove the swoop up camera movement? No need for a jank implementation like this.

And even if it wasn't stupid, you can just get the position of the player from self::pos. IDK why he's referencing self::ped and not self::pos.

PR #1899, opened 2023-07-31
Closes #1664

### LINKED ISSUE #1664: [Request]: Debug script hosts
### Problem

When viewing a thread in Debug -> Threads, there should be an option to inspect who is currently hosting the thread, and a button to attempt to steal their host status from them.

### Solution

The native NETWORK_GET_HOST_OF_SCRIPT should give you an idea of how to read the host variable from a scrThread. The way I was doing it, was there was a class at 0x118 called scriptHandlerNetComponent. scriptHandlerNetComponent contained a list of CNetComponentOwners at 0x30. CNetComponentOwners contains a CNetGamePlayer pointer at 0x10. So from CNetGamePlayer you could do something like get_name or ID. 

### Reason

Debugging force script control is a pain in the ass by just spamming the log with log.info.

### Additional context

_No response_

gir489returns:
Closes #1664

### DIFF
--- a/src/views/debug/view_debug_threads.cpp
+++ b/src/views/debug/view_debug_threads.cpp
@@ -32,6 +32,27 @@ namespace
 
 namespace big
 {
+	static void set_clipboard(const char* message)
+	{
+		HGLOBAL h;
+		LPVOID p;
+		int size;
+		//calc the num of unicode char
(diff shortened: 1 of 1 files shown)


Yimura (approved):
Cursed way of setting clipboard data, can be improved on later (if there's even a better way). It's part of a debug feature anyways so doesn't matter too much.

PR #1908, opened 2023-08-01
Disable path traversal for Project File/Folder and added Session Switch command

### LINKED ISSUE #1888: [Improvement]: Add session switcher command to lua
### Which feature

I think it should be a command with a numeric argument like 1 is JOIN_PUBLIC, 2 is NEW_PUBLIC and so on.

### Reason

Such a thing will be much more convenient than using globals, although I did not succeed with them :(

### Additional context

_No response_

Yimura:
Probably still need to regenerate the Lua docs for commands.

Closes #1888

### DIFF
--- a/src/backend/commands/session/join_session.cpp
+++ b/src/backend/commands/session/join_session.cpp
@@ -26,26 +26,26 @@ namespace big
 			std::string valid_args;
 			if (was_session_string)
 			{
-                for (const auto& session_type_string : m_session_types | std::ranges::views::values)
-                {
-                    if (!empty(valid_args))
-                    {
-                        valid_args += ", ";
-                    }
-                    valid_args += session_type_string;
-                }
+				for (const auto& session_type_string : m_session_types | std::ranges::views::values)
+				{
+					if (!empty(valid_args))
+					{
+						valid_args += ", ";
+					}
+					valid_args += session_type_string;
+				}
 
-			    return valid_args;
+				return valid_args;
 			}
-            
-            for (const auto& session_type_id : m_session_types | std::ranges::views::keys)
-            {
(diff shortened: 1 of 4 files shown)


gir489returns:
Why is the CLOSED_FRIENDS scenario labeled NEWFRIENDS?

ghost:
Shouldn't NEWPUB create a solo public session? When I run the `command.call("joinsession", {"NEWPUB"})` command then I get into an open session with other people. Or did I do something wrong?

Yimura:
> Shouldn't NEWPUB create a solo public session? When I run the `command.call("joinsession", {"NEWPUB"})` command then I get into an open session with other people. Or did I do something wrong?

I'll take a look on why it's behaving oddly this evening, I might have missed checking certain edge cases.

Yimura:
@kbabrmover1995 due to limitations in the current command API you need to pass the ID of the join type not the string identifier.

The following code works as intended:
```lua
test_gui = gui.get_tab("GUI_TAB_SETTINGS"):add_tab("Debug Lua")

NEW_PUB = 1

test_gui:add_button("Test Join New Pub", function()
    command.call("joinsession", { NEW_PUB })
end)
```

If you need to know which ID's exist for other session types, these are it:
```cpp
enum class eSessionType
{
	JOIN_PUBLIC, // 0
	NEW_PUBLIC, // 1
	CLOSED_CREW, // 2
	CREW, // 3
	CLOSED_FRIENDS = 6,
	FIND_FRIEND    = 9,
	SOLO, // 10
	INVITE_ONLY, // 11
	JOIN_CREW, // 12
	SC_TV, // 13
	LEAVE_ONLINE = -1
};
```

PR #1919, opened 2023-08-04
Fixed Vehicle using a decorator that impacted removing the vehicle fr…

gir489returns:
…om the pool.

Fixed Teleport to Selected not appearing in hotkeys list.

### DIFF
--- a/scripts/gtav-classes.cmake
+++ b/scripts/gtav-classes.cmake
@@ -2,8 +2,8 @@ include(FetchContent)
 
 FetchContent_Declare(
     gtav_classes
-    GIT_REPOSITORY https://github.com/Yimura/GTAV-Classes.git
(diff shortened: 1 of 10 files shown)


Yimura (commented):
This PR makes use of a field outside of the CVehicle to store some data, I'm aware this field will only contain a value as long as the vehicle remains in render distance.

We'll have to either add CAutomobile, cast to that and write the field or find another different unused field in CVehicle to temporarily write data to...
Still a jank solution in the end, decorators aren't a solution either because this is synced over the network.

gir489returns:
I propose we move to a static map of something. Either the Vehicle ID or a pointer, but the map would need a demon to maintain the validity of the list and remove stale/deprecated entries.

Issue #1876, opened 2023-07-28
[Request]: Personal Vehicle Control / Favorite Vehicle Control


GTMoraes:
### Problem

Currently, when you open the Vehicle Controls, it only controls the last vehicle you drove.  
Sometimes the last vehicle you drove was mission related, and you don't want to control it.

### Solution

A way to set a specific vehicle as favorite, or have your current Personal Vehicle as a permanent favorite.

### Reason

Most of the time, the last vehicle you drove was something related to a mission, thus you lose control of the actual vehicle you want to control.

### Additional context

I'm mostly requesting because I like using the Summon Vehicle function. I've found a PR with a way to remove the summon distance and I have a blast summoning the vehicle from afar.

However, most of the time I'd use the feature, was because I had to get in another vehicle and left my previous somewhere. Having a way to summon and it drive to me would be amazing.

Tayipc38:
Reasonable request. Ill probably address this in a slightly larger PR than just this feature.

dynamoNg:
also can we have radio control as well inside `vehicle > vehicle controller ` if its related to that.

GTMoraes:
> Reasonable request. Ill probably address this in a slightly larger PR than just this feature.

This is great! I'm looking forward to it.

So much so, that I wonder if you've already solved it, but will release it in a larger batch change.
I'd be happy to try to implement just that feature on my own fork (keyword: try. I'm a tinkerer, not a programmer) if you could provide the code.

Thanks for considering my suggestion, by the way!

Tayipc38:
> > Reasonable request. Ill probably address this in a slightly larger PR than just this feature.
> 
> This is great! I'm looking forward to it.
> 
> So much so, that I wonder if you've already solved it, but will release it in a larger batch change. I'd be happy to try to implement just that feature on my own fork (keyword: try. I'm a tinkerer, not a programmer) if you could provide the code.
> 
> Thanks for considering my suggestion, by the way!

It is a very simple addition to the code. But the feature itself is rather a letdown as is, so I've just now started working on it. Won't be revamping it completely, just some small changes and additions like this request here.

PR #1926, opened 2023-08-05
Vehicle control additions & fixes

Tayipc38:
Ill draft my current changes and take any suggestion into consideration.

Changes:
- [X] Improve the method of playing animations
- [X] Convert UI elements to components to make use of in-built fiber pool usage.
- [X] Remove summon distance (Nobody seemed to need it lol).
- [X] Allow vehicles other than cars.
- [x] Introduce a method of manual controlled vehicle selection #1876.
- [x]  Optional vehicle information.
- [x] Patch bugs.
- [x] Other suggestions...


GTMoraes:
I'm using the changes on my own fork.

Few bugs I noticed so far:
- Selecting to control the Closest Vehicle and then choosing to control the Personal Vehicle, without a spawned personal vehicle, crashes the game.
- When selected to control the Personal Vehicle, the checkmark to turn engine on/off is bugged. It's always unchecked and doesn't change according to status. Also, it only turns the engine on, when the "Use Animation" checkbox is marked, and only turns the engine off when the same checkbox is unmarked.
Using the "Last Driven", it behaves correctly, although it doesn't turn the engine on or off, when the Use Animations is checked off (which was the behavior from before this update)

I'm still trying it out so far, but looks promising! Thanks for considering my suggestion to control other vehicles than the previous one.

Tayipc38:
> I'm using the changes on my own fork.
> 
> Few bugs I noticed so far:
> 
> * Selecting to control the Closest Vehicle and then choosing to control the Personal Vehicle, without a spawned personal vehicle, crashes the game.
> * When selected to control the Personal Vehicle, the checkmark to turn engine on/off is bugged. It's always unchecked and doesn't change according to status. Also, it only turns the engine on, when the "Use Animation" checkbox is marked, and only turns the engine off when the same checkbox is unmarked.
>   Using the "Last Driven", it behaves correctly, although it doesn't turn the engine on or off, when the Use Animations is checked off (which was the behavior from before this update)
> 
> I'm still trying it out so far, but looks promising! Thanks for considering my suggestion to control other vehicles than the previous one.

Thanks for the heads up. This definitely needs more before it can be considered remotely done.

Yimura:
@DayibBaba stale PR

Tayipc38:
> @DayibBaba stale PR

Not planning on working on this for the foreseeable future.

EDIT: The future is now old man

GTMoraes:
This PR as is, works reasonably OK.

The bugs I mentioned before are still present, and sometimes, switching from "Personal" to "Last driven" crashes the game when something that I can't really put my finger at, happens with the last driven vehicle. 
It's a rare event, but when it happens, it's after a mission and the vehicle despawns/disappears. I can read that the last vehicle model is a NULL and the game immediately crashes. Most of the time, though, even though the mission vehicle disappeared or whatever, it just displays the "last vehicle" as if it still existed, but does nothing. Next vehicle you enter, it works fine.

Last time I recall that it happened, was when I had to fly a police helicopter to take pictures of someone in a balcony. After the mission ended, I landed the chopper somewhere and when I opened the vehicle menu, it listed the last vehicle as NULL, then the game froze and crashed.   
But it also happened with normal vehicles.

Probably not ready for prime time, but looks close.

This is, easily, my most used feature from the whole menu. 99% of the time it works fine, and the known bugs I listed before are workaroundable.

Tayipc38:
> This PR as is, works reasonably OK.
> 
> The bugs I mentioned before are still present, and sometimes, switching from "Personal" to "Last driven" crashes the game when something that I can't really put my finger at, happens with the last driven vehicle. 


Happens due to the personal vehicle not existing in the script. There is no check whether one exists before switching to it. The handle is fetched from a global and casted into a Vehicle. I have this fixed on my local but have recently deleted gta to reserve space for Starfield.

src/services/vehicle/vehicle_control_service.cpp
```diff
@@ -399,13 +362,42 @@ namespace big
...
 		keep_controlled_vehicle_data_updated(m_controlled_vehicle);
+
+		if (!g.window.vehicle_control.opened)
+			return;
+
+		switch (m_selection_mode)
+		{
+		case eControlledVehSelectionMode::LAST_DRIVEN: get_last_driven_vehicle(); break;
+		case eControlledVehSelectionMode::PERSONAL: get_personal_vehicle(); break;
+		case eControlledVehSelectionMode::CLOSEST: get_closest_vehicle(); break;
```
Yimura, line 400: case, function and break on one line?

Yimura (changes requested):
Waiting for PR on translation repo.

src/services/vehicle/vehicle_control_service.cpp
Tayipc38: The format shortcut did this, not me.

Yimura (approved):
Should be ready to merge now, will handle the other PR's that still need their translations done.
Then after merging the translations this will go into master.

PR #1938, opened 2023-08-08
(feat) Seamless join. Joining without loading (sort of) - Adapted from TupoyeMenu

NjuuOwO:
removes swoop camera movement when player switch occurs

- [ ] Broken animations (eating snacks, character actions (jerk off, salute etc)... )
- [ ] Some cutscenes (fm_mission_controller) break
- [ ] Other stuff breaking


gir489returns:
This is an awful solution to a non-issue. I don't think we should merge this, not only does it add a worthless feature, but this could also break things like fm_mission_controller which relies on the camera state. You're patching like 5 things and hooking like 20 natives, if any of those break from updates we have to fix that, all so what, the camera doesn't go swoop swoop swoop?

I vote hard no.

tupoy-ya:
This feature is not finished, if it was I would have made a PR myself.

> This could also break things like fm_mission_controller which relies on the camera state.

It can, and it does sometimes, it also breaks loading into apartments, and a lot of other things.

> You're patching like 5 things and hooking like 20 natives, if any of those break from updates we have to fix that.

Haven't had anything critical broken by an update for 2-3 updates now.

> all so what, the camera doesn't go swoop swoop swoop?

Well I would prefer to keep playing the game over staring at the clouds.

Yimura:
The feature is something we could definitely add into the menu but since you didn't make it sorta feels wrong to add (even though the license of the other project allows to copy paste of features). Would you be willing to maintain it in the future or would you have to rely on @tupoy-ya to fix it?

Tayipc38:
> This is an awful solution to a non-issue. I don't think we should merge this, not only does it add a worthless feature, but this could also break things like fm_mission_controller which relies on the camera state. You're patching like 5 things and hooking like 20 natives, if any of those break from updates we have to fix that, all so what, the camera doesn't go swoop swoop swoop?
> 
> I vote hard no. 

Its a little more than the swoop swoop. I had adopted this and worked on it a fair while. Its rather cool how the transition works, you can see the population of the player pool around you and keep playing. I had to add a ton more hooks and patches so you could remain in a car. Only problem was the interior issue (spawn location setting) that was so persistenly annoying. From animations to tasks I had successfully prevented all instances where the player ped was interrupted, so you could litterally run between sessions. When the hooks are done right and made very case sensitive, I wouldn't worry too much about it breaking other stuff.

NjuuOwO:
> This is an awful solution to a non-issue. I don't think we should merge this, not only does it add a worthless feature, but this could also break things like fm_mission_controller which relies on the camera.

I did encounter it breaking stuff, after all it was meant to be a quick adaptation so it still needs to be corrected. IMO it shouldn't be called "a worthless feature"

> Would you be willing to maintain it in the future or would you have to rely on tupoy-ya to fix it?

I would maintain it but I will leave it as draft PR for now, I will work on it if I will find some time
Will also take any suggestions to consider

PR #1955, opened 2023-08-11
fix(hotkey) check for is chatbox is active

### LINKED ISSUE #1948: [Bug]: Hotkeys while in textfields
### Describe the bug

When i'm currently in a textfield (socialclub message, sending text message via phone, other popups where you need to enter text), keybinds are still working. This is very FRUSTRATING and LIFE CHANGEING when something is bound to key like a-z.

### Steps To Reproduce

Bind e.g. noclip to n.
Send someone a message (like don't send just click send so gui opens)
Type n in gui
NoClip should enable

### Expected Behavior

NoClip should not enable :D

### Logs and/or screenshots

Nah

### Store

Epic Games

### Additional context

I think that the text gui should be easy. With socialclub idk

### Build

### Coload

alyxme:
Add check to disable hotkeys if chat typing is active
Closes #1948

### DIFF
--- a/src/services/hotkey/hotkey.cpp
+++ b/src/services/hotkey/hotkey.cpp
@@ -16,9 +16,10 @@ namespace big
 		return !m_cooldown.has_value() || std::chrono::high_resolution_clock::now() >= m_wakeup;
 	}
 
+	static bool IS_MP_TEXT_CHAT_TYPING;
 	void hotkey::exec()
 	{
-		if (m_cooldown.has_value())
+		if (!IS_MP_TEXT_CHAT_TYPING && m_cooldown.has_value())
 			m_wakeup = std::chrono::high_resolution_clock::now() + m_cooldown.value();
 
 		command::get(m_command_hash)->call({});


Yimura (changes requested):
I’m pretty sure this code doesn’t work, I didn’t test it though.

Did you test it?

alyxme:
> I’m pretty sure this code doesn’t work, I didn’t test it though.
> 
> Did you test it?

yes tested right now. works fine!

xiaoxiao921:
Can you test again?

alyxme:
Id not pull that without all tests pass.

xiaoxiao921:
A similar check would make more sense than whatever non working code you just wrote https://github.com/YimMenu/YimMenu/blob/e87931bf605cf967ac5b9d490bfa0dcb3c1bb168/src/backend/looped/weapons/weapon_hotkeys.cpp#L27

alyxme:
> A similar check would make more sense than whatever non working code you just wrote
> 
> https://github.com/YimMenu/YimMenu/blob/e87931bf605cf967ac5b9d490bfa0dcb3c1bb168/src/backend/looped/weapons/weapon_hotkeys.cpp#L27

Yes you may right. Thanks for the review! I was not so into to make this. The behavior of the issue is work because of that code you mentioned!

PR #1962, opened 2023-08-12
(feat) option Fix to context menu

### LINKED ISSUE #1954: [Request]: Remote Repair Vehicle
### Problem

I can destory a vehicle remotely but i also wantto repair it :D

### Solution

Add <player> -> Vehicle -> Repair Vehicle

### Reason

See Problem

### Additional context

_No response_

alyxme:
Add option to Fix vehicle via context menu
Closes #1954


src/services/context_menu/context_menu_service.hpp
```diff
@@ -72,6 +72,18 @@ namespace big
...
+		           [this] {
+			            if (entity::take_control_of(m_handle))
+		        	    {
+			        	    VEHICLE::SET_VEHICLE_ENGINE_HEALTH(m_handle, 1000.f);
+				            VEHICLE::SET_VEHICLE_FIXED(m_handle);
+				            VEHICLE::SET_VEHICLE_DEFORMATION_FIXED(m_handle);
+				            VEHICLE::SET_VEHICLE_DIRT_LEVEL(m_handle, 0.f);
+			            }
+			            else
+				            g_notification_service->push_warning("Toxic", "Failed to take control of vehicle.");
```
Tayipc38: I don't think this falls under toxic does it?

alyxme: Could it be just "Warning" instead of "Toxic" notice?
Also Id rearrange list to remove "explode" from atop to not accidentaly blowup something like it happen so often.

TheGreenBandit: Yes, Warning would be the right fit for it. What would you suggest be at the top? iirc it saves the last option you were on

alyxme: Yes it saves, but for current game session only. So I did rearrange. Explode entity is strange behavior tbh. I moved it to bottom of list. Also I replace identity with outfit because it is more useful than frustrating if click on it and can't be revert until rejoin.

pierrelasse:
While we are at it.. Remote Vehicle godmode toggle? Is it possible? Cuz enable vgod in vehicle and then exeting and someone elde driving, the vehicle still has gos.

Btw thx for making pull request.

Yimura (approved):

PR #1969, opened 2023-08-14
(feat) Horn boost cam

oPiloto:
a horn boost guided by the camera

I will create a pr in the translations


gir489returns:
This could be a Lua script instead.

gir489returns:
@Yimura can we close this?

Yimura (changes requested):
I'm not going to merge this PR in its current state, horn boost exists and you're using an extra check in the UI to disable the horn boost just in case the other one is enabled.

Could you instead integrate your new feature into the existing horn boost. And offer a radio switch or drop down to change between normal horn boost and your cam boost implementation?

You could make use of an enum and make it JSON serializable.

PR #1972, opened 2023-08-14
Wave strength controller

NjuuOwO:


src/views/world/view_world.cpp
```diff
@@ -137,6 +137,17 @@ namespace big
...
 				deleting = false;
 			});
+
+			ImGui::SeparatorText("Miscellaneous");
+
+			int override_waves_as_int = static_cast<int>(g.world.override_waves);
+			if (ImGui::SliderInt("Waves strength", &override_waves_as_int, 0, 5))
+			{
+				g.world.override_waves = static_cast<float>(override_waves_as_int);
+				MISC::WATER_OVERRIDE_SET_STRENGTH(g.world.override_waves);
```
Yimura, line 147: Executing a native in the UI thread, this should be done in a fiber.

Yimura (changes requested):
Also this could be made into a Lua script easily.

src/views/world/view_world.cpp
NjuuOwO: I don't know what you mean

gir489returns: > I don't know what you mean

Which is exactly why this PR is gonna be rejected and you need to make this a Lua instead.

NjuuOwO:
ye he's right

PR #1981, opened 2023-08-16
(feat) Speed Limit Override

### LINKED ISSUE #1957: [Request]: Bypass Max Speed
### Problem

_No response_

### Solution

Bypass vehicle speed caps

### Reason

Make cars go vr0000m more

### Additional context

_No response_

T14D3:
Closes #1957

### DIFF
--- a/src/gui/handling/current_profile/current_profile_other.cpp
+++ b/src/gui/handling/current_profile/current_profile_other.cpp
@@ -1,9 +1,13 @@
 #include "current_profile_tabs.hpp"
+#include "natives.hpp"
 
 namespace big
 {
 	void tab_current_profile::tab_other()
 	{
+		static bool overrideSpeedLimit = false;
+		static float speedLimitKMH     = 0.0f;
+
 		if (ImGui::BeginTabItem("HANDLING_TAB_OTHER"_T.data()))
 		{
 			ImGui::Text("HANDLING_ACCELERATION_MULTIPLIER"_T.data());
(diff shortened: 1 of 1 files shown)


src/gui/handling/current_profile/current_profile_other.cpp
```diff
@@ -23,8 +27,12 @@ namespace big
 				g_local_player->m_vehicle->m_handling_data->m_inertia_mult.y = fInertiaMult[1];
 				g_local_player->m_vehicle->m_handling_data->m_inertia_mult.z = fInertiaMult[2];
 			}
+			ImGui::Text("Speed Limit Override");
+			ImGui::SliderFloat("km/h", &speedLimitKMH, 0.f, 1000.f);
```
Yimura, line 31: People who use imperial won't understand this properly, also this feature should probably be added under the vehicle tab and not under handling.

Combined with that you don't save the speed override to the handling JSON file.
And you're not following coding guidelines with the capitalisation of your variables.

Yimura (changes requested):

src/gui/handling/current_profile/current_profile_other.cpp
T14D3: Right, should probably move the thing to the vehicle tab, will do
For the unit thing, I'll look into just pulling from whatever the user has set for the speedometer (km/h vs mp/h) and use that (and fix some capitalisation I guess)

binded-zz: does the speed work for flying stuff as well

ScottiePimp: > 

Simple fix for that. Add a bool check for which speed measurement the user has.

Then do:
```
float Speed = speedLimitInput / (imperial ? 2.236936f : 3.6f); // Convert to MPS
SET_ENTITY_MAX_SPEED(your_vehicle, Speed);
```

Yimura:
PR has gone stale.

PR #1996, opened 2023-08-19
(fix) max_vehicle/spawn_maxed still apply HSW mods

NjuuOwO:
lmk if it was left on purpose


maybegreat48:
This does the exact opposite of what you want to do

PR #2042, opened 2023-08-27
Incorporate folders into the Persist Vehicles feature. 

### LINKED ISSUE #2038: [Request]: Incorporate folders into the Persist Vehicles feature.
### Problem

The box from BBv2 wasn't meant to hold more than 5 vehicles, but I have to imagine people are using it for more than spawning vehicles in heists that they weren't supposed to. That was its original use case. The small box makes more than 5 vehicles a nightmare to find quickly, especially if the list constantly influx.

### Solution

Incorporate a drop-down box that allows the user to select a subcategory of vehicles they can also choose to save the vehicle to as well if it is selected at the time of persistence. The box should also auto size up to at least 30 vehicles, as there is no reason to save screen real-estate since it is the only submenu on screen at the time.

### Reason

Organizational purposes. 

### Additional context

Should also maintain regression with the previous use case and allow for JSONs to be emplaced/read from the root directory.

gir489returns:
(Closes #2038)

### DIFF
--- a/src/views/world/view_squad_spawner.cpp
+++ b/src/views/world/view_squad_spawner.cpp
@@ -315,7 +315,7 @@ namespace big
 			{
 				if (ImGui::Selectable("None", new_template.m_persistent_vehicle == "None"))
 					new_template.m_persistent_vehicle = "None";
-				for (auto& p : persist_car_service::list_files())
+				for (auto& p : persist_car_service::list_files(""))
 				{
 					if (ImGui::Selectable(p.data(), p == new_template.m_persistent_vehicle))
 						new_template.m_persistent_vehicle = p;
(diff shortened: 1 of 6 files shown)


src/views/world/view_squad_spawner.cpp
```diff
@@ -315,7 +315,7 @@ namespace big
 			{
 				if (ImGui::Selectable("None", new_template.m_persistent_vehicle == "None"))
 					new_template.m_persistent_vehicle = "None";
-				for (auto& p : persist_car_service::list_files())
+				for (auto& p : persist_car_service::list_files(""))
```
Yimura: Still gets called with "" when the default arg is "".

Yimura (changes requested):

Yimura (approved):
No further complaints

Issue #2022, opened 2023-08-25
[Improvement]:  move camera to high position when detecting crashes and kicks


lonelybud:
### Which feature

I was a host in a session.
Somebody started to ragdoll, camera twist, crash, gta banner, kick me etc all at once
I lagged so bad and crashed.

I believe maybe swathing to free cam mode  I could have been able to click the host kick button on that's guy inf page.

### Reason

Moving camera up at least give us time to observe and take decision

### Additional context

Using this build https://github.com/YimMenu/YimMenu/commit/ee364cd684e85b6287b4034fed5d189fc29dbe94

Yimura:
You need to always consider the weight of the offence against the reaction you'll put up against it.

GTA banner is a very light offence an can also be caused by a legit player, so I would only use kick reactions on offences like them trying to kick, crash,... you.

In regards to the crash, any clue on what happened or anything you saw? Any clue on the specific crash? We'd rather fix the crash at the root than to use some makeshift fix.

lonelybud:
my game seems to got stuck as  frame rates were very low (but i held for about 1 min in the lobby and then got desync kicked into a public lobby of my own). After that my game crashed. 
Restarted the game and then i went back into same lobby by spoofing host token and kicked that player. 

This was the guy who kicked me - `Anupam1111`.
I am using `relwithdebinfo` local build at https://github.com/YimMenu/YimMenu/commit/ee364cd684e85b6287b4034fed5d189fc29dbe94

This was the log (123MB) when i was kicked out of that lobby when i was host there -
[08-24-2023-23-16-31_cout.zip](https://github.com/YimMenu/YimMenu/files/12601301/08-24-2023-23-16-31_cout.zip)

no need to signup dropbox. just cancel the popup and click download.

xiaoxiao921:
When you upload big logs, always zip them, their usually see a 95% size decrease

0xEmiil:
Is this the cam you mean?
![image](https://github.com/YimMenu/YimMenu/assets/79391187/a8b5f825-d4ad-45ca-b047-b0285b64ecbf)

If it is i dont see any reason to use it on protection, the only way to use it is on crashes when you spawn a slod ped, maybe other thing lagged you like PTFX some kind of pool crash you but afaik that can be patched  without using the Sky Cam

lonelybud:
> Is this the cam you mean?

I meant the cam set to high altitude at random location in sky. I did not understood the picture you attached.

[...]

PR #2046, opened 2023-08-28
feat(reactions): toggle to set free camera high

lonelybud:
benefits #2022

<details>
 <summary>just temporary example showing how it will be triggered based on reaction-</summary>

  https://github.com/YimMenu/YimMenu/assets/139460769/885b6337-726d-4cb5-952e-9b188bc8bd5b
</details>

This feature will actually add a toggle alongside the existing toggles (notify, kick, log etc) in reactions UI under settings. So that one can now enable toggle for reactions like crash, end session kick, desync kick etc.

<details>
  <summary>Image</summary>

   image
</details>

### DIFF
--- a/src/views/settings/view_reaction_settings.cpp
+++ b/src/views/settings/view_reaction_settings.cpp
@@ -25,6 +25,7 @@ namespace big
 				ImGui::Checkbox("REACTION_BLOCK_JOINS"_T.data(), &reaction.block_joins);
 			ImGui::Checkbox("REACTION_KICK_PLAYER"_T.data(), &reaction.kick);
 			ImGui::Checkbox("TIMEOUT"_T.data(), &reaction.timeout);
+			ImGui::Checkbox("REACTION_FREE_CAMERA_HIGH"_T.data(), &reaction.timeout);
 			ImGui::TreePop();
 		}
 		ImGui::PopID();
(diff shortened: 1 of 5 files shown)


src/views/settings/view_reaction_settings.cpp
```diff
@@ -25,6 +25,7 @@ namespace big
 				ImGui::Checkbox("REACTION_BLOCK_JOINS"_T.data(), &reaction.block_joins);
 			ImGui::Checkbox("REACTION_KICK_PLAYER"_T.data(), &reaction.kick);
 			ImGui::Checkbox("TIMEOUT"_T.data(), &reaction.timeout);
+			ImGui::Checkbox("REACTION_FREE_CAMERA_HIGH"_T.data(), &reaction.timeout);
```
lonelybud: need to add translation here. I don't know where to add

lonelybud:
should we continue to use default ease time (500ms) or set it to zero (so that camera move to top instantly) when toggle enabled?
https://github.com/YimMenu/YimMenu/blob/9d8d71adf6a7db0ac916b35cdd8480a2bc424443/src/backend/looped/self/free_cam.cpp#L42

gir489returns:
Instead of fixing the underlying of the overloading crash, we just simply turn the other cheek and continue playing like nothing's wrong? I vote hard no on this PR.

0xEmiil:
I see this PR useless also the high cam can be done in a easy and different way than this

Yimura:
It's not even that, it's just a way of ignoring the underlying issue (the actual crash).

I'll be closing this in favour of fixing the actual crash.

PR #2063, opened 2023-08-30
Redesigned Vehicle Spawn menus to have consistent features.

### LINKED ISSUE #2048: [Improvement]: Consistency between Vehicle spawn tabs
### Which feature

The New and Personal Vehicle tabs have an option to preview the vehicle and determine if you want to be teleported into the vehicle, as well as New having Spawn Max and Personal having Spawn Cloned, but those are specific to those features. With the recent introduction of #2042, I feel each vehicle box should conform to the same standard. No more than 30 items per box, and we at least show Preview and Spawn In checkboxes for both the XML and the Persist tab. The Personal vehicle tab also shows vehicles that should not be spawned as personal vehicles, such as the Avenger, acid lab bike, Terrorbyte, and others like the RC Bandito and the Invade and Persuade tank. We should only show them if Spawn Clone is checked, as we could potentially get the user flagged by Rockstar if they assign their personal vehicle slot to a vehicle that can't normally be a personal vehicle. 

### Reason

Consistency.

### Additional context

_No response_

gir489returns:
Closes #2048


gir489returns:
I chose to ignore XML Vehicles as they are really only there to superficially import the vehicle into GTA, and then save them as a Persist Vehicle (JSON) which then can be organized much better.

gir489returns:
This also fixes a bug with Personal Vehicle preview/clone not showing clan logos if the player had them.

https://github.com/gir489returns/YimMenu/blob/makeVehicleSpawnMenuConsistent/src/util/vehicle.cpp#L358

Yimura (approved):

PR #2073, opened 2023-09-02
fix(renderer): Fix renderer callbacks called for nothing.

xiaoxiao921:


Yimura (approved):
Nice thanks for fixing.

PR #2085, opened 2023-09-03
(fix) Seatbelt values are set to opposite

### LINKED ISSUE #2082: [Bug]: the seat belt does not work
### Describe the bug

The seat belt doesn't work, the motorcycle hits an obstacle, the character still falls off the car, is my method of operation wrong? Or do I need to trigger a pre-condition to use my seat belt? My version is [c459ccc]

### Steps To Reproduce

the seat belt does not work, the motorcycle crashes into the obstacle, the character will still fall from the vehicle

### Expected Behavior

After starting this feature, the character should not fall off the car

### Logs and/or screenshots

```
<paste your logs here>
```

### Store

Steam

### Additional context

_No response_

### Build

### Coload

NjuuOwO:
Fix #2082

### DIFF
--- a/src/backend/looped/vehicle/seatbelt.cpp
+++ b/src/backend/looped/vehicle/seatbelt.cpp
@@ -1,4 +1,5 @@
 #include "backend/looped_command.hpp"
+#include "gta/enums.hpp"
 #include "natives.hpp"
 
 namespace big
@@ -10,13 +11,13 @@ namespace big
 		virtual void on_tick() override
 		{
 			PED::SET_PED_CONFIG_FLAG(self::ped, 32, false);
-			PED::SET_PED_CAN_BE_KNOCKED_OFF_VEHICLE(self::ped, eKnockOffVehicle::KNOCKOFFVEHICLE_NEVER);
+			PED::SET_PED_CAN_BE_KNOCKED_OFF_VEHICLE(self::ped, (int)eKnockOffVehicle::KNOCKOFFVEHICLE_NEVER);
 		}
 
 		virtual void on_disable() override
 		{
 			PED::SET_PED_CONFIG_FLAG(self::ped, 32, true);
-			PED::SET_PED_CAN_BE_KNOCKED_OFF_VEHICLE(self::ped, eKnockOffVehicle::KNOCKOFFVEHICLE_DEFAULT);
+			PED::SET_PED_CAN_BE_KNOCKED_OFF_VEHICLE(self::ped, (int)eKnockOffVehicle::KNOCKOFFVEHICLE_DEFAULT);
 		}
 	};
 
--- a/src/backend/looped/vehicle/seatbelt.cpp
+++ b/src/backend/looped/vehicle/seatbelt.cpp
@@ -10,13 +10,13 @@ namespace big
 		virtual void on_tick() override
 		{
 			PED::SET_PED_CONFIG_FLAG(self::ped, 32, false);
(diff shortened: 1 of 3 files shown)


src/backend/looped/vehicle/seatbelt.cpp
```diff
@@ -10,13 +10,13 @@ namespace big
...
 			PED::SET_PED_CONFIG_FLAG(self::ped, 32, false);
-			PED::SET_PED_CAN_BE_KNOCKED_OFF_VEHICLE(self::ped, false);
+			PED::SET_PED_CAN_BE_KNOCKED_OFF_VEHICLE(self::ped, 1);
 		}
 
 		virtual void on_disable() override
 		{
 			PED::SET_PED_CONFIG_FLAG(self::ped, 32, true);
-			PED::SET_PED_CAN_BE_KNOCKED_OFF_VEHICLE(self::ped, true);
+			PED::SET_PED_CAN_BE_KNOCKED_OFF_VEHICLE(self::ped, 0);
```
Yimura: They are boolean values, use `true/false` instead of `1/0`.

Yimura (changes requested):

src/backend/looped/vehicle/seatbelt.cpp
NjuuOwO: > They are boolean values, use `true/false` instead of `1/0`.

`SET_​PED_​CAN_​BE_​KNOCKED_​OFF_​VEHICLE(Ped ped, int state)`
It's an integer and these are the 4 states:

```
enum eKnockOffVehicle
{
	KNOCKOFFVEHICLE_DEFAULT,
	KNOCKOFFVEHICLE_NEVER,
	KNOCKOFFVEHICLE_EASY,
	KNOCKOFFVEHICLE_HARD
};
```

xiaoxiao921: Then define the enum and use it? instead of hardcoding a 0

src/core/enums.hpp
```diff
@@ -143,6 +143,14 @@ namespace big
 		UNK2                 = 1 << 16
 	};
 
+	enum eKnockOffVehicle
```
gir489returns, line 146: You have to strongly type the enum to the parameter instance type or class keyword, or else you are just creating an enum type.

gir489returns (changes requested):

src/core/enums.hpp
NjuuOwO: Oh, thanks for explaining.
And shouldn't this enum be in `/gta/enums.hpp`?

gir489returns: If the enumeration comes from GTA itself it should.

Yimura (changes requested):
Waiting for the changes from @gir489returns to be implemented then it'll be ready to be merged.

NjuuOwO:
I guess it should be done like that

gir489returns:
You also didn't move it to the GTA enums class like you said you were going to.

src/core/enums.hpp
```diff
@@ -143,6 +143,14 @@ namespace big
...
 	};
 
+	enum class eKnockOffVehicle
+	{
+		KNOCKOFFVEHICLE_DEFAULT,
+		KNOCKOFFVEHICLE_NEVER,
+		KNOCKOFFVEHICLE_EASY,
+		KNOCKOFFVEHICLE_HARD
+	};
+
```
Yimura: enum hasn't been moved.

Yimura (changes requested):

Yimura (approved):

Issue #2093, opened 2023-09-06
[Bug]: Freecam prevents your car from moving


shadow7483147:
### Describe the bug

When returning from freecam mode, your car is no longer able to be moved, but it is able to rev the engine. Leaving/reentering the car does not fix the issue. Using a different car (such as respawning your PV) does.

### Steps To Reproduce

1. enter a car
2. enter then exit freecam
3. try to move the car

### Expected Behavior

Returning away from freecam should allow you to control your car.

### Logs and/or screenshots

```
[05:29:04.4471945][INFO/main.cpp:65] Yim's Menu Initializing
[05:29:04.4472091][INFO/main.cpp:66] Git Info
	Branch:	master
	Hash:	1594ee1f7680731dd20c0aa194bc61137b9ebf72
	Date:	Sat Sep 2 20:58:13 2023
[05:29:04.4472150][DEBUG/thread_pool.cpp:21] Resizing thread pool from 0 to 4
[05:29:04.4473087][INFO/main.cpp:69] Thread pool initialized.
[05:29:04.4483990][INFO/main.cpp:72] Settings Loaded.
[05:29:04.4484059][DEBUG/thread_pool.cpp:95] Thread 1384 executing settings.cpp:18
[05:29:04.4485082][INFO/pointers.hpp:94] GTA5 pointers cache is up to date, using it.
[05:29:04.4485115][INFO/pointers.cpp:1697] Loading pointers instance from cache
[05:29:04.4485123][INFO/pointers.cpp:1706] Pointers cache: Loading 202 fields from the cache
[05:29:04.4485164][DEBUG/module.cpp:50] Waiting for socialclub.dll...
[05:29:04.4485470][INFO/pointers.hpp:94] Social Club pointers cache is up to date, using it.
[05:29:04.4485480][INFO/pointers.cpp:1697] Loading pointers instance from cache
[05:29:04.4485485][INFO/pointers.cpp:1706] Pointers cache: Loading 5 fields from the cache
[05:29:04.4485868][INFO/main.cpp:75] Pointers initialized.
[05:29:04.4487088][INFO/main.cpp:78] Byte Patch Manager initialized.
[05:29:04.7401496][INFO/main.cpp:81] Renderer initialized.
[05:29:04.8913355][INFO/main.cpp:85] Fiber pool initialized.
[05:29:04.9906316][INFO/main.cpp:88] Translation Service initialized.
[...]

PR #2101, opened 2023-09-08
Fix(freecam) unfreeze vehicle after freecam

alyxme:
Described in #2093


Yimura (approved):
Welp that was a stupid mistake in the code.

Thanks for the fix.

PR #2115, opened 2023-09-12
enforce utf8

xiaoxiao921:
std format currently throw if the string literal encoding of the user that compiles the project is not utf8 compatible so might aswell enforce it.

most contributors already are using a utf8 compatible string literal encoding anyway, for checking you can use this piece of code

https://www.fileformat.info/info/unicode/char/1234/index.htm
<html>
<body>

UTF-8 (hex) | 0xE1 0x88 0xB4 (e188b4)
-- | --

</body>
</html>

```cpp
constexpr bool is_utf8()
{
	constexpr char const* s = "\u1234";

	return (unsigned char)s[0] == 0xE1 && (unsigned char)s[1] == 0x88 && (unsigned char)s[2] == 0xB4 && (unsigned char)s[3] == 0x00;
}

static constexpr bool hover_me = is_utf8();
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,8 @@ file(GLOB_RECURSE SRC_MAIN
 
 if (MSVC)
   add_compile_options(/bigobj)
+  add_compile_options("$<$<C_COMPILER_ID:MSVC>:/utf-8>")
+  add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
 else ()
   add_compile_options(-Wa,-mbig-obj)
 endif ()


gir489returns:
Can't we add the is_utf8 check into a #IFDEF check as part of the build? Since you made it a constexpr, it should be evalulated at compile time, although #IFDEF is apart of the precompile phase, I'm sure there's something we can do if is_utf8 fails, like just call a function that doesn't exist and constantly make the build fail. Something like you_dont_have_utf_8_on() which is like a message in a bottle style error.

xiaoxiao921:
we can put it into a file and wrap it around a `static_assert` call, which will not allow compilation unless your string literals are utf8 compatbile

Yimura (approved):

PR #2040, opened 2023-08-27
(fix) Passive not working correctly

NjuuOwO:
I'm taking back, + 1 global was fucking things up
Fixes #2031

Issue #2010, opened 2023-08-21
[Bug]: No idle kick not working


ghost:
### Describe the bug

I get kicked out of the lobby even if this feature is active

### Steps To Reproduce

Go to Self tab

Activate no idle kick

you will be kicked out when you stand idle

### Expected Behavior

You shouldn't be thrown when you're idle

### Logs and/or screenshots

[yimidlebug.txt](https://github.com/YimMenu/YimMenu/files/12401450/yimidlebug.txt)

### Store

Epic Games

### Build

- [X] I've confirmed the issue exists on the latest version of YimMenu

### Coload

- [X] I've tested this without co-loading anything.

ScottiePimp:
I see a lot of people complaining about this. So just use this:

*Global(1654054).At(1155).As<int*>() = -1;
*Global(1654054).At(1155 + 16).As<int*>() = -1;

Want to disable the idle cam aswell? Call these natives in a loop:

CAM::INVALIDATE_IDLE_CAM();
CAM::INVALIDATE_CINEMATIC_VEHICLE_IDLE_MODE();

ghost:
> I see a lot of people complaining about this. So just use this:
> 
> _Global(1654054).At(1155).As<int_>() = -1; _Global(1654054).At(1155 + 16).As<int_>() = -1;
> 
> Want to disable the idle cam aswell? Call these natives in a loop:
> 
> CAM::INVALIDATE_IDLE_CAM(); CAM::INVALIDATE_CINEMATIC_VEHICLE_IDLE_MODE();


I don't have a lot of technical knowledge about the game.
where will I enter these numbers in the stat editor, there are only 2 places below and 3. the part just shows how much the value is, I guess.

ScottiePimp:
In the code. Same place you build the dll

gir489returns:
> In the code. Same place you build the dll

I'm making a build off it, I'll post the PR number in a moment.

ScottiePimp:
> > In the code. Same place you build the dll
> 
> 
> 
> I'm making a build off it, I'll post the PR number in a moment.

Great!

gir489returns:
PR #2011 is submitted, when the build bot is done please test it.

ScottiePimp:
> PR #2011 is submitted, when the build bot is done please test it.

Remove the for int loop. Also, set the globals to -1. I grabbed these a long time ago but I have a faint memory that they only worked with -1, could have changed throughout the updates but yea.

[...]

PR #2135, opened 2023-09-15
Various fixes.

### LINKED ISSUE #2048: [Improvement]: Consistency between Vehicle spawn tabs

gir489returns:
Closes #2048 missed issue.
Reverted #2010's Anti-AFK code.
Fixed #2040 not using the tunables service when applicable.

### DIFF
--- a/src/backend/looped/self/toggle_passive.cpp
+++ b/src/backend/looped/self/toggle_passive.cpp
@@ -4,6 +4,7 @@
 #include "script_global.hpp"
 #include "script/globals/GPBD_FM_3.hpp"
 #include "gta_util.hpp"
+#include "services/tunables/tunables_service.hpp"
 
 namespace big
 {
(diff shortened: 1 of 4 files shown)


src/views/vehicle/spawn/view_pv.cpp
```diff
@@ -8,6 +8,32 @@
...
 namespace big
 {
+	bool is_blacklist_vehicle(Hash vehicle_hash)
+	{
+		switch (vehicle_hash)
+		{
+			case RAGE_JOAAT("avenger"):
+			case RAGE_JOAAT("avenger3"):
+		    case RAGE_JOAAT("hauler2"):
+		    case RAGE_JOAAT("phantom3"):
```
Yimura: Mixed indentation

src/backend/looped/self/toggle_passive.cpp
```diff
@@ -21,15 +22,15 @@ namespace big
...
+			*scr_globals::passive.as<PBOOL>()                 = TRUE;
 		}
 
 		virtual void on_disable() override
 		{
-			*script_global(1574582).as<bool*>() = false;
+			*scr_globals::passive.as<PBOOL>() = FALSE;
 			NETWORK::SET_LOCAL_PLAYER_AS_GHOST(false, false);
-			*scr_globals::tuneables.at(27371).as<int*>() = 30000;
+			*g_tunables_service->get_tunable<int*>(-29732167) = 30000;
```
Yimura, line 33: What is this `-29732167`? Any way on finding it again if it changes?

Yimura (changes requested):

src/backend/looped/self/toggle_passive.cpp
gir489returns: It’s just the JOAAT’s integer representation. It’s what was in the tunables file. I tried looking for JOAATs that matched it, but all I could find were collisions. The tunable service will auto update it with the matching JOAAT value.

src/views/vehicle/spawn/view_pv.cpp
gir489returns: Nit picking.

Yimura (approved):

PR #2140, opened 2023-09-17
Disable input when typing into a text field.

### LINKED ISSUE #2139: [Improvement]: When typing into a field, the inputs are duplicated to GTA, causing erroneous results
### Which feature

When requesting user input, YimMenu currently allows the input to be passed into the game, thus causing undefined behavior. 

### Reason

This is not an expected result of asking for input.

### Additional context

_No response_

gir489returns:
Closes #2139

### DIFF
--- a/src/backend/backend.cpp
+++ b/src/backend/backend.cpp
@@ -117,6 +117,7 @@ namespace big
 		while (g_running)
 		{
 			looped::hud_transition_state();
+			looped::hud_disable_input();
 			looped::session_pop_multiplier_areas();
 			looped::session_force_thunder();
 			looped::session_randomize_ceo_colors();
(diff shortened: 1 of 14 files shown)


gir489returns:
Certain scripts call GET_DISABLED_CONTROL_UNBOUND_NORMAL or IS_DISABLED_CONTROL_PRESSED or IS_DISABLED_CONTROL_JUST_PRESSED or GET_DISABLED_CONTROL_NORMAL and this will not prevent those calls from succeeding, since there's no way to overwrite the disabled check. We'd either have to hook all 4 native functions and check the typing variable from within it, or we could look into somehow disconnecting the keyboard input entirely to GTA and have it exclusively run through ImGui somehow.

alyxme:
While we at here is it possible to somehow fix https://github.com/YimMenu/YimMenu/issues/1948 with this? (Possibly about fields for entering text in the SC or ingame SMS)

Yimura (approved):

PR #2142, opened 2023-09-17
Small refactor: main start/cleanup threads logic

xiaoxiao921:


src/services/player_database/player_database_service.cpp
```diff
@@ -270,22 +273,26 @@ namespace big
 
 	void player_database_service::start_update_loop()
 	{
-		if (!g.player_db.update_player_online_states)
-			return;
-
 		g_thread_pool->push([this] {
+			while (!big::g_running)
```
Yimura: Using `big::` while several lines below you didn't use it.

Yimura (changes requested):

Yimura (approved):

PR #2146, opened 2023-09-18
refactor: Implement HTTP client with proxy support

Yimura:
Ported over from one of my private projects.
Should allow for better access to remote resources for people that are behind a firewall that blocks access to certain sites.

## Todo
- [ ] ~~Create structs for data in the API service~~
- [x] Create separate settings for a proxy client (don't want this to be in your main `settings.json`)
- [x] Add UI to edit proxy settings
- [ ] Detect system settings (maybe because Windows is cursed I might not)
- [x] Move text to translation


sch-lda:
It would be even better if there could be a 'Use System Proxy' button.

gir489returns:
> It would be even better if there could be a 'Use System Proxy' button.

What's the point? If you've setup a proxy at the networking adapter level, then all traffic will be funneled through the proxy anyway.

sch-lda:
> > It would be even better if there could be a 'Use System Proxy' button.
> 
> What's the point? If you've setup a proxy at the networking adapter level, then all traffic will be funneled through the proxy anyway.

V2ray and Clash are the most popular firewall-bypassing software. 
Clash's client typically offers two modes: System proxy and Tun Mode. The former modifies the Windows system proxy settings, but usually, most applications other than the browser do not consider these system proxy settings, so you still need to manually enter the proxy server address and port. The latter can indeed route all traffic, but this isn't ideal for online games like GTA Online, as due to certain public proxy service configurations, users may not be able to join others. Even if they can join, latency will increase.
V2ray's most famous client, V2rayN, does not provide a TUN mode to set up a proxy at the networking adapter level. 
The ideal scenario is to only route traffic through the proxy that needs to bypass the firewall, with GTA Online connecting directly while Yimmenu uses the proxy. I'm not sure if Yimmenu automatically considers the system proxy settings. If it doesn't, users would need to manually input the proxy server address and port information.

Yimura:
> What's the point? If you've setup a proxy at the networking adapter level, then all traffic will be funneled through the proxy anyway.

System Proxy != VPN, applications need to implement this into their code base to read the Windows proxy settings. These are separate settings that you can set in Windows without an external program.
image
There may also be other places in Windows to set up a proxy.

Do note that proxies can be implemented as a full tunnel on a network adapter but this will/may cause lag while playing GTA V as @sch-lda mentioned in his comment.

Yimura:
Some research on detecting Windows' settings:
 - https://stackoverflow.com/questions/6192563/detect-windows-ie-proxy-settings-changes/6195711
 - https://github.com/libproxy/libproxy/blob/main/src/backend/plugins/config-windows/config-windows.c#L183

Tayipc38:
Does this mean I can download json files automatically from some github pages?

Yimura:
@sch-lda, I've implemented basic proxy support. I haven't added support for detecting the system proxy yet, but if you could test the current setup and tell me if it works?

I'm mainly looking for answers to the following questions:
 - Does the proxy work? (the test setup should make a call to raw.githubusercontent.com)
 - Does the request timeout through the proxy? (if you're using a slow proxy for example)

gir489returns (changes requested):

sch-lda:
> @sch-lda, I've implemented basic proxy support. I haven't added support for detecting the system proxy yet, but if you could test the current setup and tell me if it works?
> 
> I'm mainly looking for answers to the following questions:
> 
>     * Does the proxy work? (the test setup should make a call to raw.githubusercontent.com)
> 
>     * Does the request timeout through the proxy? (if you're using a slow proxy for example)

1.Proxy using http protocol works fine
image

image
 However, socks4/5 test failed. I cannot find any connection form GTA5.exe in Clash-verge dashboard. I confirmed that the mixed port (7890) of Clash supports both HTTP and SOCKS5. I tested it using Telegram.
image
image


image

2.I tested with my own VPS and several public proxy services, and did not encounter any timeout issues. Yimmenu language can also be downloaded successfully.

sch-lda:
Socks4/5 works after removed these lines
image

image

EDIT:The SOCKS only worked for a while before failing again. Even after attempting to reinject Yimmenu and restarting Clash, it still couldn't be restored. It's weird. Http still works.

HTTP protocol is sufficient. Almost all proxy clients support HTTP to be compatible with Windows system proxy.

sch-lda:
Tested again with the latest code, all protocols are functioning properly.

Yimura (commented):
Will merge this now and implement detecting system proxy settings at a later date.

Issue #2151, opened 2023-09-19
[Request]: Unload Button for Lua Scripts


pierrelasse:
# Problem 1 (Bug)
When reloading a script that no longer exists = crash
Same for when the folder does not exist
## How2Fix?
Ensure the folder is created before reloading
Ensure the script file exists before reloading. Else unload the script

# Problem 2 (Request)
There is currently no way of unloading a script without unloading the menu
##Solution
Add unload buttons.

Idea for ui changes:
![image](https://github.com/YimMenu/YimMenu/assets/76223390/7e06485b-221f-417b-a7dd-18f9c94e3cdb)

gir489returns:
This should be upgraded to a bug, because if you remove the script you want to unload then click reload, an unexpected behavior of the game crashing occurs.

`[0]	KERNELBASE.dll RaiseException
[1]	VCRUNTIME140.dll CxxThrowException
[2]	C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.37.32822\include\filesystem L: 1813 std::filesystem::_Throw_fs_error
[3]	C:\Users\gir489\Desktop\YimMenu\src\lua\lua_module.cpp L: 89 big::lua_module::lua_module
[4]	C:\Users\gir489\Desktop\YimMenu\src\lua\lua_manager.cpp L: 106 big::lua_manager::load_module
[5]	C:\Users\gir489\Desktop\YimMenu\src\views\settings\view_lua_scripts.cpp L: 45 big::view::lua_scripts
[6]	C:\Users\gir489\Desktop\YimMenu\src\views\core\view_active_view.cpp L: 42 big::view::active_view
[7]	C:\Users\gir489\Desktop\YimMenu\src\views\core\view_root.cpp L: 13 big::view::root
[8]	C:\Users\gir489\Desktop\YimMenu\src\gui.cpp L: 140 big::gui::dx_on_tick
[9]	C:\Users\gir489\Desktop\YimMenu\src\renderer.cpp L: 142 big::renderer::on_present
[10]	C:\Users\gir489\Desktop\YimMenu\src\hooks\gui\swap_chain_present.cpp L: 14 big::hooks::swapchain_present
[11]	GTA5.exe+0x138334A 0x7FF70626334A
[12]	GTA5.exe+0x7E8549 0x7FF7056C8549
[13]	GTA5.exe+0x2D8E06 0x7FF7051B8E06
[...]

PR #2162, opened 2023-09-19
fix(lua): Fix crash when trying to reload a non-existent script. (#2151)

xiaoxiao921:


xiaoxiao921:
Pushed to the wrong branch, wooooo!

Yimura (approved):

PR #2171, opened 2023-09-21
Add Garage drop-down to Spawn PV menu

gir489returns:
This still requires reversing how the Terrorbyte works, as the Oppressor seems to be attached to it. I know how it works, I just didn't have enough time to finish, but I wanted to push this large mess of code so you can start looking at it for recommendations on compartmentalization.


gir489returns:
Edge case: when the user has spawn clone unchecked and they have a property that contains nothing but unspawnable vehicles, the list will display nothing. This is technically the correct scenario, as the garage contains vehicles, it just doesn't contain any personal vehicles that satisfy the conditions. We have to decide if user options should affect the garage calculation.

gir489returns:
Solution to edge case. Check if the model hash is banned and the spawn clone checkbox is unticked before emplacing garage.

Yimura (approved):

PR #2186, opened 2023-09-24
feat(anticheat): add bonus http blocker

yubie-re:
I'm not going to act like this is some revolutionary new code, however it is still quite effective.
This prevents certain anticheat reports going through to R*'s servers in the case that it does happen... can never be too safe.
First time I saw this technique in use was in Impulse, so credit goes to them I guess. Kept them UD for a few years.

### DIFF
--- a/src/hooks/misc/http_start_request.cpp
+++ b/src/hooks/misc/http_start_request.cpp
@@ -1,0 +1,18 @@
+#include "hooking.hpp"
+#include "logger/stack_trace.hpp"
+#include "pointers.hpp"
+#include "util/string_conversions.hpp"
+
+namespace big
+{
+    bool hooks::http_start_request(void* request, const char* uri)
+    {
+        if(strstr(uri, "Bonus"))
+        {
+            // This is for worst case scenario where a report does slip through the cracks...
+            // Lets make it go somewhere it doesn't matter -- don't let the reports reach their servers!
+            uri = "https://www.google.com";
+        }
+        return g_hooking->get_original<hooks::http_start_request>()(request, uri);
+    }
+}
--- a/src/gta_pointers.hpp
+++ b/src/gta_pointers.hpp
@@ -212,6 +212,8 @@ namespace big
 		PVOID m_queue_dependency;
 		PVOID m_interval_check_func;
 
+		PVOID m_http_start_request;
+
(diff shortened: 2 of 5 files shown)


gir489returns:
Isn't this function integrity checked?

gir489returns:
@maybegreat48

Yimura (approved):
Implementation seems fine, waiting a bit before merging though.

src/hooks/misc/http_start_request.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+namespace big
+{
+    bool hooks::http_start_request(void* request, const char* uri)
+    {
+        if(strstr(uri, "Bonus"))
+        {
+            // This is for worst case scenario where a report does slip through the cracks...
+            // Lets make it go somewhere it doesn't matter -- don't let the reports reach their servers!
+            uri = "https://www.google.com";
```
maybegreat48, line 14: Why not `http://0.0.0.0` or something?

maybegreat48:
> Isn't this function integrity checked?

It's a different function that isn't. The name is somewhat misleading since it just encodes the endpoint

src/hooks/misc/http_start_request.cpp
yubie-re: Would work, but wouldn't really change much. If we really wanted the "best option" then we'd make some file that contains a valid transaction response (we wouldn't have to worry about having the correct playerBalance etc, as it doesn't get applied from the Bonus endpoint), but it's not really necessary.

yubie-re:
> > Isn't this function integrity checked?
> 
> It's a different function that isn't. The name is somewhat misleading since it just encodes the endpoint

It should do a bit more than encoding the endpoint, it clears the request struct out and initializes it in general.

gir489returns (approved):

gir489returns:
> It should do a bit more than encoding the endpoint, it clears the request struct out and initializes it in general.

It's calling rage::netHttpRequest::ClearCachedEncodedUri on p0 then sets -1 and 0 on a bunch of entries in the class. There's a rage::datGrowBuffer* at 0x5C8 that it calls the constructor on, after calling Vfunc 9 on p0. Then it sets the time of the request on p0 + 0x5A4. I don't see it doing much more than that.

PR #2189, opened 2023-09-24
use multikick when lobby lock is enabled

lonelybud:


gir489returns:
What is the purpose of this?

xiaoxiao921:
> What is the purpose of this?

This fix an issue where the feature simply doesn't work because breakup kick only work while being the host

Yimura (approved):

PR #2194, opened 2023-09-25
feat(persist_car): auto fill save folder name

lonelybud:
auto fill save folder name to that of selected folder (except root)

<details>
 <summary> video </summary>

https://github.com/YimMenu/YimMenu/assets/139460769/0045aa08-0fe0-4b9d-ac06-8d626167cd42
<details>

### DIFF
--- a/src/views/vehicle/spawn/view_persist_car.cpp
+++ b/src/views/vehicle/spawn/view_persist_car.cpp
@@ -44,6 +44,7 @@ namespace big
 	void view::persist_car()
 	{
 		static std::string selected_vehicle_file;
+		static char save_folder[50]{};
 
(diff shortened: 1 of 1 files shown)


lonelybud:
it was a simple change though in order to choose the folder already created. Right now we have to manually type the name of same folder again to add a new vehicle. You are right about UX - "changing functionality of a component which is not intended".

I think you understand the original pain point of this pr. Btw we can propose a different change to solve this issue. Shall i raise an issue?

lonelybud:
Btw same thing exists for animation menu as well. Why not go that much deep and add this simple change?

edit: other thing we can add is a button or checkbox to prefill from selected item from dropdown list.

PR #2206, opened 2023-09-29
New Flying feature

Tayipc38:
The "Superman" feature was undeniably trash, I intend to replace it with an even worse feature. This feature makes use of a spawned Bandito attached to the player ped to simulate smooth flight (since gravity does not sync on player peds online). This does mean network control issues might ensue at times and make it a little buggy, but I personally don't really care to circumvent any of that.

This is an old private lil thing so hold your critique in your pants.


src/backend/looped/self/super_hero_fly.cpp
```diff
@@ -0,0 +1,404 @@
...
+		using looped_command::looped_command;
+
+		bool flying            = false;
+		bool landing           = false;
+		bool launching         = false;
+		bool anims_are_playing = false;
+
+		bool timer = false;
+		std::chrono::system_clock::time_point time_stamp;
+		std::chrono::duration<double> duration;
```
gir489returns, line 25: Since you are refering to this property later in milliseconds, you should type it as std::chrono::milliseconds, not double.

gir489returns (changes requested):

src/backend/looped/self/super_hero_fly.cpp
```diff
@@ -0,0 +1,404 @@
...
+
+		bool charging             = false;
+		double charge_intensity   = 1;
+		int charge_ptfx           = 0;
+		float explosion_intensity = 0;
+
+		Vehicle veh_handle{};
+		Vector3 land_coords{};
+		Vector3 fly_motion{};
```
Yimura: Prefix member variables with `m_`.

src/util/math.hpp
```diff
@@ -36,4 +37,30 @@ namespace big::math
...
+		    coord.y,
+		    coord.z,
+		    farCoords.x,
+		    farCoords.y,
+		    farCoords.z,
+		    -1,
+		    ignore,
+		    7);
+		SHAPETEST::GET_SHAPE_TEST_RESULT(ray, &hit, &endCoords, &surfaceNormal, &hit_entity);
+		return endCoords;
```
Yimura: snake_case coding style

src/views/self/view_self.cpp
```diff
@@ -79,7 +79,22 @@ namespace big
...
+
+			ImGui::Checkbox("Gradual speed", &g.self.super_hero_fly.gradual);
+			ImGui::Checkbox("Explosions", &g.self.super_hero_fly.explosions);
+			ImGui::Checkbox("Auto land", &g.self.super_hero_fly.auto_land);
+			ImGui::Checkbox("Charge launch", &g.self.super_hero_fly.charge);
+			ImGui::Checkbox("Charge ptfx", &g.self.super_hero_fly.ptfx);
+			ImGui::SetNextItemWidth(150);
+			ImGui::InputFloat("Speed", &g.self.super_hero_fly.fly_speed, 1, 50);
+			ImGui::SetNextItemWidth(150);
+			ImGui::InputFloat("Initial launch", &g.self.super_hero_fly.initial_launch, 1, 50);
```
Yimura: Make a PR on the translation repo to add this.

src/backend/looped/self/super_hero_fly.cpp
```diff
@@ -0,0 +1,404 @@
...
+			g_notification_service->push("Super hero fly", "Use the following keys to control the Super Hero Fly:\n'Q' 'SPRINT/DUCK 'WASD'.");
+		}
+
+		virtual void on_disable() override
+		{
+			reset();
+		}
+	};
+
+	super_hero_fly g_super_hero_fly("superherofly", "Super Hero Fly", "Fly like a super hero", g.self.super_hero_fly.enabled);
```
Yimura: Make a PR on the translation repo to add this.

Yimura (changes requested):

src/backend/looped/self/super_hero_fly.cpp
Tayipc38: the `duration` variable is closely related with `charge_intensity`, this on its own might be unusual but it does force me to use double for the duration.

gir489returns: Then the name of the variable is misleading. Either rename it, or type it as a std::chrono helper time duration.

Tayipc38: na

lonelybud:
btw just asking. Why do we have superman type stuff when `no clip` is a good option?

Tayipc38:
> btw just asking. Why do we have superman type stuff when `no clip` is a good option?

You haven't seen how fun this thing is.

Yimura (approved):
Should be ready to merge now, will handle the other PR's that still need their translations done.
Then after merging the translations this will go into master.

PR #2216, opened 2023-10-04
Bypass integrity checks

### LINKED ISSUE #2215: [Bug]: Update to Game Version: 3028
### Describe the bug

Rockstar Dropped another update

### Steps To Reproduce

Updating Game 😅

### Expected Behavior

...

### Logs and/or screenshots

[cout.log](https://github.com/YimMenu/YimMenu/files/12791375/cout.log)

### Store

Rockstar

### Additional context

The menu loads fine, prolly the Globals needs to be updated for some stuff to work without triggering anti-cheats

### Build

### Coload

maybegreat48:
Might be a good idea to wait before merging this

Fixes #2215

### DIFF
--- a/src/call_hook.cpp
+++ b/src/call_hook.cpp
@@ -1,6 +1,5 @@
 #include "common.hpp"
 #include "call_hook.hpp"
-#include "memory/handle.hpp"
 
 namespace
 {
@@ -13,25 +12,26 @@ namespace big
 {
 	call_hook_memory::call_hook_memory()
 	{ 
-		m_memory = (uint8_t*)VirtualAlloc((void*)((uintptr_t)GetModuleHandle(0) + 0x20000000), 1024, MEM_COMMIT | MEM_RESERVE, PAGE_EXECUTE_READWRITE);
+		m_memory = VirtualAlloc((void*)((uintptr_t)GetModuleHandle(0) + 0x20000000), 1024, MEM_COMMIT | MEM_RESERVE, PAGE_EXECUTE_READWRITE);
 		m_offset = 0;
 	}
 
 	call_hook_memory::~call_hook_memory()
 	{
-		VirtualFree(m_memory, 0, MEM_RELEASE);
+		VirtualFree(m_memory.as<void*>(), 0, MEM_RELEASE);
 	}
 
 	void* call_hook_memory::allocate_jump_sequence(void* func)
 	{
 		m_offset = m_offset + ((16 - (m_offset % 16)) % 16); // align
 
-		*(int16_t*)&m_memory[m_offset] = 0xB848;
-		*(uint64_t*)&m_memory[m_offset+2] = (uint64_t)func;
-		*(int16_t*)&m_memory[m_offset+10] = 0xE0FF;
+		*m_memory.add(m_offset).as<int16_t*>() = 0xB848;
+		*m_memory.add(m_offset).add(2).as<void**>() = func;
+		*m_memory.add(m_offset).add(10).as<int16_t*>() = 0xE0FF;
 
 		m_offset += 12;
-		return &m_memory[m_offset - 12];
+
+		return m_memory.add(m_offset).sub(12).as<void*>();
 	}
 
 	call_hook::call_hook(void* location, void* hook) :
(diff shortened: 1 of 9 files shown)


src/hooks/misc/remove_player_from_sender_list.cpp
```diff
@@ -0,0 +1,11 @@
+#include "hooking.hpp"
+#include "pointers.hpp"
+
+namespace big
+{
+	bool hooks::remove_player_from_sender_list(void* list, uint64_t rockstar_id)
+	{
+		g_pointers->m_gta.m_remove_player_from_sender_list(list, &rockstar_id);
```
gir489returns, line 8: Incorrectly typed parameter here. rockstar_id should be a uint64_t * since we type m_rockstar_id coming from the player info class as uint64_t, but it is clearly a pointer passed to it, for an unbeknownst reason.

src/call_hook.cpp
```diff
@@ -0,0 +1,62 @@
+#include "common.hpp"
+#include "call_hook.hpp"
+#include "memory/handle.hpp"
+
+namespace
+{
+	big::call_hook_memory g_call_hook_memory;
```
gir489returns, line 7: Request the compiler to inline this and remove it from the namespace, this reeks of C. Maybe move it to the header or something.

src/call_hook.cpp
```diff
@@ -0,0 +1,62 @@
...
+		VirtualFree(m_memory, 0, MEM_RELEASE);
+	}
+
+	void* call_hook_memory::allocate_jump_sequence(void* func)
+	{
+		m_offset = m_offset + ((16 - (m_offset % 16)) % 16); // align
+
+		*(int16_t*)&m_memory[m_offset] = 0xB848;
+		*(uint64_t*)&m_memory[m_offset+2] = (uint64_t)func;
+		*(int16_t*)&m_memory[m_offset+10] = 0xE0FF;
```
gir489returns, line 31: Can't this be made into a class? You're declaring the char* member of m_memory, when you could instead declare the members and have the compiler figure it out. I don't see why all this array shifting math nonsense needs to occur.

gir489returns (changes requested):

src/call_hook.cpp
gir489returns: Wait, is this even globally referred to? I don't see anything else using it besides this class file. g_ refers to a global identifier that can be accessed by the entire program. The only function that refers to it is the call_hook constructor.

maybegreat48: > Wait, is this even globally referred to? I don't see anything else using it besides this class file. g_ refers to a global identifier that can be accessed by the entire program. The only function that refers to it is the call_hook constructor.

Yeah, that's why I placed it in the CPP file and not in the header

gir489returns (approved):
New code looks good to me.

yubie-re:
I think going down this path to deal with integ checks is not going to work out very well in the long run. If I'm reading your code correctly (in the future it might be best to leave a few comments around, esp when you're constructing some machine code) it simply replaces a call or jmp to the function you're trying to hook. What happens if the jmp/call you're replacing is *also* integged? Rockstar has done this in the past with certain sync hooks (to attempt to prevent RID spoofing), and I'm sure they'd do it again. There is a public bypass to defeat most integs somewhere on UC from EinTim; if I understand it correctly (I only skimmed over it), it replaces the function bytes it compares with the backup bytes, so when comparing there is no differences. Might be a better route to go, or even just attacking where a certain integ is coming from might be doable. However, this is a bandaid fix which won't last very long in my opinion.

gir489returns:
@yubie-re Your argument makes no sense. You're trying to say Rockstar could fix this and we should use something else, but they could also fix that just as easily? That's not an argument, dude.

yubie-re:
Harder for R* to change arxan than to add another integ.

maybegreat48:
> Harder for R* to change arxan than to add another integ.

I don't want to give them ideas, but there's a very easy way to break the integ bypass. It has actually been done before too

xiaoxiao921:
I find it much more logical to keep this PR private for the moment ( a bit too late for that ), this way r* just patch the public bypass that everyone use and then once they release their security patch we can right after use whatever you cooked privately without telling anyone, this way we are always one patch ahead of r*, and we all know how slow they are to actually put those patches out.

xiaoxiao921 (approved):

PR #2233, opened 2023-10-08
Xml Map Loader (Menyoo)

### LINKED ISSUE #2207: [Request]: XML Map Spawner
### Problem

_No response_

### Solution

Add an XML map spawner that can parse Menyoo XML maps.

### Reason

So we can spawn stunt maps and other types of fun modded maps.

### Additional context

_No response_

Tayipc38:
Closes #2207 

This feature needs testing with various xml files, I leave that to whoever was excited to have this. Some examples to show that it does work on things I have tested:

image
image

Put your xml files in `%appdata%\YimMenu\xml_maps` and press refresh in World->Xml Maps. Menu title needs translation.


tiCkuhAr:
So far, it looks to be working well. One thing I might recommend is having it delete the last spawned map when you spawn a new one. That way, you don't end up hitting the networked object limit, so other people can see them.

Tayipc38:
> So far, it looks to be working well. One thing I might recommend is having it delete the last spawned map when you spawn a new one. That way, you don't end up hitting the networked object limit, so other people can see them.

A temporary fix would be to use the entity deleter in the World tab.

xiaoxiao921:
Thought on just making a vector<vector<entity>> var and on each xml file load/spawn you just add to that?
then a small gui for deleting the entity vector in case you wanna cleanup

Tayipc38:
> Thought on just making a vector<vector> var and on each xml file load/spawn you just add to that? then a small gui for deleting the entity vector in case you wanna cleanup

Entirely possible.

tiCkuhAr:
> Thought on just making a vector<vector> var and on each xml file load/spawn you just add to that? then a small gui for deleting the entity vector in case you wanna cleanup

Sounds good to me

tiCkuhAr:
Also, could there be a little notification that shows up when you have hit the networked object limit? That way, you know if you need to clean up or if the map you are spawning has too many objects for others to see.

gir489returns:
Is there a way to consume the code from xml_vehicles_service for this instead of adding yet another vehicle XML spawner for this?

Tayipc38:
> Is there a way to consume the code from xml_vehicles_service for this instead of adding yet another vehicle XML spawner for this?

The structures are mostly the same, but `xml_vehicles_service` code is sub-par. So actually when this is done, the opposite should be done.

lonelybud:
`[10:28:53.7345821][WARN/invoker.cpp:56] Exception caught while trying to call 0x509D5878EB39E842 native.`. I think some native is not called in fiber

mangocider:
depending on the xml map you load and the objects involved, it does tend to cause false positives for the invalid sector position crash when other players are on the objects.

also not sure if most objects even network. i usually spectate or tp players to certain objects and they either drive or fall right through, even when its not a heavy heap of objects exceeding the object pool. used the constructor lua that stand has and the same xmls seem to network mostly all objects just fine.

Tayipc38:
> depending on the xml map you load and the objects involved, it does tend to cause false positives for the invalid sector position crash when other players are on the objects.
> 
> also not sure if most objects even network. i usually spectate or tp players to certain objects and they either drive or fall right through, even when its not a heavy heap of objects exceeding the object pool. used the constructor lua that stand has and the same xmls seem to network mostly all objects just fine.

That is due to how gta handles world models. You can see them loading in for you as well. Since all of these have to be clone synced with other clients, those issues are expected. Especially with large creations.

tiCkuhAr:
After a bit more testing, there is one small thing that's a bit annoying: when you spawn a map that's in the air, some times you will fall before it fully spawns. Making it freeze your position while the map is spawning would be nice.

tiCkuhAr:
The same thing happens with vehicles; it spawns them before the map, and they end up falling before the map fully loads.

gir489returns:
> The same thing happens with vehicles; it spawns them before the map, and they end up falling before the map fully loads.

https://github.com/YimMenu/YimMenu/blob/0de9182735fd952febe5fb9bd9f447502df64764/src/util/world_model.hpp#L19-L22

This probably has to do with how the creator service he is using is waiting synchronously each time an object needs to be created for it to spawn (or fail) to proceed to the next object. Ideally, he should queue up everything that needs to spawn and do it as one giant request.

Yimura (approved):
Should be ready now as well to merge, gotta merge the translations and then I can merge the other two PR's as  well.

Didn't test any XML maps however I didn't change any of the functionality with my PR's.

PR #2249, opened 2023-10-10
feat(lua): expose take_control_of

### LINKED ISSUE #2241: [Request]: take_control_of in lua
### Solution

Add entities.take_control_of(entity, timeout).

xiaoxiao921:
Closes #2241

### DIFF
--- a/docs/lua/tables/entities.md
+++ b/docs/lua/tables/entities.md
@@ -2,7 +2,7 @@
 
 Table for manipulating GTA entities.
 
-## Functions (3)
+## Functions (4)
 
 ### `get_all_vehicles_as_handles()`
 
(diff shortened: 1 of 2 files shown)


gir489returns (changes requested):
take_control_of should default try_count to 300, as that is what the C++ counterpart does, and most callers depend on that function. We should try to keep GIGO to a minimum. If something is done a certain way on the C++ side, it should be done the same way on the Lua side, although I fail to see the usecase for this anyway.

xiaoxiao921:
Unless I'm missing something that's exactly what I did

I provide two overload to Lua. One with only one parameter, where the try count will default to 300. And one with two parameters, if the Lua caller want to override 300

Yimura (approved):

PR #2263, opened 2023-10-12
refactor(settings): Small framework for generation of nlohmann json macros

xiaoxiao921:
This PR makes it so that `NLOHMANN_DEFINE_TYPE_INTRUSIVE` don't have to be written by hand anymore.

The python script behavior is as follow:

- Fields with no Json Parser comments will be serialized to the settings.json file (if possible/compatible (basically a primitive type or if not one which already have a from_json / to_json functions defined))

- Fields marked with `// Json Parser: Generator Ignore` will never be serialized to the file no matter what. `Generator Ignore` is used when a given setting must always be excluded from JSON serialization, even when generating the structural representation of the settings. That mode is active when `SERIALIZE_EVERYTHING` is defined. This is mostly used when binding / exposing the settings to some external system outside of the DLL.

- Fields marked with `// Json Parser: Ignore` will not be serialized to the file when in non `SERIALIZE_EVERYTHING` mode. `Ignore` is used when a given setting should not be persistently stored between multiple instances of the DLL being loaded.

This is worthy to note that some fields really have no business being in the menu settings class in the first place, it seems it's getting abused by some other totally unrelated system than user settings. I consider a setting a variable that can be interacted directly by the user (through any of the views / commands for example), plenty of fields don't fill that requirement. I have marked all(?) of them using a TODO and Generator Ignore flag comments. Some of them are not marked but are ignored by the python serializer generator already because of being unserializable / unparsable due to their nature (pointers for example).

I have made sure this change is non breaking, you can verify this yourself by backing up your current settings json file, pulling these change in and recompiling and checking your new json file. Some fields were also not serde even though they were good candidates. (ptfx effect strings come to mind)


gir489returns:
How does this work? Does the Python script run while building?

xiaoxiao921:
Just found a way to make it run automatically before compilation, gonna pr soon

gir489returns:
It should only run when the settings header enters a dirty state, otherwise we'd have to recompile every class that depends on the settings header every time we run a local build and that would be a pain in the ass. I don't like this change, as we're depending on an automated runner to basically just write the declarators based on comments we leave. If you need to save everything in the class to disk situationally, why not just run the entire class through the nholman to_json function?

xiaoxiao921:
because this is fucking annoying to maintain, and the script automate the whole process
> It should only run when the settings header enters a dirty state

this is exactly the case with my last commit

gir489returns:
This now adds a dependency on the user installing Python 3 into their environment, which is annoying.

I get the following error when trying to build your branch.

```
1> [CMake] -- Could NOT find Python3 (missing: Python3_EXECUTABLE Interpreter) 
1> [CMake] Not generating settings.hpp json macros, Python not found.
```

xiaoxiao921:
Well in this case you can just add the field to the macro exactly like before

Yimura:
Will take a closer look at this tomorrow.

lonelybud:
i thought just adding `NLOHMANN_DEFINE_TYPE_INTRUSIVE ` line was simple and not complicated.

also one can split settings file into multiple data and settings -
https://github.com/lonelybud/YimMenu/commit/182e9607d81738b6a10bb2f146a3db4a7c873cfd

this can improve local compilation time
Edit: for example only recompile those whose setting/data has changed

gir489returns:
> also one can split settings file into multiple data and settings - [lonelybud@182e960](https://github.com/lonelybud/YimMenu/commit/182e9607d81738b6a10bb2f146a3db4a7c873cfd)
> 
> this can improve local compilation time Edit: for example only recompile those whose setting/data has changed

This is not true, as it will cascade into the root header, as the settings header will still have to be recompiled. Header files are analyzed and the data is generated during the precompile phase.

lonelybud:
> > also one can split settings file into multiple data and settings - [lonelybud@182e960](https://github.com/lonelybud/YimMenu/commit/182e9607d81738b6a10bb2f146a3db4a7c873cfd)
> > this can improve local compilation time Edit: for example only recompile those whose setting/data has changed
> 
> This is not true, as it will cascade into the root header, as the settings header will still have to be recompiled. Header files are analyzed and the data is generated during the precompile phase.

I tested it. so for like if you change only a file having settings related to self eg add `self.god_mod2` it compiled only 13-14 files  (instead of say 80) that depended on it -
<details>
<summary>log</summary/>

```
[build] [10/16   6% :: 5.462] Building CXX object CMakeFiles\YimMenu.dir\src\backend\looped\self\godmode.cpp.obj
[build] [11/16  12% :: 5.475] Building CXX object CMakeFiles\YimMenu.dir\src\backend\looped\self\unlimited_oxygen.cpp.obj
[build] [12/16  18% :: 5.593] Building CXX object CMakeFiles\YimMenu.dir\src\backend\looped\player\clean_player.cpp.obj
[build] [13/16  25% :: 6.056] Building CXX object CMakeFiles\YimMenu.dir\src\backend\looped\self\noclip.cpp.obj
[build] [14/16  31% :: 6.237] Building CXX object CMakeFiles\YimMenu.dir\src\version.cpp.obj
[build] [15/16  37% :: 6.293] Building CXX object CMakeFiles\YimMenu.dir\src\backend\looped\self\off_radar.cpp.obj
[build] [15/16  43% :: 6.293] Building CXX object CMakeFiles\YimMenu.dir\src\backend\looped\self\free_cam.cpp.obj
[build] [15/16  50% :: 6.453] Building CXX object CMakeFiles\YimMenu.dir\src\services\custom_text\custom_text_callbacks.cpp.obj
[build] [15/16  56% :: 10.720] Building CXX object CMakeFiles\YimMenu.dir\src\backend\looped\self\no_ragdoll.cpp.obj
[build] [15/16  62% :: 11.684] Building CXX object CMakeFiles\YimMenu.dir\src\views\self\view_self.cpp.obj
[build] [15/16  68% :: 11.755] Building CXX object CMakeFiles\YimMenu.dir\src\core\settings.cpp.obj
[build] [15/16  75% :: 12.136] Building CXX object CMakeFiles\YimMenu.dir\src\views\core\view_overlay.cpp.obj
[build] [15/16  81% :: 12.180] Building CXX object CMakeFiles\YimMenu.dir\src\backend\looped\self\toggle_passive.cpp.obj
[build] [15/16  87% :: 14.537] Building CXX object CMakeFiles\YimMenu.dir\src\native_hooks\native_hooks.cpp.obj
[build] [15/16  93% :: 17.198] Building CXX object CMakeFiles\YimMenu.dir\src\main.cpp.obj
[build] [16/16 100% :: 18.761] Linking CXX shared module YimMenu.dll
```
</details>

xiaoxiao921:
Closing because I won't be maintaining this specific system in the future. It doesnt solve issues i'm facing. Not directly related to this pr but more like a downstream thing. The pr itself does show that the system is quite limited and that there is also a lot of fields that has no business being in there. @maybegreat48 system for RDR2 is much much better as you can embed the serialization directly within the feature / command itself.

Issue #2144, opened 2023-09-18
[Improvement]: Aimbot is hot garbage.


gir489returns:
### Which feature

The aimbot uses SendInput to move the view around, when it could instead set the view angles manually, and not shake like it has Parkinson's. 

### Reason

It misses 50% of the shots due to shaking.

Peax:
> It misses 50% of the shots due to shaking.

Makes it less obvious I'd argue.

pierrelasse:
Very hot*

George0828Zhang:
I found that we can exploit the game's assisted aim, just set `GTA5.exe+257F710` (DWORD) to 0.  It works in SP (Game version 3028), I'm waiting for the AC fix before I can test it online. I'm also trying to sig the offset now.

Explanation: it's an enum for locktype for the "Expert" targetting difficulty i.e. freeaim.
```c++
class CPlayerPedTargeting__Tunables // GTA5.exe+257f148
{
    ...
    class CTargettingDifficultyInfo
    {
        enum LockType LockType; // 0x0, 0=Hardlock 1=Soft 2=None        
    } ExpertTargettingDifficultyInfo; //0x5c8
}
```
There are TONS of settings to customize for aimbot, see for yourself: https://alexguirre.github.io/rage-parser-dumps/dump.html?build=3028&game=gta5#CTargettingDifficultyInfo

gir489returns:
@George0828Zhang I really like that idea a lot. We could hijack the parameters to allow for a wide birth of aimbot features, ranging from snap rage bot to legit smooth bot.

gir489returns:
@George0828Zhang Have you done anything with your idea?

George0828Zhang:
> @George0828Zhang Have you done anything with your idea?

Nothing significant. All I know is that it locks onto torso, and it cannot lock onto peds in vehicles.

gir489returns:
> > @George0828Zhang Have you done anything with your idea?
> 
> Nothing significant. All I know is that it locks onto torso, and it cannot lock onto peds in vehicles.

I worked out where get_ped_bone while fixing ESP last night, so that's half the battle. Now I just need to write the aimbot math. However, I still don't have a way to force the player to fire via inputs and not natives. I'll continue to hammer away at this. I should be able to work out how the viewangles work by extracting the internals of GET_GAMEPLAY_CAM_ROT.

PR #2284, opened 2023-10-17
Converted all static strings to translation keys.

### LINKED ISSUE #2275: [Epic]: Replace all static text with translation keys.
### Which feature

_No response_

### Reason

Half of the menu isn't able to be translated. It looks like a giant mess when translated into another language.

### Additional context

image
image

gir489returns:
Closes #2275


gir489returns:
There are still some outliers, like https://github.com/YimMenu/YimMenu/blob/d6c32533be18aef5b35122b4f068b8844155751c/src/core/data/custom_weapons.hpp#L11-L19

It's because I can't use the _T template, it wants to use TCHAR instead. However, I think we should remove this from the header and emplace it inside the code, as there's 0 reason to have this code under "core" it does not belong. It is front-end code.

gir489returns:
OK This _**should**_ be everything.

Some changes I did that might go unnoticed:

1. Due to the nature of translation, contextual pop-ups are now and must always use the function to which it came from. The message itself should contain everything the user needs to know. 
2. I tried to when applicable reduce printf strings and replace them with compile-time std::format strings. This not only optimizes out a lot of unnecessary junk code, but it also reduces the effort on the translator's part, as they don't need to learn what printf is, although some strings may still need to be moved over to the new format, I only did the ones I saw/had patience for.
3. Childish-speak has been entirely removed. I get the nature of cheating itself is very childish, but this makes translation almost impossible sometimes. Things like emojis, tone-of-voice, vague descriptions, etc all need to be removed and replaced simple and cut-and-dry speak. Describing rapid fire as "shooting insanely fast" is beyond idiotic. 
4. Various bugs/typecasts/formatting have been fixed or addressed. For example, triggerbot had a bug where it was using the Ped's entity position instead of the bone position to shoot at. Although this isn't a proper triggerbot (it should just set the bit the game uses to interpret fire from the player), I have addressed the edge case of it not being able to hit targets in vehicles, and it will probably be addressed when we do @George0828Zhang's aimbot idea implementation for #2144.
5. British terms have all been Americanized. There was constant flip-flopping between British and American spelling of words such as Tire and Tyre. Since the file is called en_US, and most of the translations were already Americanized, I have chosen to completely move all translations to the American vocabulary. 
6.  I tried my best to remove all unnecessary .data() calls and preserve all translations as string_views and let the function that uses it call .data() as this removes code bloat. The compiler would probably optimize it out anyway, but having it go from string_view -> const char* -> string_view -> const char* seemed pointless to me. There may still be some minor outliers to find still.

DayibBaba:
My man did hard labor.

Yimura (approved):

PR #2322, opened 2023-10-21
Various translation fixes

### LINKED ISSUE #2321: [Bug]: Session types are not being properly translated.
### Describe the bug

Network -> Session Switcher is not having its list box members properly translated.

### Steps To Reproduce

image

### Expected Behavior

Should be translated.

### Logs and/or screenshots

_No response_

### Store

Rockstar

### Additional context

If I do LOG(INFO) << "BACKEND_SESSION_TYPE_JOIN_PUBLIC"_T.data(); it gets properly translated, seems to be an issue with how the lambda is being done.

### Build

### Coload

gir489returns:
Closes #2321
Removed unnecessary .data() calls from components/selectable.cpp. 
Fixed incorrect translation key usage in view_animations.cpp.


gir489returns:
I believe this will be a band aid fix and we should instead look to have a proper caching system that looks up the translation at runtime but stores the hash and translation key at compile time.

Yimura (approved):
Approving it as it should be fixed but not as a permanent solution.

Issue #2266, opened 2023-10-12
[Build]: Collapse problem


zzzzzzzzndsu:
### Describe the build issues

He was in the same situation as his friend. He went to look for the ghosts in the game. Then, he shot the npcs that were out of the game. Yim recognized him as a cheating gamer. After that, he became popular. After clicking on his game, he broke down

### Screenshots and/or Logs

```
<paste your logs here>
```
![55f52a0ef67a0a40e056c51a6d0fb699_image_ex=653a589d is=6527e39d hm=b1103151a7dda10edb3c19f1f5a8b249c91f7fef9473d1b3bbfbf3682a2aa29c =](https://github.com/YimMenu/YimMenu/assets/147297293/c70a35dc-c5a9-4efb-b7ed-2a0c19f0aedc)
![a66e6a7d60d9d36c2b50a5fc5f5c0804_image_ex=653a58bb is=6527e3bb hm=393c7350c783782658244f611e9b7629bccf25b46bdfbf80cd97721617a0f43b =](https://github.com/YimMenu/YimMenu/assets/147297293/dd33230a-aac5-4b89-9cfe-06b1f223e04e)

### IDE/Compiler

Visual Studio 2022

### Setup Guide

- [X] I confirm that I followed the setup guide correctly.

### YimMenu Source Code

- [X] I've confirmed this is an issue with building unmodified YimMenu.

sch-lda:
batter translation:
I was playing Ghost Hunt with my friends when my friend triggered the Yimmenu cheat detection, causing my friend's name to appear in red on the player list. When I clicked on the red area, my game crashed.

xiaoxiao921:
We need the %appdata%/YimMenu cout.log when that happened

lisarun:
When I inject YimMenu, there is a certain chance of a crash, which can happen. I don't know why it's like this
![1697132762870](https://github.com/YimMenu/YimMenu/assets/147649364/484086d6-0766-4d56-857c-181a0371a2a8)
![344eea7d5493303517404c442e3133b](https://github.com/YimMenu/YimMenu/assets/147649364/a648978d-1934-4e06-b192-676f638509a6)


When I look at the last few lines of the log, it looks like this


[17:49:17.5847257][WARN/reaction.cpp:25] Received Modder Detection from DEFTTTZ (142291331)
[17:49:17.6361242][WARN/reaction.cpp:25] Received Null Function Kick from DEFTTTZ (142291331)
[...]

Issue #2326, opened 2023-10-22
[Bug]: Sending Chat Msg Crashes Game


ghost:
### Describe the bug

When sending a chat message via NETWORK > Chat > the game instantly closes, giving no error or anything at all in the console log. 

### Steps To Reproduce

Start GTA V
Inject in story mode
Start new public session 
after loading in try to send any chat message using the NETWORK > Chat feature in the menu (might take a lot of tries)
Game will crash

### Expected Behavior

No crash

### Logs and/or screenshots

```
nothing out of the ordinary in cont.log so here's 2 event viewer logs.
-CRASH 1-
Faulting application name: GTA5.exe, version: 1.0.3028.0, time stamp: 0x650b237f
Faulting module name: unknown, version: 0.0.0.0, time stamp: 0x00000000
Exception code: 0xc0000005
Fault offset: 0x00007ff7ec082374
Faulting process id: 0x652c
Faulting application start time: 0x01da04b9cb065306
Faulting application path: O:\Epic Games\GTAV\GTA5.exe
Faulting module path: unknown
Report Id: 3bf6c5d6-5adc-42af-b63e-1f3a6e386613
Faulting package full name: 
Faulting package-relative application ID: 
-END-

-CRASH 2-
Faulting application name: GTA5.exe, version: 1.0.3028.0, time stamp: 0x650b237f
Faulting module name: unknown, version: 0.0.0.0, time stamp: 0x00000000
Exception code: 0xc0000005
Fault offset: 0x00007ff7ec082374
Faulting process id: 0x5044
Faulting application start time: 0x01da04ba6855cf7f
Faulting application path: O:\Epic Games\GTAV\GTA5.exe
Faulting module path: unknown
Report Id: 23b602e9-7f18-4138-be9d-0e5c22a63ee6
Faulting package full name: 
Faulting package-relative application ID: 
-END-
```

### Store

Epic Games

### Additional context

It’s important to know that the crash is extremely inconsistent. Sending one chat message could crash your game while sometimes you can send hundreds before it crashes. Although enabling chat reactions seems to increase the chance of you crashing I cannot confidently say that’s the root cause, I think it’s sending chat messages as a whole that’s behind all the crashes but I know nothing.

There was another issue thread I opened regarding this issue but it was messy so I closed it and created a proper issue that respects the format, in hopes it will bring some attention to the issue. #2281 

[...]

PR #2331, opened 2023-10-23
chore(reaction): A few minor changes

sch-lda:
1. Allow interloper_reaction to use timeout.
2. Fix crash reaction(except TSE) cannot use announce in chat.
3. Allow all friends to bypass auto kick spammers, modder flag and  reactions.
Provides an alternative for users who experience this weird bug https://github.com/YimMenu/YimMenu/issues/2266 . They can just trust their friends until someone provides valuable information for developers to fix.
4. Allow trusting specific players (bypass auto kick spammers, modder flag and reactions).
5. Trust the whole session temporarily (Avoid team members being kicked when in mission).
6. Remove duplicate logs when blocking crash (except for TSE crash)
7. remove unused gamer_instruction_kick in reaction list
8. Allow announce in team only chat


src/backend/reactions/reaction.cpp
```diff
@@ -54,10 +57,12 @@ namespace big
 	}
 
 
-	void reaction::process(player_ptr player)
+	void reaction::process(player_ptr player, bool is_crash_exclude_TSE)
```
Yimura: Every call to `reaction::process` you're hard coding the value `true/false`.

Perhaps this value should be passed in the constructor of the reaction?
https://github.com/YimMenu/YimMenu/blob/da028f03b3120a50de5e93bf1d4bc9a07fe9d3f8/src/core/settings.hpp#L176-L209
https://github.com/YimMenu/YimMenu/blob/da028f03b3120a50de5e93bf1d4bc9a07fe9d3f8/src/backend/reactions/reaction.hpp#L25

To keep it readable you could either choose a logical default in the constructor.

```cpp
reaction(const char* event_name, const char* notify_message, const char* announce_message, bool is_crash_exclude_tse = false /* true */ );
```

src/views/network/view_network.cpp
```diff
@@ -148,6 +148,14 @@ namespace big
 				ImGui::InputInt("##playercount", &g.session.player_magnet_count);
 			}
 
+			ImGui::Checkbox("Trust friends", &g.session.trust_friends);
+			if (ImGui::IsItemHovered())
+				ImGui::SetTooltip("Friends won't be flagged as modders or taken actions by reactions");
+			
+			ImGui::Checkbox("Trust this session", &g.session.trust_session);
+			if (ImGui::IsItemHovered())
+				ImGui::SetTooltip("Players in this session won't be flagged as modders or taken actions by reactions");
```
Yimura: Should be converted to `bool_command`, can be declared within this file.

For example like the below commands:
https://github.com/YimMenu/YimMenu/blob/da028f03b3120a50de5e93bf1d4bc9a07fe9d3f8/src/views/view_vehicle_control.cpp#L317-L320

src/backend/reactions/reaction.cpp
Yimura: It might just be me but the `is_crash_exclude_TSE` variable is quite confusing on what it does, could you clarify on what it means exactly? That way we could search for a different name.

src/util/notify.hpp
```diff
@@ -29,8 +29,9 @@ namespace big::notify
 			if (g.reactions.crash.log)
 				LOG(WARNING) << "Blocked " << crash << " crash from " << player->get_name() << " ("
 				             << (player->get_net_data() ? player->get_net_data()->m_gamer_handle.m_rockstar_id : 0) << ")";
-		
-			g.reactions.crash.process_common(g_player_service->get_by_id(player->m_player_id));
+			
+			g.reactions.crash.process(g_player_service->get_by_id(player->m_player_id), true);
+			g.reactions.crash.process_common(g_player_service->get_by_id(player->m_player_id), false);
```
Yimura: Ok this is weird, this feels a lot like a bandage fix, if this can be sufficiently backed up on why it was implemented like this I could allow it but for now it seems like it would be a good idea to do a partial rewrite.

Yimura (changes requested):
There's obviously the requested changes and I agree with a few of the new things you've added though the changes to `class reaction` seems a bit like a bandage fix.

If we continue on this it'd would most definitely lead to more spaghetti code down the line and cause a general dis grunt with other developers using the code base.

Yimura:
Is the below issue:
> Fix crash reaction(except TSE) cannot use announce in chat.

A consequence of a bad implementation of a snprintf string?

gir489returns:
> Is the below issue:
> 
> > Fix crash reaction(except TSE) cannot use announce in chat.
> 
> A consequence of a bad implementation of a snprintf string?

Could this be the cause of #2326?

Yimura:
> Could this be the cause of #2326?

It's possible, there was a crash like it some time ago but I don't remember too well.

sch-lda:
The code for Crash reaction was written by different contributors at different times. The script event crash and other crash reactions have different implementations. To distinguish the types of crashes, unique log and notify codes were written in notify.hpp. In commit f73c1bd, I added kick and timeout to it, but I used the generic process_common from reaction.cpp. However, the announce in chat is still not working, and duplicate logs are being written.

crash_blocked in notify.hpp has log and notify
process in reaction.cpp has announce and notify, it will call process_common 
process_common has log, add_to_player_db, kick and timeout

In order to gather all the reactions for other crashes and remove duplicate logs, I added new boolean parameters to "process" and "process_common" functions to restrict their functionality. This is indeed a kind of bandage fix.

We have different solutions:
1. Add all crash types to the constructor and use process globally. Remove crash_blocked notifications for crashes wwith unknown senders won't show. Users can then individually select the reaction for each crash type. This will result in a longer reaction list in the GUI.
2. Stop distinguishing between crash types and simply inform users with a message like "Blocked a crash from...". Use process globally.Remove crash_blocked, notifications for crashes wwith unknown senders won't show.
3. Make minimal changes by copying the code from announce_in_chat in reaction.cpp to notify.hpp and remove the bandage fix. This will still result in two lines of logs being output when blocking a crash.
4. Reassign four function's tasks
Before:
reaction::process_common log add_to_player_db kick timeout
interloper_reaction::process announce_in_chat notify (Can't log victim currently)
reaction::process announce_in_chat notify
notify::crash_blocked log notify (Writes duplicate logs and can't use announce currently)
After:
reaction::process_common add_to_player_db kick timeout
interloper_reaction::process announce_in_chat log notify
reaction::process announce_in_chat log notify
notify::crash_blocked announce_in_chat log notify
~~5. Use the bandage fix, which may make the code messy, but users won't notice any changes.~~

Which solution do you prefer?

src/backend/reactions/reaction.cpp
sch-lda: false: script event crash and all other infraction
true: exclude other crashes except for script event crash

sch-lda:
If no one makes a choice, I'll make my own. 4.

Yimura (dismissed):
Any translations that still need to be added with this PR?

sch-lda:
> Any translations that still need to be added with this PR?

All tasks have been completed.

sch-lda:
All tasks have been completed.

src/hooks/player_management/assign_physical_index.cpp
```diff
@@ -127,10 +131,10 @@ namespace big
 
 					if (g.session.lock_session && g_player_service->get_self()->is_host() && *g_pointers->m_gta.m_is_session_started)
 					{
-						if (plyr->is_friend() && g.session.allow_friends_into_locked_session)
+						if ((plyr->is_friend() && g.session.allow_friends_into_locked_session) || plyr->is_trusted)
 						{
 							g_notification_service->push_success("Lock Session",
-							    std::format("A friend with the name of {} has been allowed to join the locked session",
+							    std::format("A friend or trusted player with the name of {} has been allowed to join the locked session",
```
gir489returns, line 137: This needs to be replaced with a translation.

src/util/notify.hpp
```diff
@@ -19,17 +21,70 @@ namespace big::notify
...
 				             << (player->get_net_data() ? player->get_net_data()->m_gamer_handle.m_rockstar_id : 0) << ")";
-		
+			
+			if (g.reactions.crash.announce_in_chat)
+			{
+				g_fiber_pool->queue_job([player, crash] {
+					char chat[255];
+					snprintf(chat,
+					    sizeof(chat),
+					    std::format("{} %s tried to crash me using %s!", g.session.chat_output_prefix).data(),
```
gir489returns, line 76: This needs to be replaced with a translation.

gir489returns (changes requested):

Yimura (approved):

PR #2336, opened 2023-10-23
Prevent blank issues

pierrelasse:
I thought it would be a good idea to prevent blank issues so that people are forced to use a template.
If anyone doesn't like this idea, then just close this PR.


Yimura (approved):
Sure, no harm disabling this.

Idiots will still find ways to report multiple different things into a single issue....

PR #2361, opened 2023-10-31
Made X button unique in Persist Vehicle.

gir489returns:
Fixed issue reported by @lonelybud.

### DIFF
--- a/src/views/vehicle/spawn/view_persist_car.cpp
+++ b/src/views/vehicle/spawn/view_persist_car.cpp
@@ -135,7 +135,7 @@ namespace big
 					}
 
 					ImGui::SameLine();
-					if (ImGui::SmallButton("X"))
+					if (ImGui::SmallButton(std::string("X##" + pair).c_str()))
 					{
 						file_name_to_delete = pair;
 					}
--- a/src/views/vehicle/spawn/view_persist_car.cpp
+++ b/src/views/vehicle/spawn/view_persist_car.cpp
@@ -135,10 +135,12 @@ namespace big
 					}
 
 					ImGui::SameLine();
-					if (ImGui::SmallButton(std::string("X##" + pair).c_str()))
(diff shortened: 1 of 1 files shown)


lonelybud:
it is working fine. Files can be deleted this way. Folder need to be deleted manually.

src/views/vehicle/spawn/view_persist_car.cpp
```diff
@@ -135,7 +135,7 @@ namespace big
 					}
 
 					ImGui::SameLine();
-					if (ImGui::SmallButton("X"))
+					if (ImGui::SmallButton(std::string("X##" + pair).c_str()))
 					{
 						file_name_to_delete = pair;
 					}
```
Yimura: Probably best to use ImGui::PushID and ImGui::PopID functions.

```cpp
ImGui::PushID(pair_lower.c_str());
ImGui::SameLine();
if (ImGui::SmallButton("X"))
{
	file_name_to_delete = pair;
}
ImGui::PopID();
```

And example code in the codebase:
https://github.com/YimMenu/YimMenu/blob/990946119f81420bc1d38d0f1a2d0b580dd6ad82/src/views/settings/view_reaction_settings.cpp#L9-L12

Yimura (changes requested):

src/views/vehicle/spawn/view_persist_car.cpp
gir489returns: There is no int to use that's readily available, I'd have to make one, for something that is only used situationally.

Yimura: PushID supports char array and int.

gir489returns: So it does. Fixed.

Yimura (approved):

PR #2378, opened 2023-11-03
fix(SpeedoMeter): drawing inside of phone

Yimura:
Before:
image
After:
image


gir489returns:
Bro you shouldn't be on your phone while driving. 😔

PR #2382, opened 2023-11-04
refactor(Reactions): replace snprintf strings with std::format

### LINKED ISSUE #2372: [Improvement]: Reactions can't be translated in its current state.
### Which feature

https://github.com/YimMenu/YimMenu/blob/master/src/core/settings.hpp#L174-L223

Due to the reactions being declared with their reactive text as a precompiled statement, we would have to either fix the translation class fetching the translation at runtime or rewrite the reactions to not be declared in the settings HPP. Both are a glaring issue.

### Reason

We need to translate these strings.

### Additional context

Another @xiaoxiao921 spot.

Yimura:
Closes #2372

### DIFF
--- a/src/core/settings.hpp
+++ b/src/core/settings.hpp
@@ -191,7 +191,7 @@ namespace big
 			reaction personal_vehicle_destroyed{"Personal Vehicle Destroyed", "REACTION_PERSONAL_VEHICLE_DESTROYED_NOTIFY", "REACTION_PERSONAL_VEHICLE_DESTROYED_ANNOUNCE"};
 			reaction remote_off_radar{"Remote Off Radar", "REACTION_OFF_RADAR_NOTIFY", "REACTION_OFF_RADAR_ANNOUNCE"};
 			reaction rotate_cam{"Rotate Cam", "REACTION_ROTATE_CAM_NOTIFY", "REACTION_ROTATE_CAM_ANNOUNCE"};
-			reaction send_to_cutscene{"Send To Cutscene", "REACTION_SEND_TO_CUTSCENE_NOTIFY", "{} tried to force me into a cutscene!"};
+			reaction send_to_cutscene{"Send To Cutscene", "REACTION_SEND_TO_CUTSCENE_NOTIFY", "REACTION_SEND_TO_CUTSCENE_ANNOUNCE"};
 			reaction send_to_location{"Send To Location", "REACTION_SEND_TO_LOCATION_NOTIFY", "REACTION_SEND_TO_LOCATION_ANNOUNCE"};
 			reaction send_to_interior{"Send To Interior", "REACTION_SEND_TO_INTERIOR_NOTIFY", "REACTION_SEND_TO_INTERIOR_ANNOUNCE"};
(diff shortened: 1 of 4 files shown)


gir489returns:
If we're gonna touch reactions we should address #2372.

Yimura:
> If we're gonna touch reactions we should address #2372.

Easy to address just gotta move the strings to the translation repo and add a few calls to the translation service.

Yimura:
Should be ready to merge though I'd like to get this tested by a few people first. I'll merge the translations already so they're present.

sch-lda:
image
notify works for me

sch-lda:
image
It seems to disrupt crash announce. Attacker and crash type are missing

Yimura:
> image
> 
> It seems to disrupt crash announce. Attacker and crash type are missing

I'll check, I might have missed those translation strings.

sch-lda:
> > image
> > It seems to disrupt crash announce. Attacker and crash type are missing
> 
> I'll check, I might have missed those translation strings.

image
That could be the problem.

Yimura:
@sch-lda should be fixed now, I have also adjusted the string for the Chinese translations.

sch-lda:
Okay, I'll test it when I get home.

sch-lda:
> @sch-lda should be fixed now, I have also adjusted the string for the Chinese translations.

image

src/core/settings.hpp
```diff
@@ -173,52 +173,54 @@ namespace big
...
+			reaction force_teleport{"Force Teleport", "REACTION_FORCE_TELEPORT_NOTIFY", "REACTION_FORCE_TELEPORT_ANNOUNCE"};
+			reaction gta_banner{"GTA Banner", "REACTION_GTA_BANNER_NOTIFY", "REACTION_GTA_BANNER_ANNOUNCE"}; // please don't enable this
+			reaction kick_from_interior{"Kick From Interior", "REACTION_KICK_FROM_INTERIOR_NOTIFY", "REACTION_KICK_FROM_INTERIOR_ANNOUNCE"};
+			reaction mc_teleport{"MC Teleport", "REACTION_MC_TELEPORT_NOTIFY", "REACTION_MC_TELEPORT_ANNOUNCE"};
+			reaction network_bail{"Network Bail", "REACTION_NETWORK_BAIL_NOTIFY", "REACTION_GENERIC_KICK_ANNOUNCE"};
+			reaction personal_vehicle_destroyed{"Personal Vehicle Destroyed", "REACTION_PERSONAL_VEHICLE_DESTROYED_NOTIFY", "REACTION_PERSONAL_VEHICLE_DESTROYED_ANNOUNCE"};
+			reaction remote_off_radar{"Remote Off Radar", "REACTION_OFF_RADAR_NOTIFY", "REACTION_OFF_RADAR_ANNOUNCE"};
+			reaction rotate_cam{"Rotate Cam", "REACTION_ROTATE_CAM_NOTIFY", "REACTION_ROTATE_CAM_ANNOUNCE"};
+			reaction send_to_cutscene{"Send To Cutscene", "REACTION_SEND_TO_CUTSCENE_NOTIFY", "{} tried to force me into a cutscene!"};
```
sch-lda: There is an untranslated text.

Yimura: Good catch, during all the copy pasting yesterday I forgot to put the translation string.

PR #2389, opened 2023-11-06
New stuff - 2

thehorizon84:
changelog:
+ World -> Gravity Editor (with presets for ease of use)
+ World -> Waypoint Beacon
+ World -> Objective Beacon
+ World -> Time And Weather -> Ground Snow

+ Self -> Mobile -> Request Gun Van (spawns the gun van right in front of you no matter where you are)

### DIFF
--- a/src/backend/looped/world/groundsnow.cpp
+++ b/src/backend/looped/world/groundsnow.cpp
@@ -10,13 +10,9 @@ namespace big
 	{
 		using looped_command::looped_command;
 
-		virtual void on_enable() override
-		{
-			g_tunables_service->set_tunable<bool>(RAGE_JOAAT("TURN_SNOW_ON_OFF"), true);
-		}
-
 		virtual void on_tick() override
 		{
+			g_tunables_service->set_tunable<bool>(RAGE_JOAAT("TURN_SNOW_ON_OFF"), true);
 		}
 
 		virtual void on_disable() override
--- a/src/backend/looped/world/groundsnow.cpp
+++ b/src/backend/looped/world/groundsnow.cpp
@@ -1,0 +1,29 @@
+#include "backend/looped_command.hpp"
+#include "fiber_pool.hpp"
+#include "gta_util.hpp"
+#include "script.hpp"
+#include "services/tunables/tunables_service.hpp"
+
+namespace big
+{
+	class ground_snow : looped_command
+	{
+		using looped_command::looped_command;
+
+		virtual void on_enable() override
+		{
+			g_tunables_service->set_tunable<bool>(RAGE_JOAAT("TURN_SNOW_ON_OFF"), true);
+		}
+
+		virtual void on_tick() override
+		{
+		}
+
+		virtual void on_disable() override
+		{
+			g_tunables_service->set_tunable<bool>(RAGE_JOAAT("TURN_SNOW_ON_OFF"), false);
+		}
+	};
+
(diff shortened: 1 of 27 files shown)


src/backend/looped/world/groundsnow.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "fiber_pool.hpp"
+#include "gta_util.hpp"
+#include "script.hpp"
+#include "services/tunables/tunables_service.hpp"
+
+namespace big
+{
+	class ground_snow : looped_command
+	{
+		using looped_command::looped_command;
```
Yimura, line 11: A bool command may be better suited for this since you're not making use of the on_tick function.

Yimura (changes requested):

src/backend/looped/world/groundsnow.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+		virtual void on_enable() override
+		{
+			g_tunables_service->set_tunable<bool>(RAGE_JOAAT("TURN_SNOW_ON_OFF"), true);
+		}
```
gir489returns, line 16: You may want to move this to on_tick, as it's possible the tunable service will overwrite this the next time they refresh, which is usually done during transition or session change. That is your goal I'm assuming, since you are saving it to a variable called ground_snow.

src/views/world/view_gravity.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+		if (g.world.gravity.modify_gravity)
+		{
+			ImGui::SliderFloat("GRAVITY_LEVEL"_T.data(), &g.world.gravity.current_gravity, 0.f, 1000.f, "%.1f");
+
+			ImGui::Separator();
+			components::sub_title("GRAVITY_LUNAR_PRESETS"_T);
+
+			// non-planets
+			components::button("GRAVITY_MOON"_T, [] { g.world.gravity.current_gravity = 1.6f; });
```
gir489returns, line 17: Refactor this to be a drop-down. There is no reason to have every single item be a button. You also do not need components::button, as you are just writing to something within the cheat's own memory and do not need access to the natives. This can be done within the normal ImGui callback scope.

gir489returns (changes requested):

src/views/world/view_gravity.cpp
thehorizon84: This layout was just to ease testing the different values. Didn't think it looked bad ingame so didn't change it but I'll switch it up for you guys, no problem!

src/backend/looped/world/groundsnow.cpp
thehorizon84: Haven't looked into if this specific tunable gets refreshed anytime, but now you mention it you are also able refresh other player's tunables through a script event so I'll move it to on_tick, appreciate the heads-up.

src/backend/looped/world/groundsnow.cpp
thehorizon84: I'll move it to on_tick and keep it as a looped_command, explanation in gir489returns's comment.

thehorizon84: Also is it possible to send you a message on telegram or discord? I have some plans I'd like to present for future PR's and I'm not sure whether they fit here or not.

Yimura: Here's a link to our Matrix server, a lot more private compared to alternatives.

https://matrix.to/#/#yimmenu:matrix.org

thehorizon84: Thanks!

thehorizon84:
Closing, re-uploading fixed PR.

PR #2393, opened 2023-11-07
New stuff - 2.1

thehorizon84:


src/util/mobile.hpp
```diff
@@ -190,5 +190,28 @@ namespace big::mobile
...
+			        forward_vector.z,
+			        25.f,
+			        &spawn_point))
+			{
+				*scr_globals::gun_van.as<Vector3*>() = spawn_point;
+
+				return g_notification_service->push_success("Request Service", "The Gun Van is now spawned near your location, keep in mind that this is local and cannot be seen by other players.");
+			}
+
+			g_notification_service->push_warning("Request Service", "Couldn't find a suitable spawn point for the Gun Van, try moving to a different location.");
```
Yimura: Missing from translations.

Yimura (dismissed):

Yimura:
@thehorizon84, you can keep pushing your changes on your repository @ master. There's no need to close your PR and open a new PR to submit more changes.

src/util/mobile.hpp
Yimura: I approved by accident, anyways still need them added to the translations.

thehorizon84: These need to be translated? I can't seem to find any other notifications that are translated. For example the notification: "Mechanic is not ready to deliver a vehicle right now." is not in "en_US".

src/backend/looped/world/groundsnow.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+namespace big
+{
+	class ground_snow : looped_command
+	{
+		using looped_command::looped_command;
+
+		virtual void on_tick() override
+		{
+			g_tunables_service->set_tunable<bool>(RAGE_JOAAT("TURN_SNOW_ON_OFF"), true);
```
gir489returns, line 15: Do not change this, but just for reference, each Tunable/Global/Local/whatever is always 4 bytes wide. It is generally good practice to type it as BOOL not bool, as BOOL is the C-style boolean, and will force the compiler to write 4 bytes, whereas bool will only force it to write 1 byte. Assuming a 0-contamination environment, there should never be a scenario where the upper 3 bytes of this are written to, but the script VM only types things as 4 bytes wide, there is no actual type of bool to the script VM.

That being said, I am the only one of the YimMenu contributors who does this, as:
A: I learned on C a million years ago, never gave it up.
B: I always program the cheat to use whatever the game is using, an old habit from the C64 days as poking outside the memory space of what you were given always resulted in a HCF scenario.

You are free to ignore this.

thehorizon84: Always open to new information. I'll keep it in mind! :)

src/util/mobile.hpp
Yimura: Just search for g_notification_service->push in the project, you'll find plenty of results making use of translations.

https://github.com/search?q=repo%3AYimMenu%2FYimMenu+g_notification_service-%3Epush&type=code

thehorizon84: Fixed now.

src/backend/looped/world/groundsnow.cpp
maybegreat48: I believe the local weather changer enables snow automatically

thehorizon84: > I believe the local weather changer enables snow automatically

It doesn't. The weather "snow" or "snowlight" (can't remember what its called) just enables the snowing effect. Ground snow has to be toggled through the tunable.

maybegreat48: It's [supposed to](https://github.com/YimMenu/YimMenu/blob/master/src/backend/looped/world/weather_override.cpp#L31), though it's been a while since I've used this menu

thehorizon84: > It's [supposed to](https://github.com/YimMenu/YimMenu/blob/master/src/backend/looped/world/weather_override.cpp#L31), though it's been a while since I've used this menu

Oh never noticed that, although when I tried changing the weathers ingame nothing happened. Anyhow the option allows the user to enable ground snow whenever, not just with the snow weather.

Yimura (approved):

PR #2411, opened 2023-11-12
Player Database Enhancement

thehorizon84:
Added 'Send Friend Request' through the player's RID.


pierrelasse:
Cool cool, could you mybe make the 'Join Session' and 'Invite Player' next to each other with ImGui::SameLine to safe space? :)
image

Yimura (dismissed):

thehorizon84:
Added more stuff:

Player Database:
- 'open social club profile' via rockstar id, and replaced ugly native method in player selected window
- as requested, SameLine'd the options in pairs of 2, looks nice ingame

 Player:
 - Spectate camera distance overrider

thats about it i think

gir489returns:
Why do we need this? What's wrong with using the natives?

pierrelasse:
> Spectate camera distance overrider

Does that mean you fixed the issue that when you use freecam, esp uses the position of the player instead of the cam? If so or if not thanks for the pr anyways.

thehorizon84:
> Why do we need this? What's wrong with using the natives?

Ease of use. 1 line instead of 3. 

Also, can you even do these things with natives when the player isn't in your lobby?

pierrelasse:
Would it make sense to add "Spoof as friend" or something like that which adds the player as your friend in socialclub locally?
Most of the menus have this and sometimes when using the "Join Session" button, it gives an error but when you add them as a friend and use the native join game button, it says they use a diffrent targeting mode and when you accept, you join completly fine. Maybe the friend spoof could be a hotfix?

Yimura:
@pierrelasse, I'd consider the comments under a PR to report bugs with the current implementation or to recommend certain changes to be made. Not to add more features to be added into the same PR creating more work for the person making the PR, creating more work for us reviews and making it take longer before it can be merged into master.

src/views/players/view_player.cpp
```diff
@@ -7,6 +7,14 @@ namespace big
 	{
 		ImGui::Checkbox("SPECTATE"_T.data(), &g.player.spectating);
 
+		if (g.player.spectating)
+		{
+			ImGui::Checkbox("OVERRIDE_SPECTATE_CAM_DISTANCE"_T.data(), &g.player.override_cam_distance);
+
+			if (g.player.override_cam_distance)
+				ImGui::SliderInt("OVERRIDE_SPECTATE_CAM_DISTANCE_INT"_T.data(), &g.player.cam_distance, 1, 200);
```
Yimura: Should probably be bool and int commands, you can register them at the bottom of the `src/backend/looped/player/spectate.cpp` file.

Combined with `components::command_checkbox` and `components::command_int_slider` functions.

Yimura (changes requested):
As an addendum to what @gir489returns said, more signatures will require more work to update once a game update drops. Imo it's fine in the current use case as it's useful for the Player DB to directly go from RID to the function.

pierrelasse:
> @pierrelasse, I'd consider the comments under a PR to report bugs with the current implementation or to recommend certain changes to be made. Not to add more features to be added into the same PR creating more work for the person making the PR, creating more work for us reviews and making it take longer before it can be merged into master.

I just wanted to ask if he changed this or what he meant with this. And yes i agree that it would be better to make a bug report instead of telling someone in their pr to fix some issue. Again my comment was intended to be a question.

who-care123:
> > Why do we need this? What's wrong with using the natives?
> 
> Ease of use. 1 line instead of 3.
> 
> Also, can you even do these things with natives when the player isn't in your lobby?

Yes actually we can, we can use native `NETWORK_HANDLE_FROM_USER_ID`/`NETWORK_HANDLE_FROM_MEMBER_ID`(rockstar id) to get the gamer handle then use native `NETWORK_ADD_FRIEND`
But use what method depends on personal preference

src/views/players/view_player.cpp
thehorizon84: Not sure I agree, what's bad about the current implementation? It's supposed to run while spectate is active, so having it inside the spectate function only makes sense, does it not?

gir489returns:
> > > Why do we need this? What's wrong with using the natives?
> > 
> > 
> > Ease of use. 1 line instead of 3.
> > Also, can you even do these things with natives when the player isn't in your lobby?
> 
> Yes actually we can, we can use native `NETWORK_HANDLE_FROM_USER_ID`/`NETWORK_HANDLE_FROM_MEMBER_ID`(rockstar id) to get the gamer handle then use native `NETWORK_ADD_FRIEND` But use what method depends on personal preference

I prefer if we use the natives to accomplish this. Reducing the call set by 2 calls on something that isn’t done every frame is not an argument. Its called only when the user clicks the button and I doubt that would be very often.

thehorizon84:
I like to reverse, purely relying and using prebuilt functions that never change (natives), to me, is not an exciting way to mod a game. That way you never get a grasp on how things are actually done, you just simply know what it does. And for this case, I see this as an enhancement as it defeats the need to use another thread and call 2 additional natives to achieve the same result (goes for both signatures).

At the end of the day you've been here longer than I have, it's you guys who decide how you want it. I just like to share random stuff that I find interesting.

thehorizon84:
Wasn't referring purely to this exact scenario. This case was just a pointer to a help-function that calls the scapi functions through a rid, nothing special at all, it just a quick enhancement in my eyes.

I'm talking about it in general. I've released 2 signatures so far (these here) that have been related to natives and if this is the critique it gets, I will keep it in mind to keep that stuff for my own entertainment, no worries lmao

Yimura:
> I just wanted to ask if he changed this or what he meant with this. And yes i agree that it would be better to make a bug report instead of telling someone in their pr to fix some issue. Again my comment was intended to be a question.

No if there's a problem with the current PR it's fine to comment about it here, if there's a problem and the PR has been merged then you make an issue.

src/views/players/view_player.cpp
thehorizon84: > Should probably be bool and int commands, you can register them at the bottom of the `src/backend/looped/player/spectate.cpp` file.
> 
> Combined with `components::command_checkbox` and `components::command_int_slider` functions.

Shouldn't the main spectating bool also be a command::checkbox then, its just a regular ImGui::Checkbox?

This really overcomplicates things, I would rather just remove the entire function at this point, it works as intended with the least amount of code.

thehorizon84: > Should probably be bool and int commands, you can register them at the bottom of the `src/backend/looped/player/spectate.cpp` file.
> 
> Combined with `components::command_checkbox` and `components::command_int_slider` functions.

Apologies, sort of misunderstood what you implied with this, for some reason I thought you meant I needed to create a whole new class just for this addition. Anyways, it's fixed now!

Yimura (approved):

PR #2416, opened 2023-11-14
Removed globalplayer_bd IsInvisible set call.

### LINKED ISSUE #2413: [Improvement]: make your vehicle visible, even when you are invisible
### Which feature

self > invisible also makes the vehicle you are driving invisible for other people, even if vehicle > vehicle invisibility is off

### Reason

so i can troll as a self driving car

### Additional context

_No response_

gir489returns:
Closes #2413


gir489returns:
The user has confirmed this solves his issue.

Yimura (approved):

PR #2419, opened 2023-11-15
Clone Debug Globals into Debug Locals

gir489returns:
Redesigned components::input_text to return the InputText result, and add an overloaded method to take a std::string. Fixed Debug Globals having static text and not using components::input_text to prevent sending the input to the game. Fixed view_debug_threads::thread_states not properly ending its string causing ImGui to read more than it should from memory. Made Debug Globals cap the input index to UINT32, as the user could potentially overflow the get_ptr function and cause the game to crash.


gir489returns:
image
[locals.json](https://github.com/YimMenu/YimMenu/files/13369031/locals.json)

Yimura (approved):

PR #2421, opened 2023-11-16
Fix jittering Player ESP

### LINKED ISSUE #2388: [Bug]: ESP flickers its position on players that have velocity
### Describe the bug

Players that are either in very fast cars or flying seem to have their ESP flicker and move.

### Steps To Reproduce

https://www.youtube.com/watch?v=W_jhhotBoiQ

(The box's further down position was purposeful testing, disregard it)

### Expected Behavior

The ESP box should maintain perfect position on the model as other hacks do. 

### Logs and/or screenshots

_No response_

### Store

Rockstar

### Additional context

I tried changing over to get_bone_coords, thinking that maybe m_navigation was the issue, but it is not. 

### Build

### Coload

gir489returns:
Closes #2388


gir489returns:
@maybegreat48 The function that's signatured here for get_bones just eventually calls what we have labeled as _0x68 in fwEntity's Vtable. I tried calling it with PedBones::SKEL_Pelvis, but it didn't return anything useful. Do you know what this int parameter is?

src/views/esp/view_esp.cpp
```diff
@@ -180,8 +183,6 @@ namespace big
 
 		if (const auto draw_list = ImGui::GetBackgroundDrawList(); draw_list)
 		{
-			draw_player(g_player_service->get_self(), draw_list);
```
Yimura, line 183: Removing this will make it so you'll never draw yourself.

There would still be the remaining setting:
https://github.com/YimMenu/YimMenu/blob/af287d259328c55b7a2d689fe2013d1521d62efe/src/core/settings.hpp#L942

src/views/esp/view_esp.cpp
```diff
@@ -20,12 +21,14 @@ namespace big
 
 	void esp::draw_player(const player_ptr& plyr, ImDrawList* const draw_list)
 	{
-		if (!plyr->is_valid() || !plyr->get_ped() || !plyr->get_ped()->m_navigation)
+		if (!plyr->is_valid() || !plyr->get_ped() || !plyr->get_ped()->m_navigation || !plyr->get_ped()->m_model_info)
 			return;
 		if (g.esp.hide_self && plyr->is_valid() && plyr->id() == g_player_service->get_self()->id())
```
Yimura: You never call to draw your own player (you removed it further down).

Making this piece of code redundant and it's associated setting:
https://github.com/YimMenu/YimMenu/blob/af287d259328c55b7a2d689fe2013d1521d62efe/src/core/settings.hpp#L942

Yimura (changes requested):

src/views/esp/view_esp.cpp
gir489returns: This option is functionally useless. I figured it was just there only to check if the loop was returning a player as well. My only guess was this was included for testing purposes, as there is never a reason the user should want or have the ESP drawn on themselves.

Yimura (approved):

PR #2429, opened 2023-11-18
Skeleton ESP

thehorizon84:
Untested and might look like ass, put this together in about 10 minutes from knowledge i got from making cs trainers a while back, but it should work here aswell, I don't see why it wouldn't. (Can't test due to detections)


gir489returns:
https://youtu.be/AM2okr58OGI

src/views/esp/view_esp.cpp
```diff
@@ -71,6 +71,87 @@ namespace big
...
+					PedBones::SKEL_R_Hand
+				};
+
+				std::vector<rage::fvector4> bones = {};
+				const size_t skel_size            = 18;
+
+				for (int i = 0; i < skel_size; i++)
+				{
+					rage::fvector4 out = {};
+					if (g_pointers->m_gta.m_get_ped_pone(plyr->get_ped(), out, bone_ids[i]))
```
gir489returns, line 104: This function has a spelling mistake in it, and it was corrected in master, reflect it in your branch.

gir489returns (changes requested):

src/views/esp/view_esp.cpp
thehorizon84: fixed, and pushed a tiny improvement to cut down on repeated code, plus I think I missed to draw spine -> pelvis (not sure if spine and pelvis connect or not) so added that just incase

src/views/esp/view_esp.cpp
```diff
@@ -85,6 +85,85 @@ namespace big
...
+				    PedBones::SKEL_L_Foot,
+				    PedBones::SKEL_R_Foot,
+				    PedBones::SKEL_L_Clavicle,
+				    PedBones::SKEL_R_Clavicle,
+				    PedBones::SKEL_L_UpperArm,
+				    PedBones::SKEL_R_UpperArm,
+				    PedBones::SKEL_L_Forearm,
+				    PedBones::SKEL_R_Forearm, 
+					PedBones::SKEL_L_Hand, 
+					PedBones::SKEL_R_Hand
```
Yimura: Different indentations, try to run the code formatter on this file to fix it.

Yimura (changes requested):

src/views/esp/view_esp.cpp
thehorizon84: Fixed

src/views/esp/view_esp.cpp
```diff
@@ -85,6 +85,85 @@ namespace big
 			if (distance < g.esp.box_render_distance[1] && distance > g.esp.box_render_distance[0] && g.esp.box)
 				draw_list->AddRect({esp_x - (62.5f * multplr), esp_y - (175.f * multplr)}, {esp_x - (62.5f * multplr) + (125.f * multplr), esp_y - (175.f * multplr) + (350.f * multplr)}, esp_color);
 
+			if (distance < g.esp.skeleton_render_distance[1] && distance > g.esp.skeleton_render_distance[0] && g.esp.skeleton)
+			{
+				std::vector<PedBones> bone_ids = // size: 18
```
gir489returns, line 90: Is there a reason this isn't a const or constexpr?

gir489returns (changes requested):

src/views/esp/view_esp.cpp
```diff
@@ -85,6 +85,85 @@ namespace big
...
+					PedBones::SKEL_L_Forearm, 
+					PedBones::SKEL_R_Forearm, 
+					PedBones::SKEL_L_Hand, 
+					PedBones::SKEL_R_Hand
+				};
+
+				std::vector<rage::fvector4> bones = {};
+				const size_t skel_size            = 18;
+
+				for (int i = 0; i < skel_size; i++)
```
gir489returns, line 115: use for (auto bone : bone_ids) and pass bone to get_ped_bone. You do not need or care about the index.

src/views/esp/view_esp.cpp
```diff
@@ -85,6 +85,85 @@ namespace big
...
+				{
+					rage::fvector4 out = {};
+					if (g_pointers->m_gta.m_get_ped_bone(plyr->get_ped(), out, bone_ids[i]))
+					{
+						bones.push_back(out);
+					}
+				}
+
+				if (bones.size() == skel_size)
```
gir489returns, line 124: This check is erroneous. Instead, do a model check or something.

src/views/esp/view_esp.cpp
```diff
@@ -85,6 +85,85 @@ namespace big
...
+					for (int i = 0; i < skel_size; i++)
+					{
+						if (!g_pointers->m_gta.m_get_screen_coords_for_world_coords(bones[i].data, &bones_screen[i].x, &bones_screen[i].y))
+						{
+							render_skeleton = false;
+							break;
+						}
+						else
+						{
+							auto buffer   = ImVec2((float)*g_pointers->m_gta.m_resolution_x * bones_screen[i].x, (float)*g_pointers->m_gta.m_resolution_y * bones_screen[i].y);
```
gir489returns, line 139: ImVec2 is explicitly float, why are you C-style casting? Shouldn't the compiler do this implicitly?

src/views/esp/view_esp.cpp
thehorizon84: Oh yea, was originally planning on making a side-function to grab all the required bone ids, but got lazy and just decided to do it manually. Looking back must've just forgot to class it as a const afterwards. I'll fix it in a min

gir489returns (changes requested):

src/views/esp/view_esp.cpp
thehorizon84: why is it wrong lol? it works, just to check if all the bones got added, if one didn't make it through for some reason then the size would be below 18, returning false.

You're the one who added the sig, if its hardcoded to always return true then yes this check is useless, haven't taken a look at it in IDA.

src/views/esp/view_esp.cpp
thehorizon84: i just copied the resolution stuff from another esp category in the func, as i said put this together in about 10 minutes so that tiny little detail didn't really cross my mind

src/views/esp/view_esp.cpp
thehorizon84: Uhm no, I need to have it as a "for (int)" to get the position offset for "bones_screen[i]" as you can see in the next for-loop. Arguably, this one you highlighted could be switched, although i like to keep everything looking the same. So if Yimura wants it changed I will change it.

gir489returns: Nothing you said made sense. The index has nothing to do with the size, as that comes from the bones vector. You can simply refactor the code like this and it will work the same way. 

```
for (auto bone : bone_ids)
{
	rage::fvector4 out = {};
	if (g_pointers->m_gta.m_get_ped_bone(plyr->get_ped(), out, bone))
	{
		bones.push_back(out);
	}
}
```

src/views/esp/view_esp.cpp
gir489returns: We are getting crashes from get_ped_bone, and this might be too late in the call chain to deal with it. 

See #2433 (@maybegreat48 personally reported it to me too, which lead to) and #2431. We can't figure out why they were crashing, and if you add this, you'll probably just resurrect the problem for people who have skeletin mode on. You could surround the call in a try catch and swallow the exception, but that's a terrible solution in my book. Ideally, we should create a thread-safe way to call the natives from the DX layer like YimMenuv2 has, maybegreat suggested we port it over so we don't have to deal with this shit, but I digress, If this scenario occurs, it will be too late in the call chain to deal with it.

src/views/esp/view_esp.cpp
thehorizon84: Read what I said again

src/views/esp/view_esp.cpp
thehorizon84: I'll take a look at it, using a signature to a function that's a native shouldn't create any threading issues though. I've never experienced that personally when I experimented with those threading related stuff.

thehorizon84: Solved it, @gir489returns

Switching byte 60 to ? on the sig seemed to do the trick. I've pushed the sig and now this commit is ready to be merged. No issues (from my testing)

fixed_skelly

Yimura (dismissed):

maybegreat48:
> [get_ped_bone sig fix](https://github.com/YimMenu/YimMenu/pull/2429/commits/57253a19c209eb20fd00ee750097c0ca1190ac24)

It still lands in the same function for me

thehorizon84:
> > [get_ped_bone sig fix](https://github.com/YimMenu/YimMenu/pull/2429/commits/57253a19c209eb20fd00ee750097c0ca1190ac24)
> 
> It still lands in the same function for me

Yes, it's still the same function, but for some reason this sig works and his doesn't. I don't know what's wrong, all I did was to re-scan it and create a new signature and for some reason that did the trick. But hey, I don't make the rules lol

src/views/esp/view_esp.cpp
```diff
@@ -69,6 +69,85 @@ namespace big
...
+					PedBones::SKEL_L_Forearm, 
+					PedBones::SKEL_R_Forearm, 
+					PedBones::SKEL_L_Hand, 
+					PedBones::SKEL_R_Hand
+				};
+
+				std::vector<rage::fvector4> bones = {};
+				const size_t skel_size            = 18;
+
+				for (int i = 0; i < skel_size; i++)
```
gir489returns, line 99: Maybe if you fucking wrote it in English I could fucking understand it.

thehorizon84: Calm, there's no need for aggression mate. My english is fine. 

Ask "wdym" next time and I'll try to give you a more indepth explanation. Let's try to keep a friendly tone between one another. 👍

pierrelasse:
ESP looking kinda spicy, would it be very easy to add the line thiccness? And maybe also options to select what bones to draw?

I just noticed that the healthbar looks a bit out of place without the box. Would it make sense to move it below the name when the box is disabled maybe?

pierrelasse:
I crashed multiple times on this PR's binary. Can anyone tell if the crash is related or if i should wait until the download is available again and i make a new issue?
[cout.log](https://github.com/YimMenu/YimMenu/files/13425910/cout.log)

thehorizon84:
> I crashed multiple times on this PR's binary. Can anyone tell if the crash is related or if i should wait until the download is available again and i make a new issue? [cout.log](https://github.com/YimMenu/YimMenu/files/13425910/cout.log)

This is basically all the tests I can do due to detections in online. My concern was that this function only worked on the local CPed handle, but that isn't true as you can see in the picture below. It seems to be working fine, so it shouldn't be an issue passing another player's CPed handle through this function - in other words, I have no idea why you're crashing with this ESP.

npctest

pierrelasse:
I often crash when joining online or joining a new session from a session. Maybe this helps.

thehorizon84:
> I often crash when joining online or joining a new session from a session. Maybe this helps.

It doesn't. That doens't even sound related to ESP if you're crashing when "joining a new session from a session". If this was ESP related you would've crashed in the old session and not when joining the new one. ESP can't work in one session then not work in another, it either works or it doesn't. Try adding a CPed valid check in your source right before the get_ped_bone function, you might be passing some invalid handles for whatever reason.

pierrelasse:
**While** joining a session then...

thehorizon84:
As I said, try adding a check to see if all the CPed's your passing through the function are valid. Because get_ped_bone will crash if the handle is invalid, which is odd considering its a bool option, it should just return false.

gir489returns:
You're not going to fix this on your end, as I've said twice now, there's a crash relating to get_ped_bone. It has to do with CSekeltonData and it not being fully loaded while transitioning. However, without grabbing the data manually, you're not going to know if it's there or not. 

image

i is garbage, but they don't check for that in the loop, so it just crashes. It checks if it's null, but not if it's garbage. In maybegreat's case, his was 0xBF81BBB8. 

I vote to either ditch this function and rewrite it with YimMenuv2's native caller, or we scrap this entirely, it causes too many problems.

thehorizon84:
Why not just wait?

In the cases I've been presented you crash while transitioning into a new lobby because (I would assume) you're enabling the esp as soon as players appear. Why not just add some transition checks to only enable the esp once you're on your feet and let the esp figure the skeleton data up until that point (I think you can check the state with a global?) it shouldn't take longer than that.

maybegreat48:
> Why not just wait?
> 
> In the cases I've been presented you crash while transitioning into a new lobby because (I would assume) you're enabling the esp as soon as players appear. Why not just add some transition checks to only enable the esp once you're on your feet and let the esp figure the skeleton data up until that point (I think you can check the state with a global?) it shouldn't take longer than that.

I've had this crash mid-session too

gir489returns:
> Why not just wait?
> 
> In the cases I've been presented you crash while transitioning into a new lobby because (I would assume) you're enabling the esp as soon as players appear. Why not just add some transition checks to only enable the esp once you're on your feet and let the esp figure the skeleton data up until that point (I think you can check the state with a global?) it shouldn't take longer than that.

Literally blaming the user. That's not an argument.

thehorizon84:
What? I'm not blaming the user, I'm proposing a solution to what I **think** is the problem out of information I got from users/issues. What I've heard is that people crash instantly when joining online, so from experience I would assume it would have to with calling a function too early.

No idea how you classed this as "blaming the user". Give your passive-aggressive comments to someone else, will ya?

Yimura:
The code works and it looks nice in-game however we'll close this for now as @thehorizon84's master branch is in a dirty state.

We'll come back to this whenever we've found a non-janky fix for the crashes.

PR #2443, opened 2023-11-22
Override game checksum

maybegreat48:
Instructions to test:
1. Start GTA without any mods except YimMenu
2. Enable the `TODO` option in `Network > Spoofing` and load into an invite-only session
3. Restart game with mods, load menu, and then try joining a session


src/backend/looped/system/override_game_hashes.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+	char* base64_encode(const unsigned char* data, size_t input_length, size_t& output_length)
+	{
+		const int mod_table[] = {0, 2, 1};
+
+		output_length = 4 * ((input_length + 2) / 3);
+
+		char* encoded_data = (char*)malloc(output_length);
```
Yimura: Replace with `new char[]` or since you're disposing of it immediately after using it anyways consider switching to `unique_ptr<char[]>`.

src/backend/looped/system/override_game_hashes.cpp
```diff
@@ -0,0 +1,129 @@
...
+		virtual void on_enable() override
+		{
+			memcpy(original_checksum_block, *g_pointers->m_gta.m_game_checksum_data, checksum_block_size);
+
+			if (g.spoofing.last_game_version != std::stoi(g_pointers->m_gta.m_game_version))
+			{
+				g.spoofing.last_game_version = std::stoi(g_pointers->m_gta.m_game_version);
+				size_t out_size;
+				char* encoded = base64_encode((const unsigned char*)*g_pointers->m_gta.m_game_checksum_data, checksum_block_size, out_size);
+				g.spoofing.game_checksum_data_b64 = std::string(encoded, out_size); free(encoded);
```
Yimura: `free` will be redundant after switching to `unique_ptr`.

Kryo77:
Could you not just spoof CRC in CNetworkAssetVerifier or does that not work for everything?

maybegreat48:
> Could you not just spoof CRC in CNetworkAssetVerifier or does that not work for everything?

The full CRC isn't stored in `CNetworkAssetVerifier`, and is calculated in multiple places, including the matchmaking code and some more obfuscated stuff. So I thought that writing to that struct would be a better idea

Kryo77:
> > Could you not just spoof CRC in CNetworkAssetVerifier or does that not work for everything?
> 
> The full CRC isn't stored in `CNetworkAssetVerifier`, and is calculated in multiple places, including the matchmaking code and some more obfuscated stuff. So I thought that writing to that struct would be a better idea

Ah makes sense

PR #2452, opened 2023-11-25
Removed GET_STAT_HASH_FOR_CHARACTER_STAT dependency.

gir489returns:


gir489returns:
@Yimura I don't know if you want to change out the code for this, the way I had it before was a 1:1 copy to what the scripts were doing, this also makes it slightly harder to update, as subsequent additions to the script will have to reverse GET_STAT_HASH_FOR_CHARACTER_STAT to find the new stat, assuming they add anything to this game anymore, also assuming GET_STAT_HASH_FOR_CHARACTER_STAT still exists.

Yimura (approved):
I'm fine with this change as the previous one was hard coding property indexes per char.
This is more readable in the long run.

PR #2453, opened 2023-11-26
refactor(Metrics): filter out "bad" metrics

Yimura:
Thinking about adding the "bad metric" log behind a toggle to prevent users from worrying about the log.

### DIFF
--- a/src/hooks/info/prepare_metric_for_sending.cpp
+++ b/src/hooks/info/prepare_metric_for_sending.cpp
@@ -3,41 +3,52 @@
 
 namespace big
 {
-	const auto bad_metrics = std::to_array<const std::string>({
+	const auto bad_metrics = std::unordered_set<std::string>({
+	    "REPORTER",
+	    "REPORT_INVALIDMODEL",
+	    "MEM_NEW",
+	    "DEBUGGER_ATTACH",
 	    "DIG",
+	    "XP_LOSS",
 	    "AWARD_XP",
 	    "CF",
 	    "CC",
 	    "CNR",
 	    "SCRIPT",
+	    "CHEAT",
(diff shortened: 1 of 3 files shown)


src/hooks/info/prepare_metric_for_sending.cpp
```diff
@@ -3,50 +3,53 @@
...
+	    "CF",
+	    "CC",
+	    "CNR",
+	    "SCRIPT",
+	    "AUX_DEUX",
+	    "HARDWARE_OS",
+	    "HARDWARE_GPU",
+	    "HARDWARE_MOBO",
+	    "HARDWARE_MEM",
+	    "HARDWARE_MEM",
```
tupoy-ya: Duplicate `HARDWARE_MEM` entry.

Yimura: I will adjust the list with a couple more ones when I get home.

gir489returns:
Will this fix detections?

Yimura:
> Will this fix detections?

No but it's just improving old code so why not.

gir489returns:
> > Will this fix detections?
> 
> No but it's just improving old code so why not.

The title is misleading.

Yimura:
> The title is misleading.

Well it sorta says what it does without the "allow normal metrics to go through".

gir489returns:
> > The title is misleading.
> 
> Well it sorta says what it does without the "allow normal metrics to go through".

I'm confused. Are we sending the normal metrics or not?

Yimura:
> I'm confused. Are we sending the normal metrics or not?

Yes with this PR we're allowing normal ones and filtering out "bad" ones.

PR #2469, opened 2023-12-01
feat(anticheat): Fix nullsub scanning in PE header

yubie-re:
Should fix uncommon crash after #2463 @gir489returns

### DIFF
--- a/src/pointers.cpp
+++ b/src/pointers.cpp
@@ -1735,7 +1735,7 @@ namespace big
         // Nullsub
         {
             "NS",
-            "C3",
+            "90 C3",
             [](memory::handle ptr)
             {
                 g_pointers->m_gta.m_nullsub = ptr.as<void(*)()>();


gir489returns:
Thise closes https://github.com/YimMenu/YimMenu/issues/2468

gir489returns (approved):

maybegreat48:
Can't you simply create a empty function and use that instead? The AC doesn't check if the address is within the GTA module

gir489returns:
> Can't you simply create a empty function and use that instead? The AC doesn't check if the address is within the GTA module

We already have one.

https://github.com/YimMenu/YimMenu/blob/062c95b3745ff622b60f4a7bb0c564639b6f2d90/src/native_hooks/all_scripts.hpp#L182

While the parameters are not the same, the compiler should still optimize it out to simple RETN block.

yubie-re:
The reasoning behind scanning for a nullsub instead of implementing our own func is so the game won't crash when the DLL is uninjected, rather than evading ac checks.

gir489returns:
> The reasoning behind scanning for a nullsub instead of implementing our own func is so the game won't crash when the DLL is uninjected, rather than evading ac checks.

Wouldn't that leave the game in a dirty state? Shouldn't we undo that?

yubie-re:
It could be argued that allowing the AC to run again after uninjection could be dangerous. I figured that it is for the best that it stay patched even after uninjected, as it shouldn't effect much and may make stuff just a little bit safer

lonelybud:
is it possible that to see these logs after player join the session? bcz i am seeing it after starting the game second time

`14:53:11.2458370][INFO/queue_dependency.cpp:63] Blocking AC Verifier 95bd58`

Yimura (approved):

PR #2493, opened 2023-12-07
feat(LuaMgr): Disable and Enable scripts from UI and prevent crash when renaming the `scripts/` dir.

### LINKED ISSUE #2437: [Bug]: Crash when renaming scripts folder
### Describe the bug

See Steps To Reproduce and Excpected Behavior

### Steps To Reproduce

1. Start GTA
2. Inject YimMenu
3. Ensure Auto Reload Scripts is on (atleast i have it on)
4. Rename the `%appdata%/YimMenu/scripts` folder to `scripts_` or whatever
5. Look at the error in the console

### Expected Behavior

No errors
Does the same when a script FILE is deleted or renamed

### Logs and/or screenshots

[cout.log](https://github.com/YimMenu/YimMenu/files/13415159/cout.log)

### Store

Steam

### Language

English

### YimMenu Language

English (US)

### Additional context

_No response_

### Build

### Coload

Yimura:
Fixes #2437

### Todo
 - [x] Add Disable/Enable functionality
 - [x] Add Disable All/Enable All functionality
 - [x] Cleanup UI

### DIFF
--- a/src/file_manager.cpp
+++ b/src/file_manager.cpp
@@ -44,15 +44,15 @@ namespace big
     
 	std::filesystem::path file_manager::ensure_folder_exists(const std::filesystem::path folder_path)
     {
-        bool create_path = !std::filesystem::exists(folder_path);
+        bool create_path = !exists(folder_path);
 
-        if (!create_path && !std::filesystem::is_directory(folder_path))
+        if (!create_path && !is_directory(folder_path))
         {
-            std::filesystem::remove(folder_path);
+            remove(folder_path);
             create_path = true;
         }
         if (create_path)
-            std::filesystem::create_directory(folder_path);
+            create_directories(folder_path);
 
         return folder_path;
     }
--- a/src/file_manager/folder.cpp
+++ b/src/file_manager/folder.cpp
@@ -19,6 +19,15 @@ namespace big
 		return file(m_folder_path / file_path);
 	}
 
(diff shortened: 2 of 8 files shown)


gir489returns:
Since you're tackling this, there was a (semi) related xiao and I discovered yesterday while testing the io manager. Scripts like Casino Pacino that now depend on other scripts (like the JSON script) will contaminate the scripts window. When we eventually create some kind of automated script downloader, this will be confusing to many people as to why there's random scripts in the script window that they didn't ask for or don't understand and they may reload it or even delete it.

We both agreed that for the time being that really only developers use the Lua runner, so it isn't that big of an issue. But even if we wanted to, I don't really know how we could fix this? If we add a whitelist like "dependency" folder, that will hide it from the requires, as all the Lua scripts have to compile and be loaded into memory, since we're not using LuaJIT, we can't eagerly instantiate the script from disk.

All that being said, it's a super non-issue, but some kind of system to omit dependency scripts would be nice. I just have no idea how the system would work.

Yimura:
> All that being said, it's a super non-issue, but some kind of system to omit dependency scripts would be nice. I just have no idea how the system would work.

Perhaps Lua scripts should report this through the same method that they would which version of Lua API they support of YimMenu.

Yimura:
Ready for review

gir489returns (commented):
Looks good to me.

PR #2502, opened 2023-12-09
Enable strict sol for type safety

gir489returns:


gir489returns:
[According to the sol documentation](https://sol2.readthedocs.io/en/latest/safety.html) there is a runtime cost associated with doing this change. When writing the Lua scripts, it is very helpful to have things like SOL_SAFE_FUNCTION_CALLS, as it will tell you when you pass incorrectly typed arguments or even an invalid number of arguments to a function. Because of the nature of Lua, a nil argument gets silently converted to 0 and if the 2nd argument happens to be an integer or float, the function will happily cast this and silently do nothing about it, even though that many not be what the developer intended.

Having strict type checking while developing the script is almost an necessity, however, regular users of the menu do not need this at all. Maybe we have something where the nightly omits this define?

xiaoxiao921:
I personally think it's straight better to have it like this because it  makes debugging Lua scripts easier. If this doesn't go through for any reason the include change should definitely go through due to the way the sol lib is defined. it's pretty standard to have this lib only included once because it's not even pragma once and the compiler will have to parse it everytime it's included

xiaoxiao921 (approved):

PR #2503, opened 2023-12-10
fix(player_command): execution failure

Aure7138:
Fix the issue where player_command always fails when executed with the me/self parameter.

### DIFF
--- a/src/backend/player_command.cpp
+++ b/src/backend/player_command.cpp
@@ -36,7 +36,7 @@ namespace big
 		g_fiber_pool->queue_job([this, args, ctx] {
 			command_arguments new_args(m_num_args.value(), args);
 
-			if (g_player_service->get_self()->id() == args.get<int>(0))
+			if (g_player_service->get_self()->id() == args.get<uint8_t>(0))
 			{
 				execute(g_player_service->get_self(), new_args, ctx);
 				return;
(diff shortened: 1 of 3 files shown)


src/backend/player_command.cpp
```diff
@@ -36,7 +36,7 @@ namespace big
 		g_fiber_pool->queue_job([this, args, ctx] {
 			command_arguments new_args(m_num_args.value(), args);
 
-			if (g_player_service->get_self()->id() == args.get<int>(0))
+			if (g_player_service->get_self()->id() == args.get<uint8_t>(0))
```
gir489returns, line 39: While this enforces size safety on the compiler's end of this particular function, this doesn't enforce size (or even type) safety on the caller's side, are we sure that when this is called from the command runner and not the ImGui menu that we'll be given an int8?

Aure7138: I believe this example can elucidate the situation.

```cpp
#include<iostream>

#pragma pack(push, 8)
struct structA
{
	uint8_t var_0 = 8; //0-255
	char pad_0 = 'a';
	char pad_1 = 'b';
	char pad_2 = 'c';
};
#pragma pack(pop)

int main()
{
	structA a;
	int num = static_cast<int>(a.var_0);
	int& refNum = reinterpret_cast<int&>(a.var_0);
	std::cout << ((a.var_0 == num) ? "true" : "false") << '\n'; //output: true
	std::cout << ((a.var_0 == refNum) ? "true" : "false") << '\n'; //output: false
}
```

gir489returns: I see, so if we downcast to an int8, we'll always get the original caller's value.

Yimura (approved):

PR #2504, opened 2023-12-10
fix(player_command): memory leak

Aure7138:
Fix the undefined behavior brought about by memory leaks when dealing with more than one parameter (e.g., /wanted).

### DIFF
--- a/src/backend/player_command.cpp
+++ b/src/backend/player_command.cpp
@@ -92,7 +92,7 @@ namespace big
 			result.push(plyr_id);
 		}
 
-		new_args.resize(args.size() - 1);
+		new_args.reserve(args.size() - 1);
 		std::copy(++args.begin(), args.end(), new_args.begin());
 		// for (int i = 1; i < args.size(); i++)
 		// 	new_args.push_back(args[i]);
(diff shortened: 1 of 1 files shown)


gir489returns:
Why don't we just use new_args = args and let the equals operator deal with this?

Aure7138:
> Why don't we just use new_args = args and let the equals operator deal with this?

We should remove the first data in "args"

gir489returns:
> > Why don't we just use new_args = args and let the equals operator deal with this?
> 
> We should remove the first data in "args"

The behavior of the equals operator is well defined, and will always perform a deep copy, replacing anything in the original vector itself.

image

Aure7138:
You are correct; we can also proceed in this manner.

```cpp
new_args = args;
new_args.erase(new_args.begin());
```

gir489returns:
new_args's memory should be released once it falls out of scope.

Aure7138:
> new_args's memory should be released once it falls out of scope.

Indeed, a std::vector should automatically release memory once it goes out of scope; it serves as a practical example of RAII (Resource Acquisition Is Initialization).

Yimura (approved):

PR #2509, opened 2023-12-12
Updated vehicle_abilities.cpp to work with latest GTAV-Classes PR

horizon-84:
https://github.com/Yimura/GTAV-Classes/pull/148


Yimura (changes requested):
`GIT_TAG` under scripts/gtav-classes.cmake needs to be updated.

Yimura (approved):

PR #2516, opened 2023-12-12
feat(Pointers): Update signatures for 1.68 b3095

horizon-84:
Havent injected YimMenu, these are fixs to failed scans from my own project.

### DIFF
--- a/src/pointers.cpp
+++ b/src/pointers.cpp
@@ -202,7 +202,7 @@ namespace big
         // Trigger Script Event
         {
             "TSE",
-            "45 8B F0 41 8B F9 48 8B EA",
+            "45 8B F8 41 8B F9",
             [](memory::handle ptr)
             {
                 g_pointers->m_gta.m_trigger_script_event = ptr.sub(0x1C).as<decltype(gta_pointers::m_trigger_script_event)>();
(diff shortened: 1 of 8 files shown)


src/pointers.cpp
```diff
@@ -256,7 +256,7 @@ namespace big
         // Read Bitbuffer Boolean
         {
             "RBB",
-            "E8 ? ? ? ? 84 C0 74 41 48 8D 56 2C",
+            "E8 ? ? ? ? 84 C0 74 ? 48 8D 56 ? 44 8D 43 ? 48 8B CF E8 ? ? ? ? 84 C0 74 ? 48 8D 56 ? 44 8D 43 ? 48 8B CF E8 ? ? ? ? 84 C0 74 ? 48 8D 56 ? 44 8D 43 ? 48 8B CF E8 ? ? ? ? 84 C0 74 ? BB",
```
Yimura: This signature should be updated to be shorter, it won't degrade the speed of the scanner much however it's most likely better to use a shorter signature that also points to it.

Be it from a different part of the assembly, adding, ripping; differently.

Yimura (changes requested):

src/pointers.cpp
```diff
@@ -202,7 +202,7 @@ namespace big
         // Trigger Script Event
         {
             "TSE",
-            "45 8B F0 41 8B F9 48 8B EA",
+            "45 8B F8 41 8B F9",
```
Aure7138, line 205: This update is actually for SCRIPT::TRIGGER_SCRIPT_EVENT, not SCRIPT::SEND_TU_SCRIPT_EVENT

Yimura:
PR was superseded by #2519.

PR #2524, opened 2023-12-13
Made LS Customs accept any car.

gir489returns:
The Start LS Customs button is still broken, but this allows you to take any car to LS Customs. This could be used for recovery if you know what you're doing, however, so could Start LS Customs before.

### DIFF
--- a/src/backend/script_patches.hpp
+++ b/src/backend/script_patches.hpp
@@ -60,20 +60,15 @@ namespace big
 		    5,
 		    {0x2E, 0x01, 0x00},
 		    &g.vehicle.ls_customs}); // disable camera
-		g_script_patcher_service->add_patch(
-		    {RAGE_JOAAT("carmod_shop"), "carmod_shop1", "2D 02 10 00 00 2C", 5, {0x71, 0x2E, 0x02, 0x01}, &g.vehicle.ls_customs});
-		g_script_patcher_service->add_patch(
-		    {RAGE_JOAAT("carmod_shop"), "carmod_shop2", "2D 00 B8 00 00", 5, {0x2E, 0x00, 0x00}, &g.vehicle.ls_customs});
-		g_script_patcher_service->add_patch(
-		    {RAGE_JOAAT("carmod_shop"), "buy mods for free in SP", "06 1F 56 ? ? 62 ? ? ? 56", 0, {0x2B, 0x2B, 0x55}, &g.vehicle.ls_customs}); // buy mods for free in SP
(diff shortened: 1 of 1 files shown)


src/backend/script_patches.hpp
```diff
@@ -60,20 +60,15 @@ namespace big
...
 		    &g.vehicle.ls_customs}); // disable camera
-		g_script_patcher_service->add_patch(
-		    {RAGE_JOAAT("carmod_shop"), "carmod_shop1", "2D 02 10 00 00 2C", 5, {0x71, 0x2E, 0x02, 0x01}, &g.vehicle.ls_customs});
-		g_script_patcher_service->add_patch(
-		    {RAGE_JOAAT("carmod_shop"), "carmod_shop2", "2D 00 B8 00 00", 5, {0x2E, 0x00, 0x00}, &g.vehicle.ls_customs});
-		g_script_patcher_service->add_patch(
-		    {RAGE_JOAAT("carmod_shop"), "buy mods for free in SP", "06 1F 56 ? ? 62 ? ? ? 56", 0, {0x2B, 0x2B, 0x55}, &g.vehicle.ls_customs}); // buy mods for free in SP
-		g_script_patcher_service->add_patch(
-		    {RAGE_JOAAT("carmod_shop"), "allow all vehicles", "2D 03 16 00 00 5D", 5, {0x72, 0x2E, 0x03, 0x01}, &g.vehicle.ls_customs}); // allow all vehicles
+		g_script_patcher_service->add_patch({RAGE_JOAAT("carmod_shop"), "carmod_shop1", "2D 02 10 00 00 2C", 5, {0x71, 0x2E, 0x02, 0x01}, nullptr});
```
lonelybud, line 63: check for `&g.vehicle.ls_customs` is replaced with `nullptr` ?

Yimura: If it's `nullptr` the patch will be permanently enabled, without requiring a value to be toggled:
https://github.com/YimMenu/YimMenu/blob/efef962fe8406e88f57982606a2530b1de1b4838/src/services/script_patcher/script_patch.cpp#L65-L69

gir489returns:
> I wonder if there are any adverse effects from permanently enabling some of these patches, @gir489returns did you notice anything off while testing this?

Nope. I'm not sure what the purpose of the other two hooks are/were, but I took my car to several LS Customs that were off the street and were already in my garage. The game basically treated it like it was a normal car. I left the camera hijack as a bool toggle.

Issue #2589, opened 2023-12-18
[Request]: 1. xml vehicles search or 2. xml vehicle preview


bob6456:
### Solution

1. add a search bar to search for an xml vehicle or 
2. add a preview for xml vehicles

### Reason

1. if you have to many xml vehicles you cant find the car you looking for 
2. you need to spawn the xml vehicle to see how it looks

TheGreenBandit:
Adding searching in #2594, test and see how you like it

PR #2594, opened 2023-12-18
Feat (Xml Vehicles): Added searching for filenames

TheGreenBandit:
partly solves #2589


TheGreenBandit:
if yall feel like closing it close it, even if xml is just for transferring to json, it dont really matter as this is just a search bar lol

Yimura:
Personally, I'm fine with adding this. However, it would be nice to have the design and behaviour be uniform between all tabs (where it makes sense to add it).

gir489returns:
> Personally, I'm fine with adding this. However, it would be nice to have the design and behaviour be uniform between all tabs (where it makes sense to add it).

I agree with the latter.

TheGreenBandit:
Alright, i'll take a look at the gui for json spawning and transfer it over to xml, saving wont be included however as you said it isn't for saving files.

DayibBaba:
> XML Vehicles were only meant to import the files from Menyoo into GTA so that you could resave them into the JSON format and use that interface. However, if there is a usecase that the XML format performs that the JSON does not, then the UI should be redesigned to match New/Personal/Persist.
> 
> If the XML does not serve a higher purpose than what I mentioned, then this PR is invalid.

I don't recall that purpose. I simply wrote an interpeter using the functions of YimMenu to support spawning Menyoos xml vehicles. The facilities of the feature can be refactored to suit the theme better, such as the UI. And also, your idea of using this interpeter for xml maps feature is sound and should be realized to remove redundancy. Thats a mistake on my part.

gir489returns:
@TheGreenBandit This ticket is 2 months stale with no pushes to your branch. What is the status of the changes we requested?

Yimura (changes requested):
Unify design and behaviour between all the tabs (wherever it seems logic and the data is there).

Adding this code will add more tech debt to the whole code base. Once is fine but doing it consistently is bad.

Yimura:
No response from PR author, neither commenting on the requested changes or committing anything.

As such the PR will be closed.

PR #2604, opened 2023-12-19
Refactored Allow All Vehicles

gir489returns:
Refactored Allow All Vehicles to use the current seat the ped is in to fetch the animation info that the car is currently using. This fixes an edge case with certain vehicles like hanging off the back of the garbage truck not working.


src/backend/looped/vehicle/allow_all_weapons.cpp
```diff
@@ -13,48 +14,49 @@ namespace big
 	{
 		CVehicle* vehicle_ptr = (CVehicle*)g_pointers->m_gta.m_handle_to_ptr(self::veh);
```
Yimura: Couldn't we access the current vehicle from `g_local_player->m_vehicle`?

Not really a hard-requirement, would just avoid the call to m_handle_to_ptr. Keep in mind that you may need to implement extra logic to see if the person is still actually in the vehicle.

gir489returns: > Couldn't we access the current vehicle from `g_local_player->m_vehicle`?
> 
> Not really a hard-requirement, would just avoid the call to m_handle_to_ptr.

That's a good idea. This was originally past ported from BBv2 and I just converted the function calls to whatever Yim was using. Due to the refactor, I wouldn't even need the vehicle pointer anymore, except to call set_vehicle_model_flag. Since I do a safety check on the g_local_player pointer.

Yimura (approved):

Issue #2586, opened 2023-12-18
[Bug]: Drop-down selection list bug inside specific windows


gamerboy1992:
### Describe the bug

When Pressing on "Extra Info" on myself or any other player and going to the drop-down selection of "Chat Command Permissions" and I press it again anywhere and it just closes the entire "Extra Info" window.

### Steps To Reproduce

1. Go to your profile or any other player profile in the main menu (under Settings)
2. Press "Extra Info"
3. Press the selection thing in "Chat Command Permissions"
4. Whilst having the options listed, pressing anywhere on the screen will just close the entire "Extra Info" window

### Expected Behavior

I would expect to just go back to the normal "Extra Info" window, deselecting that drop-down list of options.

### Store

Epic Games

### Language

English

### YimMenu Language

English (US)

### Build

- [X] I've confirmed the issue exists on the latest version of YimMenu

### Coload

- [X] I've tested this without co-loading anything.

gamerboy1992:
Upon further testing, this issue seems to be present with a lot of drop-down list style of selections.
Here's what I found so far:
Self->Enable PTFX (Options)->Both of them
Vehicle->Fun Features->Custom Vehicle Weapons (Options)->All of them
Network->Teleport->All of them

From this, I can conclude that this bug is present only within the windows that are created for detailed options and such. ("Extra Info" style windows, I'm not sure what else to call them)

Yimura:
Found the culprit:
https://github.com/YimMenu/YimMenu/blob/9bedcca8a40d89653f442d4ad07e727dc954f674/src/gui/components/options_modal.cpp#L22-L23

Yimura:
Doesn't seem like it's possible to fix this without some very hacky code.

PR #2621, opened 2023-12-20
feat(cmake): update ImGui to latest release

Yimura:
Had to fix some stuff and remove a Lua binding that doesn't exist anymore in the latest version of ImGui.

### DIFF
--- a/scripts/imgui.cmake
+++ b/scripts/imgui.cmake
@@ -2,7 +2,7 @@ include(FetchContent)
 FetchContent_Declare(
     imgui
     GIT_REPOSITORY https://github.com/ocornut/imgui.git
-    GIT_TAG        94c46d74869ec991c101c187088da0f25d6c8e40
+    GIT_TAG        v1.90
     GIT_PROGRESS TRUE
 )
 message("ImGui")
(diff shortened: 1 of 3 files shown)


gir489returns:
Is there a reason we're fast forwarding our version? Did they add/implement/fix something we need?

Yimura:
Tried to fix #2586 but couldn't find a good, non-hacky way of fixing it without handicapping the current behaviour.

Decided since I did the work to fix the issues the new version introduced, I'll open a PR anyway.

gir489returns:
> Tried to fix #2586 but couldn't find a good, non-hacky way of fixing it without handicapping the current behaviour.
> 
> Decided since I did the work to fix the issues the new version introduced, I'll open a PR anyway.

As long as it doesn't break anything, I'm OK with this, but I think we should regression test this before pushing it, they could've changed the way a lot of these behave, I had this happen before with BBv2 when we would constantly keep playing catch up with ImGui. Eventually we just decided to say fuck it and stay on an old version because it was too much work constantly updating and fixing ImGui.

Yimura:
> As long as it doesn't break anything, I'm OK with this, but I think we should regression test this before pushing it, they could've changed the way a lot of these behave, I had this happen before with BBv2 when we would constantly keep playing catch up with ImGui. Eventually we just decided to say fuck it and stay on an old version because it was too much work constantly updating and fixing ImGui.

Yup, I completely agree however previously we didn't stick to actual version numbers and instead were just using the latest commit from their active dev branch it wasn't too good either. At least with this we'll start by using their versioned releases.

gir489returns (commented):
Tested last night, no issues detected.

PR #2625, opened 2023-12-21
Fix debug build

xiaoxiao921:
allow injecting the menu with debug flags, remove unused pattern double ?? support as it was never used and was causing UB since a long time now (out of range array access)

### DIFF
--- a/CMakeSettings.json
+++ b/CMakeSettings.json
@@ -1,6 +1,17 @@
-﻿{
+﻿﻿{
   "configurations": [
     {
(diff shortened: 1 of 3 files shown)


gir489returns:
What's the difference between this and just using the PDB?

Yimura:
> What's the difference between this and just using the PDB?

No issues with certain breakpoints never being hit because of compiler optimizations.

Yimura (approved):

PR #2630, opened 2023-12-21
chore(log): Co-loading detected

Aure7138:
idea comes from @Yimura


Aure7138:
The detection would be ineffective if YimMenu is the first injected, but currently, the pattern mode in Pointers (like REH) doesn't support injection afterward. So, this PR is useless for now

PR #2649, opened 2023-12-27
Optimizations

maybegreat48:
* The invoker has been rewritten to be more efficient and thread-safe
* Zydis has been replaced by the Hacker Disassembler Engine found in MinHook
* Compiler optimizations have been enabled, halving DLL size
* Refactored hooking code into a new directory
* Bug fixes


src/lua/natives/__pycache__/natives_gen.cpython-311.pyc
Yimura, line 1: Please remove `__pycache__` folder and add it to `.gitignore`.

src/native_hooks/tunables.hpp
```diff
@@ -21,7 +21,7 @@ namespace big
 
 		inline void NETWORK_ACCESS_TUNABLE_BOOL_MODIFICATION_DETECTION_REGISTRATION_HASH(rage::scrNativeCallContext* src)
 		{
-			g_tunables_service->register_tunable(src->get_arg<Hash>(1), src->get_arg<BOOL*>(2));
+			g_tunables_service->register_tunable(src->get_arg<Hash>(1), src->get_arg<bool*>(2));
```
Yimura, line 24: Intentional change from `BOOL` to `bool`?

Yimura (changes requested):

Yimura:
Any chance we can rename the `scripts/` folder to `cmake/` and start moving some python helper scripts into a new `scripts/` folder?

I have been planning on doing this but haven't made to time to get around to it.

src/lua/natives/__pycache__/natives_gen.cpython-311.pyc
maybegreat48: VS didn't show me that for some reason

src/native_hooks/tunables.hpp
maybegreat48: No, but it shouldn't matter

xiaoxiao921:
Very confused about the release with deb info stuff, do we even want all of this? Why don't we just remove all the info and just use a good classic release mode? Also the nightly never used it also unless it I'm missing something

Yimura:
> Very confused about the release with deb info stuff, do we even want all of this? Why don't we just remove all the info and just use a good classic release mode? Also the nightly never used it also unless it I'm missing something

He probably wants users to start taking the pdb as well so the crash dumps are easier to decipher.

xiaoxiao921:
https://github.com/return-of-modding/ReturnOfModding/commit/aff7004aec33aea67e00dd72bf84d6bec760bd2d

Yimura (approved):
Fine by me to merge it now, I or someone else can do the change in regards to the scripts/ directory.

PR #2651, opened 2024-01-01
Added x64-Release-FastCompile build.

gir489returns:
The optimizations added in #2649 cause excessive build and linkage time when developing. The Debug build throws a lot of pedantic errors that the runtime swallows. Until it is fixed, I suggest we use this instead, and also remove a redundant compile flag.


CMakeLists.txt
```diff
@@ -68,10 +68,12 @@ add_compile_definitions(YimMenu
 
 # Optimizations
 if(MSVC)
-  set(CMAKE_MODULE_LINKER_FLAGS
-      "${CMAKE_MODULE_LINKER_FLAGS} /LTCG /OPT:REF,ICF /GUARD:NO")
-
-  set(CMAKE_CXX_FLAGS_RELWITHDEBINFO
-      "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} /O2 /Ox /Oi /Ot /Oy /Ob3 /sdl- /GL /GF /GS- /Gw")
-  string(REPLACE "/Ob1" "/Ob3" CMAKE_CXX_FLAGS_RELWITHDEBINFO ${CMAKE_CXX_FLAGS_RELWITHDEBINFO})
+    if(OPTOMIZE)
```
Yimura: `OPTOMIZE` => `OPTIMIZE`

Yimura (changes requested):

xiaoxiao921:
How about we jst have Release, ReleaseWithDebInfo, Debug config options?

Release by default should have optimization enabled anyway, i'm not sure why all the things in the cmake file are needed

xiaoxiao921:
CMAKE_CXX_FLAGS_RELWITHDEBINFO and CMAKE_CXX_FLAGS_RELWITHDEBINFO are supposed to be identical in term of optimizations, it seems that none of the added cmake changes added by @maybegreat48  are actually doing anything and if they are it just means it's a cmake bug

gir489returns:
> How about we jst have Release, ReleaseWithDebInfo, Debug config options?
> 
> Release by default should have optimization enabled anyway, i'm not sure why all the things in the cmake file are needed

ReleaseWithDebInfo compiles it with debug info, which allows our crash logger to figure out the filename/line number where the crash happened. If we even allow people to compile without it, it would make those YimMenu.dll+0xworthless offsets that mean nothing in the log. We already have enough problems with that as-is. I vote no to this.

xiaoxiao921:
For what it's worth the nightly already do exactly that, it compiles as Release instead of RelWithDebInfo

gir489returns:
> For what it's worth the nightly already do exactly that, it compiles as Release instead of RelWithDebInfo

Which is see column A:

>We already have enough problems with that as-is.

gir489returns:
@Yimura This PR is ready, as MBG has looked over the changes I did.

xiaoxiao921:
> > For what it's worth the nightly already do exactly that, it compiles as Release instead of RelWithDebInfo
> 
> Which is see column A:
> 
> > We already have enough problems with that as-is.

Which makes it even more questionable? Why is the nightly still compiling as Release instead of DebInfo?
Also the CMake flags need to be cleaned, none of them are actually doing something, expect the /gs- one

PR #2652, opened 2024-01-16
fix(lua): allow natives to be called with nil strings.

xiaoxiao921:


Yimura (approved):
Yeah you had to fix the script after I did the move to the scripts folder, I didn't want to figure out where and what I need to change stuff so I left it to you until we needed it again.

PR #2653, opened 2024-01-21
Optimized Aimbot code.

gir489returns:
@navmodder Please check this commit over and make sure that the math and behavior still operate the same.

### DIFF
--- a/src/backend/looped/weapons/aimbot.cpp
+++ b/src/backend/looped/weapons/aimbot.cpp
@@ -101,10 +101,9 @@ namespace big
 				if (aim_lock.x == 0.f && aim_lock.y == 0.f && aim_lock.z == 0.f)
 					return;
 
-				float RADPI = 180.0f / std::numbers::pi;
-				float camera_heading = atan2f(camera_target.x, camera_target.y) * RADPI;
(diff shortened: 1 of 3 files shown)


gir489returns:
@navmodder I noticed there are several issues with the aimbot, some of which I fixed myself, some of which I'm going to fix, and some of which I have no idea how to fix.

Issue #1: The aimbot sometimes aims above their head. This seems to be inverse to the distance of their position. I tried changing the magnitude calculation to use hypot (a more accurate version of the Euclidean Distance calculation you were doing) but the issue still occurs.
Issue #2: The visibility checks are dogshit. It tries to aim at targets behind vehicles or behind cover. I tried to compensate for this with various inline natives, however we're probably going to have to result to using a shapetest for each NPC, however, this will delay the visibility result until the shapetest is finished, which may not be until a few ticks later.
Issue #3: The triggerbot doesn't work at all. It does the typical nonsense fire_bullet_between_coords, but it never hits them, and it fires slower than the actual weapon. We should look into just setting the firing flag to true for the player's input.
Issue #4: There's no way to select the hitbox the aimbot targets.

Yimura (approved):

PR #2654, opened 2024-01-26
(improvement) load_3d_ground_at_coords

99Anvar99:
I have been testing teleports, and how to improve loading ground at coords faster. Without having to for loop 10K times, and i have came with this solution.

Testing outputs: 

- Empty areas: 0 - 10 attempts.
- Populated areas: around 20 - 50 attempts.
- Highest point on the map: Around 135 attempts

### DIFF
--- a/src/util/entity.cpp
+++ b/src/util/entity.cpp
@@ -201,7 +201,7 @@ namespace big::entity
 
 	    do {
 	        found_ground = MISC::GET_GROUND_Z_FOR_3D_COORD(location.x, location.y, max_ground_check, &ground_z, FALSE, FALSE);
-	        STREAMING::REQUEST_ADDITIONAL_COLLISION_AT_COORD(location.x, location.y, location.z);
+	        STREAMING::REQUEST_COLLISION_AT_COORD(location.x, location.y, location.z);
 
 	        if (current_attempts % 10 == 0)
 	        {
--- a/src/util/entity.cpp
+++ b/src/util/entity.cpp
@@ -194,26 +194,35 @@ namespace big::entity
 	bool load_ground_at_3dcoord(Vector3& location)
 	{
 		constexpr float max_ground_check	= 1000.f;
-		constexpr int max_attempts			= 200;
-		float ground_z						= location.z + 1.f;
+		constexpr int max_attempts			= 300;
+		float ground_z						= location.z;
 		int current_attempts				= 0;
 
-		while (!MISC::GET_GROUND_Z_FOR_3D_COORD(location.x, location.y, max_ground_check, &ground_z, FALSE, FALSE))
+		while (!MISC::GET_GROUND_Z_FOR_3D_COORD(location.x, location.y, max_ground_check, &ground_z, FALSE, FALSE) && current_attempts != max_attempts)
 		{
-			STREAMING::REQUEST_ADDITIONAL_COLLISION_AT_COORD(location.x, location.y, location.z);
-			if (current_attempts >= max_attempts)
+			if (MISC::GET_GROUND_Z_FOR_3D_COORD(location.x, location.y, max_ground_check, &ground_z, FALSE, FALSE))
 			{
-				return false;
+				break;
 			}
-			current_attempts++;
+
+			STREAMING::REQUEST_ADDITIONAL_COLLISION_AT_COORD(location.x, location.y, location.z);
+
 			if (current_attempts % 10 == 0)
 			{
 				location.z += 25.f;
 			}
+
+			current_attempts++;
+
 			script::get_current()->yield();
 		}
 
-		location.z = ground_z;
+		if (current_attempts >= max_attempts)
+		{
+			return false;
+		}
+
+		location.z = ground_z + 1.f;
 		return true;
 	}
 
--- a/src/util/entity.cpp
+++ b/src/util/entity.cpp
@@ -193,37 +193,33 @@ namespace big::entity
 
 	bool load_ground_at_3dcoord(Vector3& location)
 	{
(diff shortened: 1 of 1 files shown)


src/util/entity.cpp
```diff
@@ -193,43 +193,37 @@ namespace big::entity
...
+		constexpr float max_ground_check	= 1000.f;
+		constexpr int max_attempts			= 300;
+		float ground_z						= location.z;
+		int current_attempts				= 0;
 
-		for (int i = 0; i < 10; i++)
+		while (!MISC::GET_GROUND_Z_FOR_3D_COORD(location.x, location.y, max_ground_check, &ground_z, FALSE, FALSE) && current_attempts != max_attempts)
 		{
-			for (int z = 0; z < 1000; z += 25)
+			if (MISC::GET_GROUND_Z_FOR_3D_COORD(location.x, location.y, max_ground_check, &ground_z, FALSE, FALSE))
```
Yimura: `MISC::GET_GROUND_Z_FOR_3D_COORD`, is getting called twice per frame.

I suggest the following changes:
```cpp
constexpr int max_attempts = 300;
bool found_ground = false;

int i = 0;
do {
    found_ground = MISC::GET_GROUND_Z_FOR_3D_COORD(location.x, location.y, max_ground_check, &ground_z, FALSE, FALSE);
    ++i;
} while(!found_ground && i < max_attempts);
```
It's an oversimplified example but at the least with this we prevent calling the same function twice on the same frame.

Yimura (changes requested):

src/util/entity.cpp
99Anvar99: Oh yeah, i have thought about that but forgot to make the change. Thank you for pointing out.

src/util/entity.cpp
```diff
@@ -193,43 +193,33 @@ namespace big::entity
...
-				float ground_iteration = static_cast<float>(z);
-				// Only request a collision after the first try failed because the location might already be loaded on first attempt.
-				if (i >= 1 && (z % 100 == 0))
-				{
-					STREAMING::REQUEST_COLLISION_AT_COORD(location.x, location.y, ground_iteration);
-					script::get_current()->yield();
-				}
+	    do {
+	        found_ground = MISC::GET_GROUND_Z_FOR_3D_COORD(location.x, location.y, max_ground_check, &ground_z, FALSE, FALSE);
+	        STREAMING::REQUEST_ADDITIONAL_COLLISION_AT_COORD(location.x, location.y, location.z);
```
gir489returns, line 204: Why did you change this out for REQUEST_ADDITIONAL_COLLISION_AT_COORD? The script only uses this in tandem with REQUEST_COLLISION_AT_COORD at the same exact coordinates.

src/util/entity.cpp
```diff
@@ -193,43 +193,33 @@ namespace big::entity
...
 
-				if (MISC::GET_GROUND_Z_FOR_3D_COORD(location.x, location.y, ground_iteration, &groundZ, false, false))
-				{
-					location.z = groundZ + 1.f;
-					done       = true;
-				}
-			}
+	        if (current_attempts % 10 == 0)
+	        {
+	            location.z += 25.f;
```
gir489returns, line 208: If my math is correct here, you're going to only top off here at 750 height units. Isn't the max height for GTA 800 or something?

src/util/entity.cpp
```diff
@@ -193,43 +193,33 @@ namespace big::entity
...
+	        }
 
-			float height;
-			if (done && WATER::GET_WATER_HEIGHT(location.x, location.y, location.z, &height))
-			{
-				location.z = height + 1.f;
-			}
+	        ++current_attempts;
```
gir489returns, line 211: There is no reason to use the ++ operator here before the variable, as nothing is accessing it on this line, although the compiler will probably just optimize this out.

gir489returns (changes requested):

src/util/entity.cpp
Yimura: Chiliad is around the ~800-1000 range, couldn’t say anything about the actual height.

src/util/entity.cpp
Yimura: ++i  is never slower than i++, whereas i++ may in some cases be slower than ++i. For this usecase it doesn’t really matter.

src/util/entity.cpp
gir489returns: When MBG first submitted his new teleport code, we had a user create a ticket about how he couldn't teleport to the top of the Maze Bank tower. IDK how tall it was, but IIRC it was 800 something units, and I had to adjust his code to fix it.

src/util/entity.cpp
99Anvar99: alright, switching back to REQUEST_COLLISION_AT_COORD

src/util/entity.cpp
99Anvar99: Well, i have tested it and it works just fine.
highest ground collision point is: 806
image

src/util/entity.cpp
99Anvar99: I guess this works out.

gir489returns (approved):
Looks great, tested well.

src/util/entity.cpp
gir489returns: > ++i is never slower than i++, whereas i++ may in some cases be slower than ++i. For this usecase it doesn’t really matter.

It usually is faster in a for loop, however, I'd imagine the opcode it generates is still the same. (INC register) but the compiler knows that since nothing is referring to it, it needs not care about order of operations.

Yimura (approved):

PR #2656, opened 2024-01-27
feat(lua): expose the self class and add new menu events

ShinyWasabi:
doc_gen.py needs to be fixed.

### DIFF
--- a/docs/lua/tables/menu_event.md
+++ b/docs/lua/tables/menu_event.md
@@ -2,7 +2,7 @@
 
 Table containing all possible events to which you can respond.
 
-## Fields (6)
+## Fields (8)
 
 ### `PlayerLeave`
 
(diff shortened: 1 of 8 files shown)


gir489returns:
MenuUnloaded will crash if you call a native within it and not gracefully exit. This should be noted in the documentation that we can't afford to wait for the next fiber to fire.

xiaoxiao921:
Well this could be changed to
```cpp
				// allow to unload in the main title screen.
				if (g_script_mgr.can_tick())
				{
					// empty the pool, we want the that job below run no matter what for clean up purposes.
					g_fiber_pool->reset();
					g_fiber_pool->queue_job([] {
					g_lua_manager->trigger_event<menu_event::MenuUnloaded>();
						for (auto& command : g_looped_commands)
							if (command->is_enabled())
								command->on_disable();

						g_running = false;
					});
				}
				else
				{
              				g_lua_manager->trigger_event<menu_event::MenuUnloaded>();
					g_running = false;
				}
```
				
If you are in the main menu title you could still crash trying to call natives I think? But you could have checks in your lua that could handle this case probably

Yimura (approved):

PR #2669, opened 2024-01-29
(Improvement) Request Model

99Anvar99:
- We don't want to request models that does not exist! or invalid

### DIFF
--- a/src/util/entity.cpp
+++ b/src/util/entity.cpp
@@ -234,6 +234,11 @@ namespace big::entity
 
 	bool request_model(rage::joaat_t hash)
 	{
+		if (STREAMING::HAS_MODEL_LOADED(hash))
+		{
+			return true;
+		}
+
 		bool has_loaded;
 
 		if (STREAMING::IS_MODEL_VALID(hash) && STREAMING::IS_MODEL_IN_CDIMAGE(hash))
@@ -249,11 +254,6 @@ namespace big::entity
 				script::get_current()->yield();
 			} while (!has_loaded);
 
-			return true;
-		}
-
-		if (STREAMING::HAS_MODEL_LOADED(hash))
-		{
 			return true;
 		}
 
--- a/src/util/entity.cpp
+++ b/src/util/entity.cpp
@@ -198,6 +198,7 @@ namespace big::entity
 	    float ground_z = location.z;
 	    int current_attempts = 0;
 	    bool found_ground;
+		float height;
 
 	    do {
 	        found_ground = MISC::GET_GROUND_Z_FOR_3D_COORD(location.x, location.y, max_ground_check, &ground_z, FALSE, FALSE);
@@ -218,7 +219,15 @@ namespace big::entity
 	        return false;
 	    }
 
-	    location.z = ground_z + 1.f;
+        if (WATER::GET_WATER_HEIGHT(location.x, location.y, location.z, &height))
+        {
+			location.z = height;
+        }
+		else
+		{
+			location.z = ground_z + 1.f;	
+		}
+
 	    return true;
 	}
 
--- a/src/util/entity.cpp
+++ b/src/util/entity.cpp
@@ -234,21 +234,25 @@ namespace big::entity
 
 	bool request_model(rage::joaat_t hash)
 	{
-		bool has_loaded = STREAMING::HAS_MODEL_LOADED(hash);
+		bool has_loaded;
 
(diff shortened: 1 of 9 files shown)


src/util/entity.cpp
```diff
@@ -218,7 +219,15 @@ namespace big::entity
 	        return false;
 	    }
 
-	    location.z = ground_z + 1.f;
+        if (WATER::GET_WATER_HEIGHT(location.x, location.y, location.z, &height))
+        {
+			location.z = height;
+        }
+		else
```
Yimura: Mixed indentation, run `clang-format -i src/**.hpp src/**.cpp` to fix.

src/util/ped.hpp
```diff
@@ -446,17 +446,20 @@ namespace big::ped
...
-		for (uint8_t i = 0; !STREAMING::HAS_MODEL_LOADED(hash) && i < 100; i++)
+		if (STREAMING::IS_MODEL_VALID(hash) && STREAMING::IS_MODEL_IN_CDIMAGE(hash) && !STREAMING::HAS_MODEL_LOADED(hash))
 		{
-			STREAMING::REQUEST_MODEL(hash);
-			script::get_current()->yield();
+			while (!STREAMING::HAS_MODEL_LOADED(hash))
+			{
+				STREAMING::REQUEST_MODEL(hash);
+				script::get_current()->yield();
+			}
```
Yimura: Since this piece of code is being duplicated quite a bit, how about making it a general purpose helper function?

Yimura (changes requested):
Just some mixed indentations here and there as well as duplication of commonly used code.

### To Do:
 - [x] Move the code block that checks if a model is valid and then loads it to a helper function
 - [x] Run clang-format to make the code that you submit uniform with the main code base.

src/util/ped.hpp
99Anvar99: Sure thing.

99Anvar99:
My bad, i meant to say created helper for request model function not class.

Edit: Everything seems to work fine!

src/util/entity.cpp
```diff
@@ -193,33 +193,67 @@ namespace big::entity
...
+
+		if (STREAMING::IS_MODEL_VALID(hash) && STREAMING::IS_MODEL_IN_CDIMAGE(hash) && !has_loaded)
+		{
+			while (!has_loaded)
+			{
+				if (has_loaded)
+					break;
+				STREAMING::REQUEST_MODEL(hash);
+				script::get_current()->yield();
+			}
```
Yimura: `has_loaded` is never updated in this loop, might cause a deadlock.

src/util/entity.cpp
```diff
@@ -193,33 +193,67 @@ namespace big::entity
...
+			return true;
+		}
+
+		if (has_loaded)
+		{
+			return true;
+		}
+
+		return false;
```
Yimura: Can be changed to `return has_loaded`.

src/util/world_model.hpp
```diff
@@ -15,26 +15,21 @@ namespace big::world_model
...
-		}
+			std::call_once(once_flag, setup_backup);
+			memset(g_pointers->m_gta.m_world_model_spawn_bypass, 0x90, patch_size);
 
-		std::call_once(once_flag, setup_backup);
-		memset(g_pointers->m_gta.m_world_model_spawn_bypass, 0x90, patch_size);
+			const auto object = OBJECT::CREATE_OBJECT(hash, location.x, location.y, location.z, is_networked, false, false);
 
-		const auto object = OBJECT::CREATE_OBJECT(hash, location.x, location.y, location.z, is_networked, false, false);
+			memcpy(g_pointers->m_gta.m_world_model_spawn_bypass, backup.data(), patch_size);
```
Yimura: I know this isn't your fault but could you update the patching of the `m_world_model_spawn_bypass` and make it into a `memory::byte_patch`?

Similarly to how it was implemented here:
https://github.com/YimMenu/YimMenu/blob/7efbee2e85ba2aee773e398cd9e66d54e640aca5/src/byte_patch_manager.cpp#L16

And here's an example usage:
https://github.com/YimMenu/YimMenu/blob/7efbee2e85ba2aee773e398cd9e66d54e640aca5/src/backend/looped/self/police.cpp#L19

Yimura (changes requested):

src/util/world_model.hpp
99Anvar99: Yes, i will update it.

Yimura (approved):
Thanks for another PR fixing and improving some of the tech debt in YimMenu!

PR #2677, opened 2024-01-31
Refractor(GET_FILENAME_FOR_AUDIO_CONVERSATION) To get_gxt_label

99Anvar99:
## Why?
- Better then using GET_FILENAME_FOR_AUDIO_CONVERSATION

## Additions And Changes
- Created a helper function for get_gxt_label
- Replaced all GET_FILENAME_FOR_AUDIO_CONVERSATION
- Clean ups (removed redundant code, fixed implicit conversions, fixed inf loop, using algorithms)

### DIFF
--- a/src/function_types.hpp
+++ b/src/function_types.hpp
@@ -205,6 +205,9 @@ namespace big::functions
 
 	using decal_manager_remove = void (*)(PVOID manager, rage::fwEntity*, DWORD a3, DWORD64 a4, DWORD ignore_bitset);
 
+	using get_gxt_label_from_table = const char* (*)(void* this_, const char* label);
+	using get_joaated_gxt_label_from_table = const char* (*)(void* this_, rage::joaat_t hash);
+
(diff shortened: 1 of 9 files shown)


Yimura:
Your **Why** section really doesn't explain anything. Because it's better?
> Better then using GET_FILENAME_FOR_AUDIO_CONVERSATION

My assumption is that `GET_FILENAME_FOR_AUDIO_CONVERSATION` requires TLS while a signature directly to the function that fetches the data does not require this.
However I'm wondering why all these extra signatures are required while we've got the following function which achieves the same with a single signature?
https://github.com/YimMenu/YimMenu/blob/ad1dfc71d2d5b2a9fb16f1c2009e4bc60bd65231/src/pointers.cpp#L489-L497

Also I'd add a function that takes a joaat value so you can use either in the code base as that will remove a bunch of strings from the final compilation of the menu.

99Anvar99:
Oh, that is true. I forgot that
- // Get Label Text 
Already exist.

Well then i could just change everything to that.

PR #2688, opened 2024-02-08
Block traffic protection

maybegreat48:


src/hooks/protections/received_event.cpp
```diff
@@ -564,6 +564,11 @@ namespace big
 				g_pointers->m_gta.m_send_event_ack(event_manager, source_player, target_player, event_index, event_handled_bitset);
 				return;
 			}
+			else if (type == WorldStateDataType::PopMultiplierArea && g.protections.stop_traffic) // never used for legitimate reasons
```
Yimura: Not even in races where traffic may end up being modified?

gir489returns: Races usually happen on areas that are like under the world like -8000 X -8000 Y or something like that IIRC.

SkiddyToast: This is used legitimately.
image

SkiddyToast: > Races usually happen on areas that are like under the world like -8000 X -8000 Y or something like that IIRC.

It modifies the entire map at times.
image

gir489returns: @SkiddyToast Do you know what circumstances the freemode script calls this function?

maybegreat48: @SkiddyToast good catch, I've assumed all of the native calls had the local parameter set to true. This one applies it across all machines
Freemode always calls it with localOnly enabled, so we don't have to worry about that

Yimura (approved):

PR #2695, opened 2024-02-09
feat(log): print if debug build

xiaoxiao921:


src/main.cpp
```diff
@@ -104,6 +104,10 @@ BOOL APIENTRY DllMain(HMODULE hmod, DWORD reason, PVOID)
 			    LOG(INFO) << "Yim's Menu Initializing";
 			    LOGF(INFO, "Git Info\n\tBranch:\t{}\n\tHash:\t{}\n\tDate:\t{}", version::GIT_BRANCH, version::GIT_SHA1, version::GIT_DATE);
 
+#ifndef NDEBUG
+			    LOG(INFO) << "Debug Build";
```
gir489returns, line 108: This should be changed to a warning or error, as the Debug build is extremely unstable.

gir489returns (changes requested):

Yimura (approved):

PR #2699, opened 2024-02-10
Fixed memory leak caused by not releasing the memory from GetWindowsVersion.

gir489returns:
BOTTOM TEXT


xiaoxiao921 (approved):
looks good to me

PR #2708, opened 2024-02-12
fixed bool_masked

Alice2333g:


Yimura (approved):

Yimura:
Anyone able to confirm this is working as intended?

gir489returns:
No, I don’t use this function.

Alice2333g:
> 有人能够确认这是否按预期工作？

it should 1 size.

it not 8 size.

if size 8 set to 1 bit, the next 7 bits will all be false.

gir489returns:
I don't understand why bit_count is statically declared. It should come from the user.

Yimura:
> I don't understand why bit_count is statically declared. It should come from the user.

No clue I have never used these functions so I wouldn't be able to confirm if this is the correct usage.
At the least this way the function will behave in an expected behaviour, we should still look at changing this or adding another function to toggle multiple bits.

xiaoxiao921:
> I don't understand why bit_count is statically declared. It should come from the user.

This is a helper function, looks like I wrote the 8 by mistake, it should be 1, so thanks for the PR

> > I don't understand why bit_count is statically declared. It should come from the user.
> 
> No clue I have never used these functions so I wouldn't be able to confirm if this is the correct usage. At the least this way the function will behave in an expected behaviour, we should still look at changing this or adding another function to toggle multiple bits.

Already provided, check the other helpers, like `get_masked_int_hash` (they can also call the native directly)

PR #2718, opened 2024-02-18
Added Korean font support.

gir489returns:
Closes https://github.com/YimMenu/Translations/pull/116


Yimura (approved):
Seems like a good enough fix for now, however I've been thinking about adding an additional field to the translation repo that specifies the fonts required. This may actually allow us to better optimize the font map (load the full Chinese font and switch between fonts only required for the current language).

We'll also have to take a look at how to properly reload the font map (without requiring an unload and inject).

Edit:
Forgot to add, "More tech debt, YAY!".

PR #2740, opened 2024-02-21
proper fix for weapon info damage crash

xiaoxiao921:


GreySideMM:
In this and #2733 and #2732 no notify::crash_blocked() is getting called. Should it be added or is it missing on purpose?

xiaoxiao921:
well I did add one in #2732, I just completly forgot to do the same here, ty for reminder

Yimura (approved):

PR #2745, opened 2024-02-21
lua: fix enabling/disabling

xiaoxiao921:


Yimura (approved):
Thanks for fixing the mess and managing to figure out how it works.

PR #2764, opened 2024-02-23
fix: ped body clean up logic

### LINKED ISSUE #2579: [Improvement]: remove clean loop and fix current clean logic
### Which feature

whenever player gets damaged it has blood decals, dirt, water & scars etc.
One can use clean button to clean player. It will not clean scars. Same is with clean loop bcz both use same function.

### remove clean loop
Clean loop purpose is to continuously clean player and so no need to manually open menu and hit the clean button. I think one can clean themselves by simply hitting clean button as necessary rather than running bunch of natives every frame.

### How to clean scars?
well one can add modify the clean function as -

```c++
	inline void clean_ped()
	{
		PED::CLEAR_PED_BLOOD_DAMAGE(ped);
		PED::CLEAR_PED_ENV_DIRT(ped);

                // https://forum.cfx.re/t/information-needed-to-clear-visible-player-damage-scars-etc/283216 
 		PED::CLEAR_PED_DAMAGE_DECAL_BY_ZONE(ped, 0, "ALL");
 		PED::CLEAR_PED_DAMAGE_DECAL_BY_ZONE(ped, 1, "ALL");
 		PED::CLEAR_PED_DAMAGE_DECAL_BY_ZONE(ped, 2, "ALL");
 		PED::CLEAR_PED_DAMAGE_DECAL_BY_ZONE(ped, 3, "ALL");
 		PED::CLEAR_PED_DAMAGE_DECAL_BY_ZONE(ped, 4, "ALL");
 		PED::CLEAR_PED_DAMAGE_DECAL_BY_ZONE(ped, 5, "ALL");
	}
```

i also removed `PED::CLEAR_PED_WETNESS(player_ped);` though its a decision one can make. It just removes water when you are wet.

### Reason

- clean loop is not such required. cleaning can be done as required.
- scars did not cleaned up. The above code guarantees one's beauty back.

### Additional context

_No response_

lonelybud:
fixes #2579

### DIFF
--- a/src/util/entity.cpp
+++ b/src/util/entity.cpp
@@ -16,15 +16,11 @@ namespace big::entity
 		PED::CLEAR_PED_WETNESS(ped);
 		PED::CLEAR_PED_ENV_DIRT(ped);
 		PED::RESET_PED_VISIBLE_DAMAGE(ped);
-			
+
 		// https://forum.cfx.re/t/information-needed-to-clear-visible-player-damage-scars-etc/283216
 		// https://docs.fivem.net/natives/?_0x397C38AA7B4A5F83
-		PED::CLEAR_PED_DAMAGE_DECAL_BY_ZONE(ped, 0, "ALL");
-		PED::CLEAR_PED_DAMAGE_DECAL_BY_ZONE(ped, 1, "ALL");
-		PED::CLEAR_PED_DAMAGE_DECAL_BY_ZONE(ped, 2, "ALL");
-		PED::CLEAR_PED_DAMAGE_DECAL_BY_ZONE(ped, 3, "ALL");
-		PED::CLEAR_PED_DAMAGE_DECAL_BY_ZONE(ped, 4, "ALL");
-		PED::CLEAR_PED_DAMAGE_DECAL_BY_ZONE(ped, 5, "ALL");
+		for (int i = 0; i <= 5; ++i)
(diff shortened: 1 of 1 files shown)


src/util/entity.cpp
```diff
@@ -12,12 +12,19 @@ namespace big::entity
...
-		PED::CLEAR_PED_ENV_DIRT(player_ped);
-		PED::RESET_PED_VISIBLE_DAMAGE(player_ped);
+		PED::CLEAR_PED_BLOOD_DAMAGE(ped);
+		PED::CLEAR_PED_WETNESS(ped);
+		PED::CLEAR_PED_ENV_DIRT(ped);
+		PED::RESET_PED_VISIBLE_DAMAGE(ped);
+			
+		// https://forum.cfx.re/t/information-needed-to-clear-visible-player-damage-scars-etc/283216
+		// https://docs.fivem.net/natives/?_0x397C38AA7B4A5F83
+		PED::CLEAR_PED_DAMAGE_DECAL_BY_ZONE(ped, 0, "ALL");
```
xiaoxiao921: Use a loop

xiaoxiao921 (changes requested):

src/util/entity.cpp
lonelybud: done

xiaoxiao921 (approved):

PR #2766, opened 2024-02-23
enhance(view_players): add search bar

### LINKED ISSUE #2562: [Improvement]: add search bar to players list & reactions
### Which feature

players list eg:
https://github.com/lonelybud/lonelyMenu/blob/a5af2840ad40e11439db59873dbd2b05d05ba315/src/views/players/view_players.cpp#L120-L151

image

reactions list eg:
https://github.com/lonelybud/lonelyMenu/blob/master/src/views/settings/view_reaction_settings.cpp

image

### Reason

improve ux

### Additional context

_No response_

lonelybud:
partial fixes #2562


src/views/players/view_players.cpp
```diff
@@ -119,25 +119,47 @@ namespace big
...
 			has_scrollbar = window_height + window_pos > (float)*g_pointers->m_gta.m_resolution_y - 10.f;
 
-			// basically whichever is smaller, the max available screenspace or the calculated window_height
-			window_height = has_scrollbar ? (float)*g_pointers->m_gta.m_resolution_y - (window_pos + 40.f) : window_height;
-
 			ImGui::PushStyleColor(ImGuiCol_FrameBg, {0.f, 0.f, 0.f, 0.f});
 			ImGui::PushStyleColor(ImGuiCol_ScrollbarBg, {0.f, 0.f, 0.f, 0.f});
 
-			if (ImGui::BeginListBox("##players", {ImGui::GetWindowSize().x - ImGui::GetStyle().WindowPadding.x * 2, window_height}))
+			auto width_of_player_list   = ImGui::GetWindowSize().x - ImGui::GetStyle().WindowPadding.x * 2;
```
gir489returns, line 125: Why are there spaces after members? That is very confusing.

src/views/players/view_players.cpp
```diff
@@ -119,25 +119,47 @@ namespace big
...
+
+			static std::string search_player_name;
+			ImGui::SetNextItemWidth(width_of_player_list);
+			if (components::input_text_with_hint("###search_player_name", "search name", search_player_name))
```
gir489returns, line 129: We no longer allow static text, this must be added to the translation dictionary located here: https://github.com/YimMenu/Translations

src/views/players/view_players.cpp
```diff
@@ -119,25 +119,47 @@ namespace big
...
+							if (lower_case_name.find(search_player_name) != std::string::npos)
+								player_button(player);
+						}
+						else
+							player_button(player);
+				}
 
 				ImGui::EndListBox();
 			}
+
```
gir489returns, line 162: Unnecessary whitespace addition.

src/views/players/view_players.cpp
lonelybud: done

src/views/players/view_players.cpp
lonelybud: can you check the new commit should fix it?

src/views/players/view_players.cpp
lonelybud: indendation to give code claridty. This separates the complete if block from the code below it

lonelybud:
please test it.

src/views/players/view_players.cpp
```diff
@@ -100,44 +100,65 @@ namespace big
...
+			// render search
+			static std::string search_player_name;
+			ImGui::SetNextItemWidth(window_width);
+			if (components::input_text_with_hint("###search_player_name", "SEARCH"_T, search_player_name))
+				std::transform(search_player_name.begin(), search_player_name.end(), search_player_name.begin(), ::tolower);
+			auto search_height = ImGui::GetItemRectSize().y; // Get the size of the last element
+			//
+
+			float raw_height = (ImGui::CalcTextSize("A").y + style.FramePadding.y * 2.0f + style.ItemSpacing.y) // button size
```
gir489returns, line 129: ImGui::CalcTextSize("A").y + style.FramePadding.y * 2.0f + style.ItemSpacing.y;

This could be factored out into a static const since it is not likely to change at runtime after calculation.

```
static const auto button_size = ImGui::CalcTextSize("A").y + style.FramePadding.y * 2.0f + style.ItemSpacing.y;
const auto raw_height = button_size * player_count + (player_count > 1) * ((style.ItemSpacing.y * 2) + 1.f) + (player_count == 1) * 2.f;
has_scrollbar = raw_height + window_pos > (float)*g_pointers->m_gta.m_resolution_y - 10.f;
const auto window_height = has_scrollbar ? (float)*g_pointers->m_gta.m_resolution_y - (window_pos + search_height + 40.f) : raw_height;
```

Not sure if it would actually speed up the calculations or not, but you also want to remember, the C++ compiler enforces PEMDAS. Splitting out things that need to be calculated before hand is the easiest way to tell the compiler which can be optimized. 

It also makes it more readable instead of one giant-ass block of code.

gir489returns (changes requested):

src/views/players/view_players.cpp
lonelybud: i am gonna look into this

lonelybud: I believe -

`m_resolution_y`  can change, `window_pos` can change when you popup different menus in nav menu. `ItemSpacing, CalcTextSize, FramePadding` and infact everything changes when you rescale UI.

lonelybud: we can remove lot of code by assuming fix height of container that is `(float)*g_pointers->m_gta.m_resolution_y - (window_pos + search_height + 40.f)` and width related code as well.

lonelybud: i wish there were stuffs like maxHeight & flex containers in imgui but it is not. We need to create those from scratch i guess.

lonelybud: also `window_pos` can also be determined simply if  `view::heading()`, `view::navigation()` & `view::players()` are part of same group.

lonelybud: i thought of doing this what you said and i did something. It is true one can calculate the size and determine it again when UI is rescaled. I will share the changes. It also fixed incorrect UI changes which exist currently

lonelybud: https://github.com/lonelybud/lonelyMenu/commit/f2c062541b36417c05a09893dbb022266fb4094a#diff-cea3cc35a3bc3e5a06c30d7eedb274e4e564ddc8ee39997a221e515d19931493

the code is weird in the sense that it has some code quality and dependency issues like calling `update_gui_info` whenever we need to update the gui variables. So it will update the gui when menu is toggled, players join/leaves/reset/click, navigation change. Not a desired solution but still you can find some part interesting,

lonelybud:
I think there is possibly to much to add. I shall close this pr. Also note that `ImGui::GetStyle()` gets the style of last element.

PR #2772, opened 2024-02-24
Fix: display of the vehicle list for Traditional Chinese users

sch-lda:
Adding the 279 additional characters for Traditional Chinese user's vehicle list.
Since vehicle list is retrieved from the game and the Traditional Chinese translation is incomplete , it is necessary for Traditional Chinese users to be able to display both Simplified and Traditional Chinese.


Yimura (dismissed):

sch-lda:
I have supplemented some additional Traditional Chinese characters based on the tester's request. If you discover any other necessary but missing Traditional Chinese characters, you can modify the tw_accumulative_offsets_from_0x4E00 array using the following tool: https://github.com/sch-lda/Imgui-unicode-offset-array-generator .

Yimura (approved):

PR #2785, opened 2024-02-27
lua natives: try to handle Any* parameters properly

### LINKED ISSUE #2774: [Bug]: Game crashes when calling natives with 'Any*' parameter
### Describe the bug

Calling a Lua native with a parameter of type `Any*` results in a game crash.

### Steps To Reproduce

1. Try to call a native with an `Any*` parameter in a Lua script.

### Expected Behavior

No crash

### Logs and/or screenshots

[cout.log](https://github.com/YimMenu/YimMenu/files/14393332/cout.log)

### Store

Steam

### Language

English

### YimMenu Language

English (US)

### Additional context

This is my test code:
```lua
function GET_INITIAL_COORD_FROM_UGC(file, dict1, dict2)
    if not DATAFILE.DATAFILE_LOAD_OFFLINE_UGC(file, 0) then
        log.warning("Could not load UGC " .. file)
        return RE.VECTOR_ZERO
    end

    local uVar0 = DATAFILE.DATAFILE_GET_FILE_DICT(0)
    local uVar1 = DATAFILE.DATADICT_GET_DICT(uVar0, "mission")
    local iVar29 = DATAFILE.DATADICT_GET_DICT(uVar1, dict1)
    local iVar30 = DATAFILE.DATADICT_GET_DICT(iVar29, dict2)
    local uVar37 = DATAFILE.DATADICT_GET_ARRAY(iVar30, "pos")
    local coord = DATAFILE.DATAARRAY_GET_VECTOR(uVar37, 0)

    DATAFILE.DATAFILE_DELETE(0)

    return coord
end
-- ...
if ImGui.Button("TEST") then
    script.run_in_fiber(function()
        local coords = GET_INITIAL_COORD_FROM_UGC("RE_Store_Robbery_1", "pd", "pds")
        log.info(string.format("Val: %.3f, %.3f, %.3f", coords.x, coords.y, coords.z))
    end)
end
```

I have also tested non-datafile natives such as `GET_ALL_VEHICLES(Any* vehsStruct)` etc., they resulted the same behaviour.

### Build

### Coload

xiaoxiao921:
Fixes #2774

Need testing!

### DIFF
--- a/scripts/natives_gen.py
+++ b/scripts/natives_gen.py
@@ -24,13 +24,17 @@ def print_hpp(text):
 class Arg:
     def __init__(self, name, type_):
         self.name = name
-        self.type_ = type_.replace("BOOL", "bool")
+        self.type_ = type_.replace("BOOL", "bool").replace("Any*", "uintptr_t")
+        if self.type_ == "uintptr_t":
+            self.is_any_ptr = True
+        else:
+            self.is_any_ptr = False
         if self.type_ == "const char*":
             self.type_ = self.type_.replace("const char*", "sol::stack_object")
             self.is_string = True
         else:
             self.is_string = False
-        self.is_pointer_arg = "*" in type_ and "const char" not in type_
+        self.is_pointer_arg = "*" in self.type_ and "const char" not in self.type_
         self.type_no_star = self.type_.replace("*", "")
 
     def __str__(self) -> str:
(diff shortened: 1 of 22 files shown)


ShinyWasabi:
I have just tested it with my test scenario in the linked issue and encountered no problems.

gir489returns (approved):
I'm sure things like WEAPON::GET_WEAPON_COMPONENT_HUD_STATS won't work because it's writing to a struct, and IDK if the Sol <-> Lua table thing works if you use a pointer, but at this time, I really don't care, since we should instead of using that, expose the weapons.json information we have via Lua.

xiaoxiao921:
So what can be done for those is use the memory lua table to allocate the right amount of memory so that the native can properly write to it, but yeah none of those structs (or almost most of them) are publicly reversed so that's why it is fine as it is. Because most of those natives are not used in the first place anyway

PR #2806, opened 2024-03-07
refactor!: switch RAGE_JOAAT with string literal functions

Yimura:
Why? Shorter to write and removes the macro usage

I used the following regex to find all occurrences:
```d
RAGE_JOAAT\("(.*?)"\)
```
then the following to replace it all:
```d
"$1"_J
```

I'm open to suggestions, might be a stupid change.


xiaoxiao921 (approved):

gir489returns:
I think it should be R_J to signify what it’s for (RAGE JOAAT) or just RJ for short. But in programming, the handling functor always comes before the item not after. Seems awful to have it come after.

Yimura:
> I think it should be R_J to signify what it’s for (RAGE JOAAT) or just RJ for short. But in programming, the handling functor always comes before the item not after. Seems awful to have it come after.

Can't really do much about the position, user defined literals have to start with `_` and be at the end of the string.
`""_RJ` would be possible though.

https://en.cppreference.com/w/cpp/language/user_literal

PR #2807, opened 2024-03-07
Chore[Translation]: Simplified zhcn Character List and Add new Traditional Chinese characters for https://github.com/YimMenu/Translations/pull/121

sch-lda:
1.Remove Simplified Chinese Level 3 Character List because most of the characters in it are specific to special place names, personal names, or are commonly used only in Traditional Chinese. 
2.Export Chinese characters from GTA5 rpf files 、zh_TW.json,  filter out traditional Chinese parts that Yimmenu has not loaded, and add them to Yimmenu so that Yimmenu can display vehicle/weapon/vehicle modification lists.

If new characters are added to the traditional Chinese translation of Yimmenu in the future, you can follow this [instruction](https://github.com/sch-lda/Imgui-unicode-offset-array-generator/blob/main/README.md) to add them.


gir489returns (approved):
I have no idea the purpose or usecase for this, but since you actually use the Chinese version of the game and I don't, I'm gonna approve this anyway.

sch-lda:
The goal is simple: to reduce unnecessary Simplified Chinese characters while ensuring that the new Traditional Chinese translation can be displayed correctly without increasing VRAM usage. 
Occasionally, there are still reports from Chinese users of menu not being displayed, so I dare not directly add new Traditional Chinese characters to avoid additional VRAM usage.

PR #2844, opened 2024-03-18
Fixed hooks::scripted_game_event using incorrect types and sizes.

gir489returns:


DayibBaba:
I don't think this fixed logging on the console. Looks to be exclusive to lua. I think we need to change it [here](https://github.com/YimMenu/YimMenu/blob/d0c5ffaa403f2ab35d0b3bc24aad0b659cd459fb/src/hooks/protections/script_event_handler.cpp#L435) too.

DayibBaba (approved):

Yimura (approved):

PR #2857, opened 2024-03-21
Add more crash models to protections

ghost:
Some skiddy menus can still crash YimMenu with the good old fashioned model crashing method. I have added some more models to the protection list.


Yimura (approved):

Yimura:
@xynnylol any false positives you observed with these being added?

ghost:
> @xynnylol any false positives you observed with these being added?

Not from what i can tell these are mostly from me being an idiot and spawning every model i can find in the game :joy:

PR #2864, opened 2024-03-22
Refactor and fix send chat

### LINKED ISSUE #1910: [Bug]: CEO/MC [ORGANIZATION] chat is visible for all users
### Describe the bug

CEO/MC [ORGANIZATION] / [TEAM] chat is visible for all users. Several players reported that they can see my chat as their Private chat. I did some experiments with removing any LUA and reloading then to disable em all but they still could see that.

### Steps To Reproduce

1. Have a friend online
2. Type in Team chat while your friend is not a member of your CEO/MC
3. So friend should see all your chat secrets

### Expected Behavior

Expected to stay private with CEO/MC chat

### Logs and/or screenshots

_No response_

### Store

Rockstar

### Additional context

After additional testing Im confirm: with Yim youre unwillingly share private chat over all lobby players.

### Build

### Coload

### LINKED ISSUE #2783: [Request]: sending chat messages in lua
### Problem

network.send_chat_message(msg, bool) uses true for team chat, false for all chat

requesting bool to be an int range 1, 2, 3 or string ("all", "team", "org") 

so it can be used for the entire organization youre in as well.

### Solution

problem sums it up, no rush, would just be nice to add.

### Reason

it already uses all/team, why not?

### Additional context

_No response_

### LINKED ISSUE #2861: [Bug]: Forged chat messages aren't seen by other players
### Describe the bug

When an authorized player makes use of the chat commands, any automatic response via the chat from our client is purely local. Other players don't see it.

### Steps To Reproduce

1. Enable chat commands
2. Allow someone access to chat commands
3. Let them input a wrong command or argument
4. Ask whether they see the response
5. They Don't.

### Expected Behavior

The messages should be networked and visible to the lobby and also restricted to the channel the command is received from (all or team).

### Logs and/or screenshots

```
<paste your logs here>
```

### Store

Steam

### Language

English

### YimMenu Language

English (US)

### Additional context

_No response_

### Build

### Coload

maybegreat48:
* Refactors the chat send method
* Adds a lua function to send private chat messages (TODO implement in core YimMenu?)
* Team chats are now only sent to players on your team

Fixes #2861
Fixes #2783 
Fixes #1910


src/views/network/view_network.cpp
```diff
@@ -402,7 +398,7 @@ namespace big
 					    toxic::start_activity(plyr.second, eActivityType::GunrunningDefend);
 				    });
 			    });
-			    ImGui::SeparatorText("Bounty");
+				ImGui::SeparatorText("Bounty");
```
Yimura, line 401: Random indent

Yimura (changes requested):
All looks good aside from a random indent I found.

Yimura:
Testing is required as @maybegreat48 has a broken GTA installation currently.

src/backend/reactions/interloper_reaction.cpp
```diff
@@ -36,15 +36,8 @@ namespace big
...
-			g_fiber_pool->queue_job([attacker, victim, this] {
-				auto chat = std::format("{} {}", g.session.chat_output_prefix, g_translation_service.get_translation(m_announce_message));
-
-				if (g_hooking->get_original<hooks::send_chat_message>()(*g_pointers->m_gta.m_send_chat_ptr,
-				        g_player_service->get_self()->get_net_data(),
-				        chat.data(),
-				        is_team_only))
-					notify::draw_chat(chat.c_str(), g_player_service->get_self()->get_name(), is_team_only);
-			});
+			auto chat = std::format("{} {}", g.session.chat_output_prefix, g_translation_service.get_translation(m_announce_message));
```
gir489returns, line 39: This variable name makes no sense. It should be message or output or something.

src/backend/reactions/interloper_reaction.cpp
```diff
@@ -36,15 +36,8 @@ namespace big
...
+			chat::send_message(chat);
```
gir489returns, line 40: Because here, lexically, it reads:

Chat interface. Send message: chat.

src/backend/reactions/reaction.cpp
```diff
@@ -62,15 +62,11 @@ namespace big
...
+			    g.session.chat_output_prefix,
+			    std::vformat(g_translation_service.get_translation(m_announce_message), std::make_format_args(player->get_name())));
 
-				if (g_hooking->get_original<hooks::send_chat_message>()(*g_pointers->m_gta.m_send_chat_ptr,
-				        g_player_service->get_self()->get_net_data(),
-				        chat.data(),
-				        is_team_only))
-					notify::draw_chat(chat.c_str(), g_player_service->get_self()->get_name(), is_team_only);
-			});
+			chat::send_message(chat);
```
gir489returns, line 69: Ditto.

gir489returns (dismissed):
Pedantic changes. Not required.

gir489returns (approved):
Looks good from a code standpoint. No idea if it actually works.

src/views/network/view_network.cpp
maybegreat48: The code was originally indented using spaces but the file is supposed to use tab indents

DayibBaba:
> Looks good from a code standpoint. No idea if it actually works.

Nope. Crashes the instant a message is supposed to be sent such as an invalid vehicle model or invalid command.

Tayipc38 (changes requested):
Does not work. Crashes the instant a message is sent. Log does not show anything of value.
[cout.log](https://github.com/YimMenu/YimMenu/files/14729409/cout.log)

gir489returns:
> Does not work. Crashes the instant a message is sent. Log does not show anything of value. [cout.log](https://github.com/YimMenu/YimMenu/files/14729409/cout.log)

It seems to be crashing in chat.hpp line 197, probably because target is null. There is a bang check on it in the previous block. Should probably move it to the head of the function and just return if target is null.

GreySideMM:
> It seems to be crashing in chat.hpp line 197, probably because target is null. There is a bang check on it in the previous block. Should probably move it to the head of the function and just return if target is null.

Also further above: https://github.com/YimMenu/YimMenu/pull/2864/commits/879146cca642053b6549afe2fdd326e3f9edf654#diff-6805e0661392393a2391f0e1916f05d8b7c617a5d66960ce2a9e87deb3803a7aR191
As I understand it if target is set the function only should send a message to this specific player instead everyone. Not entirely sure how he wanted to handle is_team in that case. However:
Replacing `PLAYER::GET_PLAYER_TEAM(target->m_player_id)`
with `PLAYER::GET_PLAYER_TEAM(player.second->get_net_game_player()->m_player_id)`
should help there.
The lines 197 and also 200 still need attention. Not sure when that is supposed to be executed. Maybe a `if (draw && target)` ?

Tayipc38:
> > It seems to be crashing in chat.hpp line 197, probably because target is null. There is a bang check on it in the previous block. Should probably move it to the head of the function and just return if target is null.
> 
> Also further above: [879146c#diff-6805e0661392393a2391f0e1916f05d8b7c617a5d66960ce2a9e87deb3803a7aR191](https://github.com/YimMenu/YimMenu/commit/879146cca642053b6549afe2fdd326e3f9edf654#diff-6805e0661392393a2391f0e1916f05d8b7c617a5d66960ce2a9e87deb3803a7aR191) As I understand it if target is set the function only should send a message to this specific player instead everyone. Not entirely sure how he wanted to handle is_team in that case. However: Replacing `PLAYER::GET_PLAYER_TEAM(target->m_player_id)` with `PLAYER::GET_PLAYER_TEAM(player.second->get_net_game_player()->m_player_id)` should help there. The lines 197 and also 200 still need attention. Not sure when that is supposed to be executed. Maybe a `if (draw && target)` ?

We can make the netMessage send to simply a single person by hooking the sender and filtering on player message id. If i remember correctly, the message is repeated for each person in the lobby, so just return without calling original once the hook tries the wrong msg_id.

lonelybud:
> It seems to be crashing in chat.hpp line 197, probably because target is null. There is a bang check on it in the previous block. Should probably move it to the head of the function and just return if target is null.

why we draw chat with target's name shouldn't be it sender's name?

lonelybud:
> Also further above 

correct it should be `(!is_team || PLAYER::GET_PLAYER_TEAM(player.second->id()) == PLAYER::GET_PLAYER_TEAM(self::id)))`

src/util/chat.hpp
```diff
@@ -125,4 +145,59 @@ namespace big::spam
...
+
+		GRAPHICS::DRAW_SCALEFORM_MOVIE_FULLSCREEN(scaleform, 255, 255, 255, 255, 0);
+
+		// fix broken scaleforms, when chat alrdy opened
+		if (const auto chat_data = *g_pointers->m_gta.m_chat_data; chat_data && (chat_data->m_chat_open || chat_data->m_timer_two))
+			HUD::CLOSE_MP_TEXT_CHAT();
+	}
+
+	// set target to send to a specific player
+	inline void send_message(const std::string& message, CNetGamePlayer* target = nullptr, bool draw = true, bool is_team = false)
```
lonelybud: could use `player_ptr` instead of `CNetGamePlayer`.

src/util/chat.hpp
```diff
@@ -125,4 +145,59 @@ namespace big::spam
...
+				if (player.second && player.second->is_valid() 
+					&& (!target || target == player.second->get_net_game_player()) 
+					&& (!is_team || PLAYER::GET_PLAYER_TEAM(player.second->get_net_game_player()->m_player_id) == PLAYER::GET_PLAYER_TEAM(self::id)))
+					msg.send(player.second->get_net_game_player()->m_msg_id);
+
+		if (draw)
+			if (rage::tlsContext::get()->m_is_script_thread_active)
+				draw_chat(message.c_str(), g_player_service->get_self()->get_name(), is_team);
+			else
+				g_fiber_pool->queue_job([message, target, is_team] {
```
lonelybud, line 236: remove `target` from dependency list

src/lua/bindings/network.cpp
```diff
@@ -183,13 +184,22 @@ namespace lua::network
...
+	// Table: network
+	// Name: send_chat_message_to_player
+	// Param: player_idx: integer: Index of the player.
+	// Param: msg: string: Message to be sent.
+	// Sends a chat message to the specified player. Other players would not be able to see the message
+	static void send_chat_message_to_player(int player_idx, const std::string& msg)
+	{
+		if (auto player = big::g_player_service->get_by_id(player_idx))
+		{
+			big::chat::send_message(msg, player->get_net_game_player());
```
lonelybud: could use `player_ptr` instead of `CNetGamePlayer`

lonelybud (commented):
these are little concerns. you are free to ignore

Tayipc38 (approved):
@Yimura We have decided to ignore the games logic for determining recipients. @maybegreat48 has added a function to determine who gets a chat message by our own logic. Should not differ much from how it originally works, barring any minor anomalies undiscovered as of now.

gir489returns (approved):

Yimura (approved):

PR #2874, opened 2024-03-26
Fixed send chat logic

Tayipc38:
Fixed an error that sent chats to wrong recipients. Also added player specific report output to chat_command_context.

### DIFF
--- a/src/util/chat.hpp
+++ b/src/util/chat.hpp
@@ -216,18 +216,29 @@ namespace big::chat
 	// set target to send to a specific player
 	inline void send_message(const std::string& message, player_ptr target = nullptr, bool draw = true, bool is_team = false)
 	{
+		if (!*g_pointers->m_gta.m_is_session_started)
+			return;
+
 		packet msg{};
 		msg.write_message(rage::eNetMessage::MsgTextMessage);
 		msg.m_buffer.WriteString(message.c_str(), 256);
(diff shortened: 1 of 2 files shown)


src/util/chat.hpp
```diff
@@ -216,18 +216,29 @@ namespace big::chat
...
+
+		for (auto& player : g_player_service->players())
+		{
+			if (player.second && player.second->is_valid())
+			{
+				if (target && player.second != target)
+					continue;
+
+				if (!target && is_team && !is_on_same_team(player.second->get_net_game_player()))
+					continue;
```
Yimura, line 237: With how the arguments to `chat::send_message` are setup the menu might not send a message to the player back unless they're in the same team. Providing no feedback to people that aren't in the team but still authorized to execute commands.

Unless I'm incorrect though.

Yimura (approved):

Yimura (dismissed):
See incorrect approval review.

src/util/chat.hpp
Tayipc38: This code is tested mind you. I don't fully understand what you mean. Ill try and explain the logic here. 
If there is no target defined, but is_team is true, the message will be sent to all team members. If target is defined, both checks do not matter since we have only 1 recipient. If both is_team and target are false/undefined, the logic filters no one.

Yimura (approved):

PR #2879, opened 2024-03-26
ACTUALLY fixes #2675 this time I swear.

### LINKED ISSUE #2675: [Bug]: Vehicle Preview glitches out and actually spawns the vehicles
### Describe the bug

Only / at least at one certain spot on the map the preview glitches out. Instead of seeing the cars nothing shows up. The cars are spawned apparently underneath the map. Only after like 10-20sec suddenly usually dozens of vehicles start showing up. Apparently being teleported to the surface again after they fell into the void beneath the map. The vehicles are spawned as persistent, they will remain there forever. They also are networked.
Closely related to #2077.

### Steps To Reproduce

1. Create a new or join a existing session. (New one seems to be easier to reproduce.)
2. Teleport to the LS Customs south of the airfield at Sandy Shores.
3. Be on the street, face towards LSC. (The rough area should be good enough.)
4. You have to spawn at least a car once. Maybe do this several times.
5. Keep moving the mouse above multiple entries in the vehicle list to trigger a new/different preview.
6. You should now notice that no longer previews appear. Wait up to half a minute or try again after walking some meters and restart at 4.
7. Enjoy the piling up of docens of cars.

### Expected Behavior

No spawned cars. No networking of these cars.

### Logs and/or screenshots

```
<paste your logs here>
```

### Store

Steam

### Language

English

### YimMenu Language

English (US)

### Additional context

_No response_

### Build

### Coload

gir489returns:
PINKY SWEAR

### DIFF
--- a/src/services/model_preview/model_preview_service.cpp
+++ b/src/services/model_preview/model_preview_service.cpp
@@ -204,12 +204,12 @@ namespace big
 					ENTITY::SET_ENTITY_COORDS(m_current_ent, location.x, location.y, location.z, 0, 0, 0, 0);
 				}
 
-				if (m_heading += 0.5f; m_heading > 359)
+				if (m_heading += 0.11111f; m_heading > 359)
 				{
 					m_heading = 0;
 				}
 
-				script::get_current()->yield(15ms);
+				script::get_current()->yield();
 			}
 
 			entity::delete_entity(m_current_ent, true);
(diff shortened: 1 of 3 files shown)


src/services/model_preview/model_preview_service.cpp
```diff
@@ -204,12 +204,12 @@ namespace big
...
 				}
 
-				if (m_heading += 0.5f; m_heading > 359)
+				if (m_heading += 0.11111f; m_heading > 359)
 				{
 					m_heading = 0;
 				}
 
-				script::get_current()->yield(15ms);
+				script::get_current()->yield();
```
Yimura, line 212: We really need to change this to time based intervals, not something that is variable/depending on the framerate of a user's PC.

Yimura (approved):

PR #2882, opened 2024-03-27
Fix chat channel (All/Team/Organization) on chat receive

### LINKED ISSUE #2878: [Bug]: Chat now prints everything as [team]
### Describe the bug

Must be due to #2864 or previous commits. Whenever someone writes something for "all" it is being displayed as "team".
There was a duplicated is_team Read<bool>() removed here: https://github.com/YimMenu/YimMenu/commit/7e7bcb155a5594a6e5955894cc2b8f05d063b31c#diff-12a073d544e519b46ab3e4f1eb08c2098f554ecbcf7aa72e61d776fc2cd9ca63L150
So that apparently should have never been there and might even had caused unnoticed misbehavior before. Also: In the chat.log everything is correct.

So it seems to be what I suspected: There is a bug with buffer.Read<bool>(1) I encountered before. Not sure about every details anymore:
While is_team logs as "0" to the console it actually has the uint8_t value of "6" (e.g. 0b0110). IIRC when passed as boolean to a function it "correctly" interprets this "6" as false (probably only checks the lowest bit). I suspect this to not work with the call of `g_pointers->m_gta.m_handle_chat_message()`.
Treating is_team as an int and then check for == 1 is a stupid workaround I used and seems to work. Everything else might be killed by the compilers optimization. (I used similar in some WIP before.)
```
if ((int)is_team == 1)
	is_team = true;
else
	is_team = false;
```

### Steps To Reproduce

-

### Expected Behavior

-

### Logs and/or screenshots

```
<paste your logs here>
```

### Store

Steam

### Language

English

### YimMenu Language

English (US)

### Additional context

_No response_

### Build

### Coload

Tayipc38:
This simply executes all our operations on the chat message before setting it back and returning the original function. This way we won't have to handle the actual message.

Fixes #2878

### DIFF
--- a/src/hooks/protections/receive_net_message.cpp
+++ b/src/hooks/protections/receive_net_message.cpp
@@ -107,8 +107,10 @@ namespace big
 			case rage::eNetMessage::MsgTextMessage2:
 			{
 				char message[256];
+				rage::rlGamerHandle temp{};
+				bool is_team = false;
 				buffer.ReadString(message, 256);
-				bool is_team;
+				gamer_handle_deserialize(temp, buffer);
 				buffer.ReadBool(&is_team);
 
 				if (player->is_spammer)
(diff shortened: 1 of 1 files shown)


lonelybud:
correct there is no need to handle the normal message, it should be handled by the game itself. I have also done similar - 
https://github.com/lonelybud/lonelyMenu/blob/24eeeae1344a0b87521a396fa3626d29f2262868/src/hooks/protections/receive_net_message.cpp#L176-L208

gir489returns:
I was just about to submit a PR to do the same thing.

src/hooks/protections/receive_net_message.cpp
```diff
@@ -107,8 +107,10 @@ namespace big
 			case rage::eNetMessage::MsgTextMessage2:
 			{
 				char message[256];
+				rage::rlGamerHandle temp{};
+				bool is_team = false;
 				buffer.ReadString(message, 256);
-				bool is_team;
+				gamer_handle_deserialize(temp, buffer);
```
gir489returns, line 113: gamer_handle_deserialize uses ReadInt64 which is incorrect. There is different logic for parsing a signed int from a bitbuffer. It should instead use ReadQWord.

gir489returns (changes requested):

Yimura:
Invalidated by #2883

PR #2883, opened 2024-03-27
Unify rlGamerHandle code and fix chat inaccuracies.

### LINKED ISSUE #2878: [Bug]: Chat now prints everything as [team]

gir489returns:
I removed a lot of derelict code and also unified the serialize/deserializing process of rlGamerHandle.

Invalidates #2882
Closes #2878

### DIFF
--- a/cmake/gtav-classes.cmake
+++ b/cmake/gtav-classes.cmake
@@ -2,8 +2,8 @@ include(FetchContent)
 
 FetchContent_Declare(
     gtav_classes
-    GIT_REPOSITORY https://github.com/Yimura/GTAV-Classes.git
-    GIT_TAG        09a586011a296cf8ce3ffb9c15db7ce474ea4363
+    GIT_REPOSITORY https://github.com/gir489returns/GTAV-Classes.git
+    GIT_TAG        bdbb7f74f30c23aa38177ae087ee66386e91f670
     GIT_PROGRESS TRUE
     CONFIGURE_COMMAND ""
     BUILD_COMMAND ""
--- a/cmake/gtav-classes.cmake
+++ b/cmake/gtav-classes.cmake
@@ -2,8 +2,8 @@ include(FetchContent)
 
 FetchContent_Declare(
     gtav_classes
-    GIT_REPOSITORY https://github.com/gir489returns/GTAV-Classes.git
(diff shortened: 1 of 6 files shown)


cmake/gtav-classes.cmake
```diff
@@ -2,8 +2,8 @@ include(FetchContent)
 
 FetchContent_Declare(
     gtav_classes
-    GIT_REPOSITORY https://github.com/Yimura/GTAV-Classes.git
-    GIT_TAG        09a586011a296cf8ce3ffb9c15db7ce474ea4363
+    GIT_REPOSITORY https://github.com/gir489returns/GTAV-Classes.git
```
Yimura: After this is changed to the correct dependency it can be merged.

Yimura (approved):

PR #2894, opened 2024-03-30
fix nlohmann potential throw in ped outfit

### LINKED ISSUE #2893: [Bug]: Crash with invalid json through self_persist_outfit looped
### Describe the bug

When i open game and go singleplayer or multiplayer and open my mod it crush my game, I tried a few injector and everything looks the same.

### Steps To Reproduce

1.open game 2.go to single/multi 3.open mod 4.see crash log

### Expected Behavior

I expected that my mod will work well not crash.

### Logs and/or screenshots

[crash log.txt](https://github.com/YimMenu/YimMenu/files/14809221/crash.log.txt)

### Store

Rockstar

### Language

Polish

### YimMenu Language

English (US)

### Additional context

_No response_

### Build

### Coload

xiaoxiao921:
Fix #2893

### DIFF
--- a/src/backend/looped/self/persist_outfit.cpp
+++ b/src/backend/looped/self/persist_outfit.cpp
@@ -50,19 +50,29 @@ namespace big
 		static nlohmann::json outfit{};
 		static std::string persisting_outfit = "";
 
-		if (persisting_outfit != g.self.persist_outfit)
+		if (persisting_outfit != g_self_persist_outfit.value())
 		{
-			persisting_outfit        = g.self.persist_outfit;
-			folder saved_outfit_path = g_file_manager.get_project_folder("saved_outfits");
-			std::ifstream i(saved_outfit_path.get_file(persisting_outfit).get_path());
-			outfit.clear();
-			try
+			persisting_outfit                   = g_self_persist_outfit;
+			folder saved_outfit_path            = g_file_manager.get_project_folder("saved_outfits");
+			const auto persist_outfit_file_path = saved_outfit_path.get_file(persisting_outfit).get_path();
+			if (std::filesystem::exists(persist_outfit_file_path))
 			{
-				i >> outfit;
-			}
-			catch (const std::exception& e)
-			{
-				outfit = {};
+				std::ifstream i(persist_outfit_file_path);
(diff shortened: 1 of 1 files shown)


src/backend/looped/self/persist_outfit.cpp
```diff
@@ -15,13 +15,13 @@ namespace big
...
-				break;
-			}
-			default: return false; //For non-normal models
+		case "mp_m_freemode_01"_J: break;
+		case "mp_f_freemode_01"_J:
+		{
+			offset = 1;
+			break;
+		}
+		default: return false; //For non-normal models
```
gir489returns, line 24: Unnecessary whitespace change.

src/backend/looped/self/persist_outfit.cpp
```diff
@@ -56,7 +56,14 @@ namespace big
...
 			std::ifstream i(saved_outfit_path.get_file(persisting_outfit).get_path());
 			outfit.clear();
-			i >> outfit;
+			try
+			{
+				i >> outfit;
+			}
+			catch (const std::exception& e)
+			{
+				outfit = {};
```
gir489returns, line 65: LOG(INFO) the exception string, and also zero out g.self.persist_outfit, otherwise it's just going to attempt to try again next frame.

gir489returns (changes requested):

src/backend/looped/self/persist_outfit.cpp
xiaoxiao921: This is just how it should've been since the beginning, my IDE apply the clang format of the YimMenu project when I save the file.

gir489returns: I think it looks stupid. The cases should be indented to the switches, since they belong to the switch's scope, not the other way around.

xiaoxiao921: I agree, but this is how the clang-format file was defined a long time ago now, any change to it would mean we'd need to do a PR that fix the files to the new formatting

gir489returns (approved):

Yimura (approved):

PR #2902, opened 2024-04-01
aimbot

xiaoxiao921:
TODO:

- Remove camera shake when looking through a sniper scope
- Better UI for ped type selection
- Working set camera rotation when driving a vehicle while in first person (currently do nothing)
- Translation support


src/core/settings.hpp
```diff
@@ -847,16 +847,16 @@ namespace big
...
+				bool enable                                     = false;
+				bool adjust_position_scalar_for_velocity        = true;
+				float adjust_position_scalar_for_velocity_value = 1.15f;
+				int64_t only_on_ped_type                        = -1;
+				bool only_on_player                             = false;
+				bool only_on_enemy                              = false;
+				float fov                                       = 90.f;
+				float distance                                  = 25.f;
+				int32_t selected_bone                           = (int32_t)ePedBoneType::HEAD;
+				NLOHMANN_DEFINE_TYPE_INTRUSIVE(aimbot, enable, adjust_position_scalar_for_velocity, adjust_position_scalar_for_velocity_value, only_on_ped_type, only_on_player, only_on_enemy, fov, distance)
```
gir489returns, line 859: Duplicate NLOHMANN_DEFINE_TYPE_INTRUSIVE,.

gir489returns (changes requested):

gir489returns:
Doesn't even compile.

Severity	Code	Description	Project	File	Line	Suppression State	Details
Error	C2084	function 'void big::menu_settings::weapons::aimbot::to_json(nlohmann::json_abi_v3_11_2::json &,const big::menu_settings::weapons::aimbot &)' already has a body	C:\Users\Admin\Desktop\YimMenu\out\build\x64-Release-FastCompile\YimMenu	C:\Users\Admin\Desktop\YimMenu\src\core\settings.hpp	860		
Error	C2084	function 'void big::menu_settings::weapons::aimbot::from_json(const nlohmann::json_abi_v3_11_2::json &,big::menu_settings::weapons::aimbot &)' already has a body	C:\Users\Admin\Desktop\YimMenu\out\build\x64-Release-FastCompile\YimMenu	C:\Users\Admin\Desktop\YimMenu\src\core\settings.hpp	860		
Error	C2679	binary '=': no operator found which takes a right-hand operand of type 'const big::menu_settings::weapons::aimbot' (or there is no acceptable conversion)	C:\Users\Admin\Desktop\YimMenu\out\build\x64-Release-FastCompile\YimMenu	C:\Users\Admin\Desktop\YimMenu\src\core\settings.hpp	889		
Error	C2672	'nlohmann::json_abi_v3_11_2::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::json_abi_v3_11_2::adl_serializer,std::vector<uint8_t,std::allocator<uint8_t>>>::get_to': no matching overloaded function found	C:\Users\Admin\Desktop\YimMenu\out\build\x64-Release-FastCompile\YimMenu	C:\Users\Admin\Desktop\YimMenu\src\core\settings.hpp	889		
Error	C2065	'HEAD': undeclared identifier	C:\Users\Admin\Desktop\YimMenu\out\build\x64-Release-FastCompile\YimMenu	C:\Users\Admin\Desktop\YimMenu\src\core\settings.hpp	858		
Error	C2653	'ePedBoneType': is not a class or namespace name	C:\Users\Admin\Desktop\YimMenu\out\build\x64-Release-FastCompile\YimMenu	C:\Users\Admin\Desktop\YimMenu\src\core\settings.hpp	858		
Error	C3861	'ePedBoneType': identifier not found	C:\Users\Admin\Desktop\YimMenu\out\build\x64-Release-FastCompile\YimMenu	C:\Users\Admin\Desktop\YimMenu\src\core\settings.hpp	858

rkwapisz:
I see a lot of new code that looks lifted from some other source, but the aimbot just doesn't work. Are there special usage instructions?

xiaoxiao921:
> I see a lot of new code that looks lifted from some other source, but the aimbot just doesn't work. Are there special usage instructions?

How so?

Please download the binary.zip from this thread and rename your %appdata%/YimMenu folder to something else, inject the dll, make sure no recoil and no spread is enabled, the aimbot of course , and prolly disable ANIMAL targetting cause it's fairly annoying, then report back.

I'm getting pretty good results but there are still afew things to take care of: 
- Clear los check is flawed but no usage of shapetest natives seems to yield any good results and i'm not sure why

i'm trying to reverse how the shapetest are done natively (outside of scripts), like how the crosshair / reticule is dimmed when targetting neutral relationship-peds / red when targetting bad relationship-peds but it comes with a whole can of worms. Once i'll have this figured out only two thing will remain
- first person aimbot while driving (none of the offsets seems to affect this particular camera currently for some reason)
- Some weird ass camera shake that seems hard coded which make us miss the targeted bone. (which uses SHOOTING_ABILITY I think?) when aiming through sniper scopes.

rkwapisz:
I've implemented the main part of the camera manipulation logic in my aimbot fork and so far it's working a lot better than trying it with SET_GAMEPLAY pitch and heading. Hooking the camera handling directly is a much more elegant solution so far.

I have a couple of questions that maybe you have answers to, since I'm trying to document this as much as possible.

1) What is the purpose of reset_aim_vectors? I don't see any differences in behavior when I compile a build with each call to it commented out.

2) What is the purpose of cam_follow_ped_camera2 and cam_follow_ped_camera3? They're different offsets, but I'm able to use cam_follow_ped_camera just fine on foot and I don't see any differences in behavior when writing to those additional vectors.

xiaoxiao921:
- reset_aim_vectors i'm not even sure at this point.
- Needed when driving a vehicle.

gir489returns:
> * reset_aim_vectors i'm not even sure at this point.

Does literally nothing. If you BP those offsets and see what reads them, clearly this was some kind of copy+pasted code from someone who copypasted from someone who also copypasted from someone who clearly doesn't understand what they do. And also the offsets are wrong.

rkwapisz:
I've been in and out of all kinds of vehicles and changing cameras between first-person and third-person, and CamGameplayDirector offsets 2C0, 2C8, and 3C0 always point to the same address. It's for this reason I'm suspicious that we're doing anything useful by writing to all three, unless you've discovered a corner case where they could be different.

rkwapisz:
Okay, so the reset_aim_vectors function - whatever it does - is what allows us to aim correctly out of a vehicle. It solves the issue of the aimbot aiming too low when the player is in a vehicle. Heh... does ANYONE know what it actually does?

On foot it seems to do nothing.

xiaoxiao921:
Switched to hooking the cam gameplay director update function and doing the logic there... works much better and also kill the need for the adjust function, can't commit right now cause I need to clean up stuff still

rkwapisz:
I've been hooking the camera gameplay director function as well using the signature **4C 8B 35 ? ? ? ? 33 FF 32 DB** which is a pointer to the exact same address that your signature in the commit leads to, I believe. But I look forward to your update where you mention you're able to do away entirely with resetting the aim vector since I'm still investigating the offsets involved.

The gameplay camera is at 0x2C0, but you can get there from 0x2C8 and 0x3C0 as I mentioned above.
Once you're in there, camera metadata is at 0x10, which is where the reset aim vector code starts doing its thing.

The only offset that I can see that matters in the metadata is 0x2AC. I've commented out all the other stuff and have noticed no impacts. When you're in a vehicle, 0x2AC is set to -2, otherwise it's 0 when on foot. The reset_aim_vector function changes this to a 0 which is what corrects the vertical aim, probably by tricking the game camera into thinking we're on foot.

@gir489returns mentioned setting some BPs on these offsets; this one in particular is accessed by the following instruction:
`GTA5.exe+207BA4 - F3 0F10 88 AC020000   - movss xmm1,[rax+000002AC]`

The entire function itself is interesting and extremely simple:
```
push rbx
sub rsp,20
mov rax,[GTA5.exe+2002888]
mov rbx,rcx
cmp byte ptr [rax+000008F4],00
je GTA5.exe+207B72
xorps xmm0,xmm0
jmp GTA5.exe+207BB7
call GTA5.exe+268950
test al,al
mov rax,[rbx+00000230]
je GTA5.exe+207BA4
movss xmm0,[rax+000002AC]
movss xmm1,[rax+000002B0]
subss xmm1,xmm0
mulss xmm1,[rbx+00000254]
addss xmm1,xmm0
jmp GTA5.exe+207BAC
movss xmm1,[rax+000002AC]
mulss xmm1,[GTA5.exe+19EF28C]
movaps xmm0,xmm1
add rsp,20
pop rbx
ret 
```
I have no idea why this function is doing what it's doing, and why it's using the offset at 2AC to adjust the way the gameplay camera behaves in a vehicle vs. on foot. My best guess is that it may be related to look-and-feel of manipulating the camera in a vehicle... maybe there are some subtle changes that R* implemented to the way the camera responds to mouse movement in a vehicle vs. on foot purely for user experience.

xiaoxiao921:
I still need to reset aim vectors I think, I was referring to adjust_position_for_target_velocity

Yimura:
Status of this PR? As it's on ready to merge currently.

xiaoxiao921:
Not done but I don't really have the energy to work on it anymore right now, it's very close to be done. I'm gonna update the original post for a todo

lonelybud:
The work is very impressive. I tested on third person and everything seems to be fine even vehicles like half_track weapons rocks. The only thing i am worried about is smoothness. For example put three same type of targets very close to each other then one should be able to select his favorite target out of three by moving mouse.

gir489returns:
@xiaoxiao921 This branch has become so stale, that merging it with master would not be possible anymore. Do you still plan to return to this branch? It might be a better idea to reset this branch back to master's head, and then redo the initial commit.

src/core/enums.hpp
```diff
@@ -296,6 +296,41 @@ namespace big
 		STEAM     = 1 << 15,
 		WATER     = 1 << 16,
 	};
+
+	enum class ePedTypeFlag : int64_t
+	{
+		PED_TYPE_PLAYER_0              = 0,
+		PED_TYPE_PLAYER_1              = 1 << 0,
```
gir489returns, line 303: PLAYER_0 cannot be switched off, because you have not declared a bit for it.

gir489returns (changes requested):

gir489returns (dismissed):
I have tested the new changes and they work very well.

Yimura (dismissed):

src/byte_patch_manager.cpp
```diff
@@ -7,6 +7,7 @@
 #include "util/explosion_anti_cheat_bypass.hpp"
 #include "util/vehicle.hpp"
 #include "util/world_model.hpp"
+#include "backend/commands/weapons/no_sway.h"
```
Yimura: C header include, please rename to `.hpp`.

Yimura (changes requested):

src/byte_patch_manager.cpp
gir489returns: @Yimura HPP stands for Header Plus Plus. It is not using anything from C++ (It's a structure)

Yimura (approved):

PR #2904, opened 2024-04-01
Fixed similar personal vehicle previews not updating if the root vehicle hash was the same.

### LINKED ISSUE #2903: [Bug]: Updated personal vehicle spawn implementation does not correctly update preview models between same-model vehicles
### Describe the bug

Commit #2888 seems to have introduced a minor bug that is encountered when scrolling between two or more same personal vehicle models. The preview will not update past the first model.

For example, I have two armored Kurumas - one white and one black. When I mouse-over the first Kuruma in the list, the preview will show me the white one. When I then mouse-over the second Kuruma, the rotation preview model will continue to display the same white Kuruma, not the second Black one.

The same thing happens going the other direction. Only the first vehicle preview model is shown (black Kuruma), and will not correctly update to the white Kuruma above it in the list. This occurs for all vehicles that have multiples across the player's garages.

### Steps To Reproduce

1. Have two identical model personal vehicles (e.g., Obey 8F, Kuruma, etc.) - the vehicles don't need to be exact duplicates in terms of configuration, they just need to have the same display name (minus the license plate) in the menu
2. Enter the personal vehicle spawn window in Yim, and mouse-over one of the vehicles in the list that has multiples
3. Observe the previewed vehicle model
4. Immediately mouse-over another vehicle model that is the same model as the first, without mousing over any other vehicles in between
5. Observe that the previewed vehicle model has not updated, and continues to show the first-selected vehicle

### Expected Behavior

The preview vehicle model should be updating with every mouse-over

### Logs and/or screenshots

_No response_

### Store

Rockstar

### Language

English

### YimMenu Language

English (US)

### Additional context

_No response_

### Build

### Coload

gir489returns:
Closes #2903

### DIFF
--- a/src/services/model_preview/model_preview_service.cpp
+++ b/src/services/model_preview/model_preview_service.cpp
@@ -71,7 +71,7 @@ namespace big
 
 	void model_preview_service::show_vehicle(const std::map<int, int32_t>& owned_mods, bool spawn_max)
 	{
-		if (m_running && m_veh_model_hash != owned_mods.find(MOD_MODEL_HASH)->second)
+		if (m_running && m_veh_owned_mods != owned_mods)
 		{
 			stop_preview();
 			return;


gir489returns:
This is ready for review and/or merging.

Yimura (approved):

PR #2926, opened 2024-04-08
New Feature: Bone ESP

rkwapisz:
Added a configurable Bone ESP which draws a player skeleton on the screen using ImGui. Player skeletons are based on ePedBoneType, which currently supports 8 "root" bones. Checks are in place to ensure validity of bones prior to drawing.

Bone ESP provides additional visual movement and position information that is not otherwise provided by Box ESP.

Configuration is accessed below Box ESP, and is done the same way. Box ESP and Bone ESP can be used concurrently or separately; each has its own configuration parameters.

A matching PR in Translations includes English descriptions for the new configurables.

TODO:
- Possibly increase the complexity/resolution of the skeleton by using additional bones outside of ePedBoneType. Good additional bones to include are shoulders, elbows, and knees.

- Determine why repeated calls to m_get_screen_coords_for_world_coords cause a crash; in the meantime we are using the native GET_SCREEN_COORD_FROM_WORLD_COORD.

### DIFF
--- a/src/core/settings.hpp
+++ b/src/core/settings.hpp
@@ -970,9 +970,11 @@ namespace big
 			float global_render_distance[2] = {0.f, 600.f};
 			float tracer_render_distance[2] = {200.f, 600.f};
 			float box_render_distance[2]    = {0.f, 150.f};
+			float bone_render_distance[2]   = {0.f, 150.f};
 			bool tracer                     = true;
 			float tracer_draw_position[2]   = {0.5f, 1.f};
-			bool box                        = true;
+			bool box                        = false;
+			bool bone                       = true;
 			bool health                     = true;
 			bool armor                      = true;
 			bool god                        = true;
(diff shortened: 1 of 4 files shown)


gir489returns:
This invokes get_bone_coords which we had to remove in the past from it causing issues. I vote we reject this request unless the author can prove the crash has been identified and fixed.

rkwapisz:
There may be long-term stability issues with get_bone_coords that I'm unaware of.
I'll submit a new PR tonight that falls back on using GET_ENTITY_BONE_POSTION instead.

The benefit is this method lets me use more bones, so it'll come with a higher-fidelity skeleton too.

rkwapisz:
No longer using get_bone_coords, though I didn't experience any instability with it during my testing with it over a few hours.

Using GET_ENTITY_BONE_POSTION also lets me grab more bones, so I built a better skeleton.

Example image of three skeletons inside a vehicle.

image

gir489returns:
Now you are invoking natives in the DX layer, which is just as unacceptable.

rkwapisz:
Thanks for your help in the discussion. I should have dug deeper and discovered that I'm replicating previous efforts.
We can block this merge until a better solution to the get_bone_coords problems.

xiaoxiao921:
Iirc get bone cords simply access the transformation matrix, I'm not sure how this can cause any crashes

gir489returns:
> Iirc get bone cords simply access the transformation matrix, I'm not sure how this can cause any crashes

Yes which is why MBG and I couldn't figure out why it was causing the crash.

Issue #2928, opened 2024-04-08
[Bug]: Chat spam problem


noNickidea:
### Describe the bug

R* fixed chat spamming, i never seen any chat spammers until i injected this menu. Normally we should not see any chat spammers even without menu, but this menu makes chat spams visible for me.

### Steps To Reproduce

inject menu and go online

### Expected Behavior

not seeing any chat spam

### Logs and/or screenshots

```
<paste your logs here>
```

### Store

Epic Games

### Language

English

### YimMenu Language

English (US)

### Build

- [X] I've confirmed the issue exists on the latest version of YimMenu

### Coload

- [X] I've tested this without co-loading anything.

Yimura:
This makes sense as they have implemented a system where all chats get sent to their server on the receiving end. Back then it was god awful slow and thus we bypassed that and directly sent messages to the game. Perhaps we should remove this and just let the game do its thing when receiving messages.

gir489returns:
@maybegreat48 was also investigating an issue I sent him where if the chat was sent before the user joined it would slip through because we do a null check on CNetGamePlayer.

GreySideMM:
I honestly never have seen spam in chat since I started to use Yim (except for once). I just get the notifications for spammers being active. Did R* change something within the last couple of weeks?

PR #2931, opened 2024-04-09
feat: Chat translator

sch-lda:
Local version 
image


gir489returns:
@DayibBaba found some kind of translating service that's completely localized to the user's machine and never phones home. While there is 0 expectation of privacy in a public game, we really don't want there to be any kind of fingerprinting available against the system, especially one that's so easily spotted from 50 miles away.

Tayipc38:
> @DayibBaba found some kind of translating service that's completely localized to the user's machine and never phones home. While there is 0 expectation of privacy in a public game, we really don't want there to be any kind of fingerprinting available against the system, especially one that's so easily spotted from 50 miles away.

I am using [This](https://libretranslate.com/). You can run it locally with docker and here is a small [POC](https://pastebin.com/ryA3wvuf).

sch-lda:
image
image

Indeed, it is a completely free solution, but it often omits a portion of the original text when punctuation marks are missing. However, we cannot expect people to use formal language while chatting. If there is time, I will test the actual experience in the game. Additionally, a detailed installation guide is required. LibreTranslate‘s official guide may be easy for developers to understand, but not for ordinary users.

Tayipc38:
> Indeed, it is a completely free solution, but it often omits a portion of the original text when punctuation marks are missing. However, we cannot expect people to use formal language while chatting. If there is time, I will test the actual experience in the game. Additionally, a detailed installation guide is required. LibreTranslate‘s official guide may be easy for developers to understand, but not for ordinary users.

It being free does come with setbacks. It is inferior to the ones you have listed in terms of translation quality.

Tayipc38:
Idk how viable it is to expect yim users to be able to run a local docker container. Either you provide an in-depth guide or hope for the best?

gir489returns:
> Idk how viable it is to expect yim users to be able to run a local docker container. Either you provide an in-depth guide or hope for the best?

Considering the use case is going to be mostly Chinese people, I think sch-ida already has a fairly elaborate onboarding process for them, so much so, he could probably even host an server within the great firewall that they can easily access that doesn't phone home for them, according to the CCP it would just look like regular glorious communist internet traffic.

Regardless, I don't know if this would do well as a major commit, there's too much changed and the use case is very specific, Not to mention, we're probably going to be rolling back a lot of our chat code due to #2928.

sch-lda:
> Regardless, I don't know if this would do well as a major commit, there's too much changed and the use case is very specific, Not to mention, we're probably going to be rolling back a lot of our chat code due to #2928.

I am now also confused about this. I posted a download link in a Discord server (with built in Microsoft translation support), but only five different IP accessed it within 48 hours. Although there are occasional inquiries about chat translation features, the actual demand is not high. Once users need to install Docker/Python and deploy LibreTranslate themselves, I believe the usage will be even lower.
I'm not sure what will be removed due to #2928. If hooks::receive_net_message is removed, then this will no longer be available. Perhaps I should wait to see what happens next? But I think I should remove send translate result to others now to avoid being detected by R*.

These codes may indeed not be suitable for Yimmenu codebase. I've also never seen this feature in other GTA Online menus; it typically only exists in their Lua. Users can choose translation services that are suitable for their region. Unfortunately, Yimmenu refuses to add network-related APIs for Lua due to security concerns.
image


> Idk how viable it is to expect yim users to be able to run a local docker container. Either you provide an in-depth guide or hope for the best?

If this is about to merge, I will write an setup guide.

Yimura:
I actually quite like the idea of a chat translator, I'm all for allowing this in YimMenu.

HOWEVER, I'm not sure if we should allow translation options for services that clearly don't want to be leached of off (Google, Microsoft, DeepL... ). And I see that latest code in this PR no longer contains these services and only the libra translate part, ideally I'd like a few different setup options (like Docker instead of direct deploying on my machine). I may contribute these after you've written the initial setup guide.

Lua's not being able to access the internet seems still for the best for now as I really don't want shell code to be injected from remote sources which is obviously possible through Lua as it has a memory writing API available as well as a memory allocation API. The most important tasks that require some kind of a network connection should be handle by YimMenu.

GreySideMM:
I also would like this since I thought about this by myself.
Maybe expose/implement some complementary Lua functions (or similar) to allow users to specify any service they want? Since well, um, some other or future local software might need a little bit different handling of stuff. We can never know.

sch-lda:
>  ideally I'd like a few different setup options (like Docker instead of direct deploying on my machine). I may contribute these after you've written the initial setup guide.

Where should the setup guide be placed? In YimMenu/docs or https://github.com/YimMenu/YimMenu/wiki?

Kurd1stan:
> > ideally I'd like a few different setup options (like Docker instead of direct deploying on my machine). I may contribute these after you've written the initial setup guide.
> 
> Where should the setup guide be placed? In YimMenu/docs or https://github.com/YimMenu/YimMenu/wiki?

I think adding a guide to YimMenu/docs would be the best solution

Yimura:
Personally I'd add it to the wiki, however there's currently no way to easily contribute to the wiki pages unless you're a collaborator.

The `docs/` folder is used because we programmatically generate the files that exist within it.

src/services/api/api_service.cpp
```diff
@@ -18,6 +18,46 @@ namespace big
...
+				std::string source_language = obj["detectedLanguage"]["language"];
+				std::string result = obj["translatedText"];
+
+				if (source_language == g.session.chat_translator_target && g.session.chat_translator_bypass)
+					return "";
+				return result;
+			}
+			catch (std::exception& e)
+			{
+				g_notification_service.push_error("END_KICK"_T.data(),
```
Rxann: Why is the end kick translation being used here?

src/services/api/api_service.cpp
```diff
@@ -18,6 +18,46 @@ namespace big
...
+				    "BACKEND_END_SESSION_KICK_FORCE_SCRIPT_HOST_FAILED"_T.data());
```
Rxann: Here aswell?

src/services/api/api_service.cpp
sch-lda: oops, idk why I dragged it over there

src/services/api/api_service.cpp
sch-lda: oops, idk why I dragged it over there

src/util/chat.hpp
```diff
@@ -280,3 +280,14 @@ namespace big::chat
...
+
+namespace big
+{
+	struct chat_message
+	{
+		std::string sender;
+		std::string content;
+	};
+
+	extern std::queue<chat_message> translate_Queue;
```
Yimura: 2 things I'd change here:
1. Change `extern` to `inline` as there's no relevant `chat.cpp` file so the instantiation origin is lost and requires further searching through the code base.
2. Odd naming, it's using a mix between camelCase and snake_case. Please swap this to snake_case.

src/core/settings.hpp
```diff
@@ -407,6 +407,12 @@ namespace big
 			bool trust_friends                     = false;
 			bool trust_session                     = false;
 			bool use_spam_timer                    = true;
+			bool chat_translator                   = false;
+			bool chat_translator_print             = false;
+			bool chat_translator_draw              = true;
+			bool chat_translator_bypass            = true;
+			std::string chat_translator_target     = "en";
+			std::string chat_translator_endpoint   = "http://localhost:5000/translate";
```
Yimura: Probably best to put it into a dedicated struct.

src/backend/looped/session/chat_translator.cpp
```diff
@@ -0,0 +1,47 @@
...
+				if (translate_result != "")
+				{
+					if (g.session.chat_translator_draw)
+					{
+						if (rage::tlsContext::get()->m_is_script_thread_active)
+							chat::draw_chat(translate_result.c_str(), sender.c_str(), false);
+						else
+							g_fiber_pool->queue_job([translate_result, sender] {
+								chat::draw_chat(translate_result.c_str(), sender.c_str(), false);
+							});
```
Yimura: With this piece of code being copy-pasted now it's probably best to separate this into a helper function (as well as replace the original location where you got this code from with that helper function).

src/services/api/api_service.cpp
```diff
@@ -18,6 +18,44 @@ namespace big
 		g_api_service = nullptr;
 	}
 
+	std::string api_service::get_translation_from_LibreTranslate(std::string message, std::string target_language)
```
Yimura: Function name is a mix between snake_case and CamelCase.

2 options, change the name completely to snake_case or rework the name and add a comment.

```cpp
/**
 * Makes an API call to a LibreTranslate endpoint.
 */
std::string api_service::get_translation(std::string message, std::string target_language)
```

Yimura (changes requested):
Doing one of the final reviews on this PR (because I want this merged at some point), being a bit nit picky currently.

Yimura (approved):
No complaints anymore, keeping it open for a couple of hours to see if anyone else has a comment.

gir489returns (approved):

PR #2941, opened 2024-04-13
fix(lua): fix pointer.allocate allocating the wrong number of bytes

### LINKED ISSUE #2940: [Bug]: Lua 'allocate' function implementation is incorrect
### Reason

Here's the problem:
https://github.com/YimMenu/YimMenu/blob/0dc5ad61374dfd9c817967ca94afe5eca7b523fe/src/lua/bindings/memory.cpp#L110

Here's an example:
https://godbolt.org/z/Eab9GEh9q

Problem: If we try to call from Lua `memory.allocate(42)`, it will allocate `1` byte with content `{ 42 }`, instead of allocating `42` bytes with content `{ 0, ..., 0 }`

Furthermore: `new uint8_t[](size)` syntax is incorrect in any compiler except MSVC, but that's not a point, just FYI

Solution:
```cpp
	static pointer allocate(int size, sol::this_state state)
	{
		std::byte* mem = new std::byte[size]();

		big::lua_module* module = sol::state_view(state)["!this"];
		module->m_allocated_memory.push_back(static_cast<void*>(mem));

		return pointer(reinterpret_cast<uint64_t>(mem));
	}

	static void free(pointer ptr, sol::this_state state)
	{
		delete[] reinterpret_cast<std::byte*>(ptr.get_address());

		big::lua_module* module = sol::state_view(state)["!this"];

		std::erase_if(module->m_allocated_memory, [ptr](void* addr) {
			return ptr.get_address() == reinterpret_cast<uint64_t>(addr);
		});
	}
```

xiaoxiao921:
Fix #2940

### DIFF
--- a/src/lua/bindings/memory.cpp
+++ b/src/lua/bindings/memory.cpp
@@ -107,7 +107,7 @@ namespace lua::memory
 	// Returns: pointer: A pointer to the newly allocated memory.
 	static pointer allocate(int size, sol::this_state state)
 	{
-		void* mem = new uint8_t[](size);
+		void* mem = new uint8_t[size]();
 
 		big::lua_module* module = sol::state_view(state)["!this"];
 		module->m_allocated_memory.push_back(mem);


playday3008 (commented):
lgtm

Yimura (approved):

PR #2967, opened 2024-04-23
Undo script patch on eject.

### LINKED ISSUE #2887: [Bug]: Script patches are not undone during module ejection.
### Describe the bug

[script_patch_service](https://github.com/YimMenu/YimMenu/blob/master/src/services/script_patcher/script_patch.cpp) never undoes its patch when ejecting.

### Steps To Reproduce

Inject YimMenu
Eject YimMenu
Reinject YimMenu
Script VM is patched twice for some reason.

### Expected Behavior

When the module is ejected, all script patches should be undone.

### Logs and/or screenshots

_No response_

### Store

Steam

### Language

English

### YimMenu Language

English (US)

### Additional context

_No response_

### Build

### Coload

gir489returns:
Closes #2887

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -270,17 +270,23 @@ BOOL APIENTRY DllMain(HMODULE hmod, DWORD reason, PVOID)
 				    std::this_thread::sleep_for(500ms);
 			    }
 
+				g_hooking->disable();
+			    LOG(INFO) << "Hooking disabled.";
+
+				script_patcher_service_instance.reset();
+			    LOG(INFO) << "Script Patcher Service reset.";
+			    byte_patch_manager_instance.reset();
+			    LOG(INFO) << "Byte Patch Manager uninitialized.";
+			    hooking_instance.reset();
+			    LOG(INFO) << "Hooking uninitialized.";
+			    native_hooks_instance.reset();
+			    LOG(INFO) << "Dynamic native hooker uninitialized.";
+
 			    g_script_mgr.remove_all_scripts();
 			    LOG(INFO) << "Scripts unregistered.";
 
 			    lua_manager_instance.reset();
 			    LOG(INFO) << "Lua manager uninitialized.";
-
-			    g_hooking->disable();
-			    LOG(INFO) << "Hooking disabled.";
-
-			    native_hooks_instance.reset();
-			    LOG(INFO) << "Dynamic native hooker uninitialized.";
 
 			    // Make sure that all threads created don't have any blocking loops
 			    // otherwise make sure that they have stopped executing
@@ -299,8 +305,6 @@ BOOL APIENTRY DllMain(HMODULE hmod, DWORD reason, PVOID)
 			    LOG(INFO) << "Player Database Service reset.";
 			    api_service_instance.reset();
 			    LOG(INFO) << "API Service reset.";
-			    script_patcher_service_instance.reset();
-			    LOG(INFO) << "Script Patcher Service reset.";
 			    gui_service_instance.reset();
 			    LOG(INFO) << "Gui Service reset.";
 			    gta_data_service_instance.reset();
@@ -323,17 +327,11 @@ BOOL APIENTRY DllMain(HMODULE hmod, DWORD reason, PVOID)
 			    LOG(INFO) << "Xml Vehicles Service reset.";
 			    LOG(INFO) << "Services uninitialized.";
 
-			    hooking_instance.reset();
-			    LOG(INFO) << "Hooking uninitialized.";
-
 			    fiber_pool_instance.reset();
 			    LOG(INFO) << "Fiber pool uninitialized.";
 
 			    g_renderer.destroy();
 			    LOG(INFO) << "Renderer uninitialized.";
-
-			    byte_patch_manager_instance.reset();
-			    LOG(INFO) << "Byte Patch Manager uninitialized.";
 
 			    pointers_instance.reset();
 			    LOG(INFO) << "Pointers uninitialized.";
--- a/src/services/script_patcher/script_patch.cpp
+++ b/src/services/script_patcher/script_patch.cpp
@@ -48,6 +48,14 @@ namespace big
 		std::memcpy(get_code_address(data, m_ip), m_original.data(), m_original.size());
 	}
 
+	void script_patch::eject()
+	{
+		if (m_data)
+		{
+			disable(m_data);
+		}
+	}
+
 	void script_patch::update(script_data* data)
 	{
 		if (m_ip == 0)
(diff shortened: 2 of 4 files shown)


gir489returns:
MBG has been silent for a week now, and I've asked him multiple times to look at the eject crash. I tried to find out why it's crashing for myself, and I can't figure it out. If I attach a debugger to GTA V when it's crashing, it says it has an access violation on some native called from the main function which just tells me we're doing something while the game is trying to do something. 

This commit reverses each script patch when the destructor is called, however, the crash from ejecting/reinjecting is still present.

Rxann:
Would a mutex fix the issue? Just something I thought may work.

gir489returns:
> Would a mutex fix the issue? Just something I thought of.

No. You have to discern between malicious traffic and legitimate traffic.

Rxann:
> > Would a mutex fix the issue? Just something I thought of.
> 
> 
> 
> No, the game itself is crashing for unknown reasons. If we could control a mutex from the game, that would help, but it doesn't use one for the main thread.

Okay, that does make sense. You said above we were accessing something the game was aswell, so could we just virtualprotect to block the game from trying to do what it needs to do at the address while we do it?

gir489returns:
> > > Would a mutex fix the issue? Just something I thought of.
> > 
> > 
> > No, the game itself is crashing for unknown reasons. If we could control a mutex from the game, that would help, but it doesn't use one for the main thread.
> 
> Okay, that does make sense. You said above we were accessing something the game was aswell, so could we just virtualprotect to block the game from trying to do what it needs to do at the address while we do it?

No, that was a shot in the dark. It only crashes "sometimes", so it stands to reason, that whatever we're doing only causes the crash "sometimes" due to multithreading. If it always occurred, it would have to be because of something we're directly doing. I believe the reason has something to do with the timing. Sometimes I can get it to occur after 3 inject/reinjects, sometimes I can never get it to occur.

rkwapisz:
I pushed these commits into my build for some testing since I've also had similar random crashes when unloading YimMenu.
My latest crash with these commits in my build is as follows. I am not sure if they are related, but since we're on the topic of unloading the menu...

```
Dumping stacktrace:
[0]	VCRUNTIME140.dll memmove
[1]	YimMenu.dll big::lua_patch::`scalar deleting destructor'
[2]	D:\Documents\GitHub\YimMenu\src\lua\lua_module.cpp L: 136 big::lua_module::~lua_module
[3]	D:\Documents\GitHub\YimMenu\src\lua\lua_manager.cpp L: 108 big::lua_manager::unload_all_modules
[4]	D:\Documents\GitHub\YimMenu\src\lua\lua_manager.cpp L: 42 big::lua_manager::~lua_manager
[5]	C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\include\memory L: 3283 std::unique_ptr<big::lua_manager,std::default_delete<big::lua_manager> >::reset
[6]	D:\Documents\GitHub\YimMenu\src\main.cpp L: 289 `DllMain'::`5'::<lambda_1>::operator()
[7]	D:\Documents\GitHub\YimMenu\src\main.cpp L: 347 `DllMain'::`5'::<lambda_1>::<lambda_invoker_cdecl>
[8]	KERNEL32.DLL BaseThreadInitThunk
[9]	ntdll.dll RtlUserThreadStart
--------End of exception--------
```

gir489returns:
@xiaoxiao921 Could the crash be caused by something Lua is doing during its cleanup phase?

Rxann:
Looking at that, in the lua_module destructor, it seems to be deleting memory. This may have something to do with it, just thought I would share:

Line 134, lua_module.cpp
```cpp
for (auto memory : m_allocated_memory)
			delete[] memory;

rkwapisz:
> Looking at that, in the lua_module destructor, it seems to be deleting memory. This may have something to do with it, just thought I would share:
> 
> Line 134, lua_module.cpp
> 
> ```c++
> for (auto memory : m_allocated_memory)
> 			delete[] memory;
> 			 
> ```

Yes, my initial guess is that the destructor is not inherently safe since we're not checking for null pointers before trying to delete blocks of memory. I've been testing loading/unloading the menu with the following change, so far without issues:

`
        for (auto memory : m_allocated_memory)
        {
            if (memory != nullptr)
            {
                delete[] memory;
                memory = nullptr;
            }
        }
`

xiaoxiao921:
Well the code is fine. Could you tell me how it is wrong?
https://github.com/search?q=repo%3AYimMenu%2FYimMenu%20m_allocated_memory&type=code

gir489returns:
> > Looking at that, in the lua_module destructor, it seems to be deleting memory. This may have something to do with it, just thought I would share:
> > Line 134, lua_module.cpp
> > ```c++
> > for (auto memory : m_allocated_memory)
> > 			delete[] memory;
> > 			 
> > ```
> 
> Yes, my initial guess is that the destructor is not inherently safe since we're not checking for null pointers before trying to delete blocks of memory. I've been testing loading/unloading the menu with the following change, so far without issues:
> 
> `for (auto memory : m_allocated_memory) { if (memory != nullptr) { delete[] memory; memory = nullptr; } }`

[This makes no sense](https://stackoverflow.com/questions/615355/is-there-any-reason-to-check-for-a-null-pointer-before-deleting).

xiaoxiao921:
Indeed, if anything you are calling pointer.free twice and in this case this is simply a skill issue.

rkwapisz:
> > > Looking at that, in the lua_module destructor, it seems to be deleting memory. This may have something to do with it, just thought I would share:
> > > Line 134, lua_module.cpp
> > > ```c++
> > > for (auto memory : m_allocated_memory)
> > > 			delete[] memory;
> > > 			 
> > > ```
> > 
> > 
> > Yes, my initial guess is that the destructor is not inherently safe since we're not checking for null pointers before trying to delete blocks of memory. I've been testing loading/unloading the menu with the following change, so far without issues:
> > `for (auto memory : m_allocated_memory) { if (memory != nullptr) { delete[] memory; memory = nullptr; } }`
> 
> [This makes no sense](https://stackoverflow.com/questions/615355/is-there-any-reason-to-check-for-a-null-pointer-before-deleting).

Understood. I thought a double-check was prudent, because otherwise I cannot understand why delete[] was being considered scalar.

gir489returns:
>I cannot understand why delete[] was being considered scalar.

What in the absolute C++20 standard are you talking about?!

xiaoxiao921:
Makes sense, so a fix would be to make sure the vm is not currently executing something we are about to patch, my gut feeling would be to suspend the thread that the vm runs on and observe its instruction pointer and making sure it's not near any of the stuff that we patch

gir489returns:
Invalidated by #3288.

PR #2973, opened 2024-04-25
Radio Sync Fix

Rxann:
I have tested, and with this change, the radios seem to sync up now.

### DIFF
--- a/src/hooks/protections/receive_net_message.cpp
+++ b/src/hooks/protections/receive_net_message.cpp
@@ -17,7 +17,6 @@
 #include <network/netTime.hpp>
 
 
-
 inline void gamer_handle_deserialize(rage::rlGamerHandle& hnd, rage::datBitBuffer& buf)
 {
 	if ((hnd.m_platform = buf.Read<uint8_t>(sizeof(hnd.m_platform) * 8)) != rage::rlPlatforms::PC)
@@ -85,6 +84,7 @@ namespace big
 
 		rage::eNetMessage msgType;
 		player_ptr player;
+		rate_limiter unk_player_radio_requests{5s, 2};
 
 		for (uint32_t i = 0; i < gta_util::get_network()->m_game_session_ptr->m_player_count; i++)
 		{
@@ -236,7 +236,19 @@ namespace big
 			switch (msgType)
 			{
 			case rage::eNetMessage::MsgScriptMigrateHost: return true;
-			case rage::eNetMessage::MsgRadioStationSyncRequest: return false;
+			case rage::eNetMessage::MsgRadioStationSyncRequest:
+			{
+				if (unk_player_radio_requests.process())
+				{
+					if (unk_player_radio_requests.exceeded_last_process())
+					{
+						// Make a translation for this new OOM kick protection
+						g_notification_service.push_error("PROTECTIONS"_T.data(), "OOM_KICK"_T.data());
+						return true;
+					}
+				}
+				return false;
+			}
 			}
 		}
 
--- a/src/hooks/protections/receive_net_message.cpp
+++ b/src/hooks/protections/receive_net_message.cpp
@@ -84,7 +84,6 @@ namespace big
 
 		rage::eNetMessage msgType;
 		player_ptr player;
-		rate_limiter unk_player_radio_requests{5s, 2};
 
 		for (uint32_t i = 0; i < gta_util::get_network()->m_game_session_ptr->m_player_count; i++)
 		{
@@ -238,6 +237,8 @@ namespace big
 			case rage::eNetMessage::MsgScriptMigrateHost: return true;
 			case rage::eNetMessage::MsgRadioStationSyncRequest:
 			{
+				static rate_limiter unk_player_radio_requests{2s, 2};
+
 				if (unk_player_radio_requests.process())
 				{
 					if (unk_player_radio_requests.exceeded_last_process())
--- a/src/hooks/protections/receive_net_message.cpp
+++ b/src/hooks/protections/receive_net_message.cpp
@@ -2,6 +2,7 @@
 #include "backend/context/chat_command_context.hpp"
 #include "backend/player_command.hpp"
 #include "core/data/packet_types.hpp"
+#include "gta/enums.hpp"
 #include "gta/net_game_event.hpp"
 #include "gta_util.hpp"
 #include "hooking/hooking.hpp"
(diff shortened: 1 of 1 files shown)


src/hooks/protections/receive_net_message.cpp
```diff
@@ -235,7 +236,7 @@ namespace big
 			switch (msgType)
 			{
 			case rage::eNetMessage::MsgScriptMigrateHost: return true;
-			case rage::eNetMessage::MsgRadioStationSyncRequest: return true;
+			case rage::eNetMessage::MsgRadioStationSyncRequest: return false;
```
gir489returns, line 239: Now we will be susceptable to OOM kicks. You have to address both scenarios.

gir489returns (changes requested):

src/hooks/protections/receive_net_message.cpp
Rxann: I will do this.

Yimura: To be fair, there was no comment at all for why that was implemented. There probably should've been a comment there explaining that it patches a certain crash/kick.

src/hooks/protections/receive_net_message.cpp
```diff
@@ -85,6 +84,7 @@ namespace big
 
 		rage::eNetMessage msgType;
 		player_ptr player;
+		rate_limiter unk_player_radio_requests{5s, 2};
```
gir489returns, line 87: Why is this outside the scope of MsgRadioStationSyncRequest? Make it a static, too, otherwise the compiler will discard it.

src/hooks/protections/receive_net_message.cpp
```diff
@@ -236,7 +236,19 @@ namespace big
 			switch (msgType)
 			{
 			case rage::eNetMessage::MsgScriptMigrateHost: return true;
-			case rage::eNetMessage::MsgRadioStationSyncRequest: return false;
+			case rage::eNetMessage::MsgRadioStationSyncRequest:
+			{
+				if (unk_player_radio_requests.process())
+				{
+					if (unk_player_radio_requests.exceeded_last_process())
```
gir489returns, line 243: Since this is for "all" people in the car, we should probably down the limit to 1-2 seconds. We need to test what happens if someone just presses next channel on the radio a ton of times.

gir489returns (changes requested):

src/hooks/protections/receive_net_message.cpp
Rxann: Ok, I have changed to 2s for now to be on the high side.

src/hooks/protections/receive_net_message.cpp
```diff
@@ -2,16 +2,16 @@
...
 #include "gta_util.hpp"
 #include "hooking/hooking.hpp"
 #include "lua/lua_manager.hpp"
 #include "natives.hpp"
 #include "script/scriptIdBase.hpp"
 #include "services/players/player_service.hpp"
-#include "util/session.hpp"
 #include "util/chat.hpp"
-#include "gta/enums.hpp"
+#include "util/session.hpp"
```
gir489returns, line 14: Revert these unnecessary changes to the header.

src/hooks/protections/receive_net_message.cpp
```diff
@@ -120,7 +120,7 @@ namespace big
 					if (g.session.log_chat_messages)
 						chat::log_chat(message, player, spam_reason, is_team);
 					g_notification_service.push("PROTECTIONS"_T.data(),
-                                      
+
```
gir489returns, line 123: Unnecessary whitespace changes.

src/hooks/protections/receive_net_message.cpp
```diff
@@ -235,7 +235,21 @@ namespace big
...
+			case rage::eNetMessage::MsgRadioStationSyncRequest:
+			{
+				static rate_limiter unk_player_radio_requests{2s, 2};
+
+				if (unk_player_radio_requests.process())
+				{
+					if (unk_player_radio_requests.exceeded_last_process())
+					{
+						// Make a translation for this new OOM kick protection
+						g_notification_service.push_error("PROTECTIONS"_T.data(), "OOM_KICK"_T.data());
```
gir489returns, line 247: ...So are you gonna make a translation for this?

gir489returns (dismissed):

src/hooks/protections/receive_net_message.cpp
GreySideMM: Don't forget to call session::add_infraction(player, Infraction::TRIED_KICK_PLAYER).
Also: There might be no need for a different notification text, imo.

src/hooks/protections/receive_net_message.cpp
GreySideMM: Serious question: What exactly is the problem with this and similar changes (i.e. whitespaces)? Merging problems / more potential work?

src/hooks/protections/receive_net_message.cpp
Rxann: The current translation includes {} for a player name, and this pr handles the case where player is nullptr(not host) so we can't get the name of the player. We also can't call an infraction because we don't know the player.

src/hooks/protections/receive_net_message.cpp
Rxann: These were made by the formatter, but i will revert

src/hooks/protections/receive_net_message.cpp
gir489returns: > Don't forget to call session::add_infraction(player, Infraction::TRIED_KICK_PLAYER). Also: There might be no need for a different notification text, imo.

To who? We don't have a player pointer.

src/hooks/protections/receive_net_message.cpp
gir489returns: > Serious question: What exactly is the problem with this and similar changes (i.e. whitespaces)? Merging problems / more potential work?

It's just unnecessary.

Yimura: The reordering of includes is fine as that's just a formatter thing to do.

src/hooks/protections/receive_net_message.cpp
Yimura: Consequence of the formatter, those tabs shouldn't have been there to begin with.

Yimura (approved):

Yimura:
I'm considering this finished, I'll wait a few hours before merging this in case it wasn't.

Rxann:
> I'm considering this finished, I'll wait a few hours before merging this in case it wasn't.

Nope, if you think it's good it's good!

gir489returns:
> I'm considering this finished, I'll wait a few hours before merging this in case it wasn't.

I was waiting on him to fix the whitespace changes, but if you want to merge it as is that's fine with me.

Yimura:
> I was waiting on him to fix the whitespace changes, but if you want to merge it as is that's fine with me.

Those whitespace changes would've happened the next time someone touched the file with the formatter active (like me for example). It's important to differentiate between random newlines, line endings being changed and tabs being removed and includes being order alphabetically.

PR #2978, opened 2024-04-26
Update [BUG] issue template

Yimura:
So done with people not adding their log file, I've separated the screenshot and log file fields and made the log file field a requirement.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -28,16 +28,23 @@ body:
     required: true
 - type: textarea
   attributes:
-    label: Logs and/or screenshots
+    label: Logs
     description: |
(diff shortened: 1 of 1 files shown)


gir489returns:
I looked into trying to make the ticket enforce that an attachment be included, but Github has no such requirement. (Stupid)

With JIRA would have that. We had problems with retarded interns jamming the ticket system with stupid tickets exactly like this and not including their Tomcat log.

gir489returns (approved):

Yimura:
> I looked into trying to make the ticket enforce that an attachment be included, but Github has no such requirement. (Stupid)
> 
> With JIRA would have that. We had problems with retarded interns jamming the ticket system with stupid tickets exactly like this and not including their Tomcat log.

Yeah I found the same thread online and came to the same conclusion this is the next best thing.

PR #2980, opened 2024-04-26
Fixed commands not working in single player.

### LINKED ISSUE #2955: [Bug]: Problem with Vehicle spawning
### Describe the bug

When I go to my YimMenu I go to the Vehicle section, and I can’t spawn a vehicle because it says “No vehicle in the register”

### Steps To Reproduce

When I go to my YimMenu I go to the Vehicle section, and I can’t spawn a vehicle because it says “No vehicle in the register”

### Expected Behavior

Help me solve the problem, I want to spawn cars without problems

### Logs and/or screenshots

```
![изображение_2024-04-20_185357634](https://github.com/YimMenu/YimMenu/assets/167634793/d79f6312-3f6f-45b2-919f-ba318ebfbbcf)

<paste your logs here>
```

### Store

Steam

### Language

Russian

### YimMenu Language

Русский (RU)

### Additional context

_No response_

### Build

### Coload

gir489returns:
Closes #2955

### DIFF
--- a/src/services/players/player.cpp
+++ b/src/services/players/player.cpp
@@ -127,7 +127,7 @@ namespace big
 
 	uint8_t player::id() const
 	{
-		return m_net_game_player == nullptr ? -1 : m_net_game_player->m_player_id;
+		return m_net_game_player == nullptr ? self::id : m_net_game_player->m_player_id;
 	}
 
 	bool player::is_host() const
--- a/src/services/players/player.cpp
+++ b/src/services/players/player.cpp
@@ -127,7 +127,10 @@ namespace big
 
 	uint8_t player::id() const
 	{
-		return m_net_game_player == nullptr ? self::id : m_net_game_player->m_player_id;
+		if (*g_pointers->m_gta.m_is_session_started)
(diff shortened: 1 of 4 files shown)


src/services/players/player.cpp
```diff
@@ -127,7 +127,7 @@ namespace big
 
 	uint8_t player::id() const
 	{
-		return m_net_game_player == nullptr ? -1 : m_net_game_player->m_player_id;
+		return m_net_game_player == nullptr ? self::id : m_net_game_player->m_player_id;
 	}
```
Yimura: Wouldn't this change certain behaviour even if it's not our player object (for example when our target leaves online which may then end up targeting ourselves).

gir489returns: Interesting point, but it should still fail safe instead of use undefined behavior. If that happens, and we revert it to how it was before (if is online, and m_net_game_player  is nullptr) then we would return -1, and the command would run like how it was before, where it would just spawn a car at 0,0,0.

Yimura (approved):

PR #2989, opened 2024-04-27
REDO: Prevent autokick of friends and trusted

rkwapisz:
Previously closed PRs were a result of me deleting the branches after I'd merged them into my own fork, but I didn't realize that would also close the PR in the base. Whoops.

### DIFF
--- a/src/backend/looped/session/auto_kick_host.cpp
+++ b/src/backend/looped/session/auto_kick_host.cpp
@@ -12,7 +12,7 @@ namespace big
 		if (kick_host && !bLastKickHost)
 		{
 			g_player_service->iterate([](auto& plyr) {
-				if (plyr.second->is_host())
+				if (plyr.second->is_host() && !plyr.second->is_friend() && !plyr.second->is_trusted)
 				{
 					dynamic_cast<player_command*>(command::get("multikick"_J))->call(plyr.second, {});
 				}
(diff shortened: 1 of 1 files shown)


gir489returns (approved):

GreySideMM:
We currently usually check against the setting trust_friends. So we don't autotrust friends. You should add this to keep it consistent.

rkwapisz:
> We currently usually check against the setting trust_friends. So we don't autotrust friends. You should add this to keep it consistent.

Good catch!

src/backend/looped/session/auto_kick_host.cpp
```diff
@@ -12,6 +12,10 @@ namespace big
 		if (kick_host && !bLastKickHost)
 		{
 			g_player_service->iterate([](auto& plyr) {
+				// Don't kick trusted players
+				if ((plyr.second->is_friend() && g.session.trust_friends) || plyr.second->is_trusted)
```
gir489returns, line 16: Order of operations:

1. If they are trusted.
2. If we have trust_friends on
3. if they're a friend

This will give optimization notes to the compiler. We don't care if they're a friend if we have trust_friends disabled. We also don't care about either of those if they're already trusted.

gir489returns (changes requested):

gir489returns (approved):

PR #2990, opened 2024-04-27
REDO: Add Protection GUI Element for Give Collectible 

rkwapisz:
Previously closed PRs were a result of me deleting the branches after I'd merged them into my own fork, but I didn't realize that would also close the PR in the base. Whoops.

### DIFF
--- a/src/views/settings/view_protection_settings.cpp
+++ b/src/views/settings/view_protection_settings.cpp
@@ -22,38 +22,40 @@ namespace big
 		ImGui::Checkbox("FORCE_TELEPORT"_T.data(), &g.protections.script_events.force_teleport);
 		ImGui::Checkbox("GTA_BANNER"_T.data(), &g.protections.script_events.gta_banner);
 		ImGui::Checkbox("MC_TELEPORT"_T.data(), &g.protections.script_events.mc_teleport);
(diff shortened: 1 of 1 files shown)


gir489returns (commented):
Is there any particular reason you reordered these?

rkwapisz:
> Is there any particular reason you reordered these?

Aesthetics. The features are currently laid out in columns with six elements per group.

Adding this new toggle would have either required a single column of seven, or a new column of one. I decided that instead of making the menu wider, that it would be better visually to reorganize the checkboxes into groups of seven.

I am happy to defer to other preferences, though.

gir489returns:
While the YImMenu project is a hodgepodge of several people's work all slammed together with no regard for UX at all, we generally what to adhere to WYSIWYG design philosophy when we can, because that's generally what ImGui is based upon, or at least tries to.

That being said, if you wish to pitch a new design, it's generally a nice idea to post a screenshot of your new design, so that we can internally weigh the options of accepting the new design, or asking to revert.

rkwapisz:
Here is a screenshot of the proposed new Protection layout with the new Give Collectible checkbox.

image

gir489returns (approved):

gir489returns:
The new design logically makes sense, I can see why you made those changes.

PR #2991, opened 2024-04-27
 REDO: Disable Vehicle Heading Updates When Vehicle Flying in First-Person Mode 

rkwapisz:
Again, apologies for the previous closed PRs.
I'm still learning how to properly collaborate with Github.


gir489returns (approved):

gir489returns (dismissed):
The same change needs to be applied to no clip as well, as the same behavior occurs.

We should also warn the user that it shouldn't be used in FPV.

rkwapisz:
I'll update the noclip implementation in this PR, since both issues are related.

Regarding a player warning, I'm thinking two options (or or the other, or both).
1) A tooltip on mouse-over of the feature. Simply, out-of-the-way, but probably more easily missed since few people read tooltips.
2) A one-time warning from g_notification_service when the player enters FPV with either noclip or vehicle flymode active. More apparent, and we can easily make sure it only shows up once.

gir489returns:
The tooltip idea is a bad idea, as it assumes the user uses the GUI to activate this feature and not the hotkey or command variant. You need to compensate for all 3 scenarios. The 2nd idea is also bad, as the user may not see it, or even if they do, they may not remember the next time and become frustrated as to why it's not working.

gir489returns:
Referring to your notes from https://github.com/YimMenu/YimMenu/pull/2985 (And your commented code)

If you change car_rot and rotation_delta to static, you can determine what the previous call's Euclidian distance was from the new rotation. Thus, you could have a delta rotation of how much to turn the car based on where the user is now looking. I feel this is a better solution than just turning off the feature entirely.

rkwapisz:
> Referring to your notes from #2985 (And your commented code)
> 
> If you change car_rot and rotation_delta to static, you can determine what the previous call's Euclidian distance was from the new rotation. Thus, you could have a delta rotation of how much to turn the car based on where the user is now looking. I feel this is a better solution than just turning off the feature entirely.

I observed a related issue where, no matter what I had set the new rotation to, flying forward in the vehicle would automatically start pitching it down, resulting in a strange parabolic trajectory. I was unable to compensate for this pitch drop with the FPV camera. I think it's related to the fact that up-down camera movements are limited more than left-right, but even with different levels of clamping for horizontal vs. vertical camera movements I had the same issue.

Also, even with a reasonable amount of clamping to prevent the super-fast spinning, the camera would still "float" to where the camera was facing. It just didn't feel good, whereas the 3rd-person view is very snappy and responsive. I will revisit this issue, though.

Regarding No Clip, this feature actually seems to be working okay for me in third and first-person camera. See video.

https://github.com/YimMenu/YimMenu/assets/13259220/38b5c69b-278f-42e5-8760-f5cececa116b

gir489returns:
Yes, we fixed that in #1791 I was wondering if we could just apply the same code from FPV noclip to FPV vehicles, but what you're telling me is that it doesn't work?

rkwapisz:
It won't work reliably in a loop because when you're FPV in a car, your gameplay cam rotation will change as the vehicle entity rotation changes, so using GET_GAMEPLAY_CAM_ROT followed by SET_ENTITY_ROTATION in a vehicle is what will make you spin. Using Euclidean distance won't help since there is no constant reference to rotate towards.

If we want to use vehicle fly in first person, we will most likely need a different control scheme that doesn't rely on GET_GAMEPLAY_CAM_ROT. The problem is that I really like the 3rd-person control scheme, and using something different first FPV would feel jarring to users.

There's probably a half-decent solution in all this, I just need to figure something out.

gir489returns:
> It won't work reliably in a loop because when you're FPV in a car, your gameplay cam rotation will change as the vehicle entity rotation changes

You just answered your own question. Keep track of static Vehicle last_vehicle, if last_vehicle == self::veh do rotation code, else reset. Make sure your reset scenario sets last_vehicle = self::veh.

rkwapisz:
I haven't forgotten about this work and tried a few different things over the weekend. All of which were unsatisfactory.

The closest I got to having something BARELY tolerable was setting
`FORCE_CAMERA_RELATIVE_HEADING_AND_PITCH(0.f, 0.f, 0.f)`

after every call to
`ENTITY::SET_ENTITY_ROTATION`

So this would let me grab the camera's rotation, then apply it directly to the vehicle's rotation, and then FORCE_CAMERA would instantly reset our camera heading to prevent the car from infinitely moving in that direction.

This allowed to a decent degree of control for left-right axis movement despite some very strong shaking even without any input. However, I still had problems with the car tipping nose down - you can actually see this behavior in 3rd-person as well, where if you fly in a straight line without mouse movement your car will start to arc down towards the ground.

src/backend/looped/vehicle/fly.cpp
```diff
@@ -32,10 +32,34 @@ namespace big
...
 
-			if (vehicle)
+			if (vehicle && CAM::GET_FOLLOW_PED_CAM_VIEW_MODE() != CameraMode::FIRST_PERSON)
 			{
-				Vector3 cam_pos = CAM::GET_GAMEPLAY_CAM_ROT(0);
-				ENTITY::SET_ENTITY_ROTATION(vehicle, cam_pos.x, cam_pos.y, cam_pos.z, 1, true);
+				Vector3 cam_rot = CAM::GET_GAMEPLAY_CAM_ROT(0);
+
+				/*
+				* Leaving this for experimentation in the future, but vehicle flying in first person is very unpleasant either way
```
Rxann: Does this really need to stay?

rkwapisz: I don't understand. Do you mean cam_rot or the comment or something else?
The whole thing isn't even close to being finished.

Rxann: I reviewed line 40. Also, if it is not done, convert this to a draft PR.

gir489returns: @Rxann Nobody asked you.

gir489returns:
@rkwapisz I know, and I too tried to get the Euclidian distance method I mentioned working as well, but to no avail. The problem stems from the wildly inaccurate Euler angles (or lack thereof) returned from this engine, and how it expects the user to supply a Δangle instead of just the angle itself.

To be honest, I think we can merge it as-is, or maybe just go with the Plan B scenario of just turning this off entirely while in FPV mode, and then alerting the user. I'll leave this up to you.

src/backend/looped/vehicle/fly.cpp
Rxann: > @Rxann Nobody asked you.

When I run the formatter and it edits 3 lines of imports based on the defined formatting rules of YimMenu it is a problem, but when 29 lines of code that will not be used are left, it is not an issue?

gir489returns: The virgin whitespace vs the chad comment.

His comments are relevant to potential developers of this feature in knowing what we have tried, versus what we haven't. 

You running the clang formatter on every file you touch is absolutly annoying, and there's no real reason to do it, as none of the other developers do it.

Rxann: It is not my fault other developers don't format the files they commit.

gir489returns: It is your fault to check commits before you push them to the repo.

rkwapisz: > I reviewed line 40. Also, if it is not done, convert this to a draft PR.

Probably a fine idea moving forward.

rkwapisz:
I really can't decide what I want to do with with this. I'm thinking we just boot the player out of FPV if they're in a car with flying on, and also push a notification that tells them FPV is unsupported.

I'd rather provide some user feedback together with an automatic workaround rather than try to half-ass a fix that would only end up getting more bug reports against it.

Notification will be PR'd in translations:
`  "BACKEND_FLYING_VEHICLE_FPV_DISABLED": "First-person view is currently unsupported in Vehicle Fly mode.",`

I've left in some commented code in case I want to revisit this in the future, but I think @gir489returns is correct in that the issue stems from the engine's piss-poor angle handling.

Yimura (dismissed):

Yimura (approved):

PR #2994, opened 2024-04-29
refector: Chinese chat spam blacklist

sch-lda:
Avoid intercepting individual Chinese characters. Reduce false positive rate. Based on chat messages from 2023/12 GTAO Asian session.


gir489returns (dismissed):

sch-lda:
In the past few months, there have been some changes in AD contents. Survey based on last year's messages may not fully meet current requirements.

Yimura (approved):

PR #2996, opened 2024-04-29
fix(vehicle): fix repair not working after going in water

### LINKED ISSUE #2995: [Bug]: Vehicle -> Repair does not fix cars that been in water.
### Describe the bug

When a vehicle enters water, it dies completely. Pressing the repair button does not bring the vehicle back to working order.

### Steps To Reproduce

1. Spawn a car.
2. Drive it into a body of water so much so that the car is submerged and dies.
3. Using whatever method, get the car out of the water.
4. Navigate to Vehicle -> Repair and click the button.

### Expected Behavior

Vehicle should restart.

### Logs

image

### Screenshots / Video

https://github.com/YimMenu/YimMenu/assets/100792176/bb7bff6d-106e-41cf-b492-4d586bd124b6

### Store

Steam

### Language

English

### YimMenu Language

English (US)

### Additional context

_No response_

### Build

### Coload

xiaoxiao921:
Closes #2995 

It's a bandaid fix but it works, there are multiple issues though. First person can't be used anymore in that specific vehicle, and can't be rentered. I tried messing with the vehicle flags and physical flags without any success sadly.

### DIFF
--- a/src/util/vehicle.cpp
+++ b/src/util/vehicle.cpp
@@ -151,7 +151,7 @@ namespace big::vehicle
 			return false;
 		}
 
-		repair_vehicle_from_water(veh);
+		repair_engine_from_water(veh);
 
 		VEHICLE::SET_VEHICLE_FIXED(veh);
 		VEHICLE::SET_VEHICLE_DIRT_LEVEL(veh, 0.f);
--- a/src/util/vehicle.cpp
+++ b/src/util/vehicle.cpp
@@ -144,9 +144,28 @@ namespace big::vehicle
 			return false;
 		}
 
-		VEHICLE::SET_VEHICLE_ENGINE_HEALTH(veh, 1000);
-		VEHICLE::SET_VEHICLE_ENGINE_ON(veh, true, true, false);
+		auto cvehicle = (uint8_t*)g_pointers->m_gta.m_handle_to_ptr(veh);
+		// fix vehicle being completly fucked after going into water.
+		cvehicle[0xD8] = 0x20;
+
 		VEHICLE::SET_VEHICLE_FIXED(veh);
+		VEHICLE::SET_VEHICLE_DEFORMATION_FIXED(veh);
+		VEHICLE::SET_VEHICLE_NEEDS_TO_BE_HOTWIRED(veh, false);
+
+		if (FIRE::IS_ENTITY_ON_FIRE(veh))
+		{
+			FIRE::STOP_ENTITY_FIRE(veh);
+		}
+
+		VEHICLE::SET_VEHICLE_ENGINE_HEALTH(veh, 1000.0f);
+		VEHICLE::SET_VEHICLE_PETROL_TANK_HEALTH(veh, 1000.0f);
+		VEHICLE::SET_VEHICLE_BODY_HEALTH(veh, 1000.f);
+
+		VEHICLE::SET_VEHICLE_UNDRIVEABLE(veh, false);
+		VEHICLE::SET_VEHICLE_ENGINE_CAN_DEGRADE(veh, false);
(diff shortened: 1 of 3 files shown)


gir489returns (changes requested):
I exclusively play in FPV, this fix is not acceptable to me as the original ticket submitter.

src/util/vehicle.cpp
```diff
@@ -144,7 +144,28 @@ namespace big::vehicle
...
+		{
+			FIRE::STOP_ENTITY_FIRE(veh);
+		}
+
+		VEHICLE::SET_VEHICLE_ENGINE_HEALTH(veh, 1000.0f);
+		VEHICLE::SET_VEHICLE_PETROL_TANK_HEALTH(veh, 1000.0f);
+		VEHICLE::SET_VEHICLE_BODY_HEALTH(veh, 1000.f);
+
+		VEHICLE::SET_VEHICLE_UNDRIVEABLE(veh, false);
+		VEHICLE::SET_VEHICLE_ENGINE_CAN_DEGRADE(veh, false);
```
gir489returns, line 165: This will make the vehicle semi-godmodded, the user did not ask for that.

src/util/vehicle.cpp
```diff
@@ -144,7 +144,28 @@ namespace big::vehicle
...
+
+		VEHICLE::SET_VEHICLE_ENGINE_ON(veh, true, true, false);
```
gir489returns, line 167: SET_VEHICLE_ENGINE_ON and SET_VEHICLE_NEEDS_TO_BE_HOTWIRED seems to be pointless together.

src/util/vehicle.cpp
```diff
@@ -144,7 +144,28 @@ namespace big::vehicle
...
+		VEHICLE::SET_VEHICLE_NEEDS_TO_BE_HOTWIRED(veh, false);
+
+		if (FIRE::IS_ENTITY_ON_FIRE(veh))
+		{
+			FIRE::STOP_ENTITY_FIRE(veh);
+		}
+
+		VEHICLE::SET_VEHICLE_ENGINE_HEALTH(veh, 1000.0f);
+		VEHICLE::SET_VEHICLE_PETROL_TANK_HEALTH(veh, 1000.0f);
+		VEHICLE::SET_VEHICLE_BODY_HEALTH(veh, 1000.f);
```
gir489returns, line 162: IIRC, SET_VEHICLE_FIXED does all of this? Check to confirm.

gir489returns (changes requested):
Tested and working, but...

rkwapisz:
@xiaoxiao921 nailed it with the 0xD8 offset. Well done.
private int _oVState = 0xD8; // int 0:Player 1:NPC 2:Unused 3:Destroyed

Don't need anything else after that besides VEHICLE::SET_VEHICLE_FIXED(veh)

rkwapisz:
I think this concept could be easily expanded to repairs from any sort of destructive event such as explosions. Would be handy since a lot of mod menus can just explode your vehicle, and not everyone likes playing with godmode.

If vehicle repair can be refactored to remember the player's last vehicle (like we can with GET_VEHICLE_PED_IS_IN(ped, lastVehicle = true), then the vehicle could be repaired even without the player having to physically be inside it. The bitmask could probably work for that case as well to recover the vehicle from any destroyed state, including a smouldering heap.

Then repair_engine_from_water could be renamed as something like repair_destroyed_vehicle. Would be great to be able to quickly repair a vehicle that was destroyed via explosion griefing without having to recall it from the PV list or call Mors.

gir489returns (approved):

Yimura (approved):

PR #3003, opened 2024-04-29
Feature: Infinite Vehicle Ammo

### LINKED ISSUE #1021: [Request]: Infinite Ammo For Vehicles
### Problem

Hi, Please Add Infinite Ammo For Vehicles.
For example, the Oppressor Mk II is limited to 20 Missiles and will not fire after 20 Missiles.
Or Vigilante is Limited to 30 Missiles and will not fire after 30 Missiles.

### Solution

.

### Reason

Infinite ammo does not work in the weapon section for vehicles.

### Additional context

_No response_

### LINKED ISSUE #1637: [Request]: infinite vehicle rockets
### Problem

_No response_

### Solution

as the tittle suggests,would be cool to have an infinite rocket/barrage option on the menu

### Reason

because im tired of missing my mk1 shots

### Additional context

_No response_

rkwapisz:
This PR adds a new "Infinite Vehicle Ammo" checkbox to the vehicle menu.
When enabled, any weapons in the player's vehicle, such as Oppressor Mk2 missiles, will have infinite ammo. This includes countermeasures and bombs.

The main difference between this and previous implementation attempts is that SET_VEHICLE_WEAPON_RESTRICTED_AMMO doesn't use weapon hashes and instead simply accesses the vehicle's weapon index.

This is a better long-term solution than trying to directly access a vehicle's weapon hashes since it does not require checks against potential new weapon hashes with future GTAO content updates.

en_us translation forthcoming.

Closes #1021 and #1637.

### DIFF
--- a/src/backend/looped/vehicle/infinite_veh_ammo.cpp
+++ b/src/backend/looped/vehicle/infinite_veh_ammo.cpp
@@ -1,0 +1,44 @@
+#include "backend/looped_command.hpp"
+#include "natives.hpp"
+
+namespace big
+{
+	class infinite_veh_ammo : looped_command
+	{
+		using looped_command::looped_command;
+
+		virtual void on_tick() override
+        {
+            Vehicle player_veh = self::veh;
+
+            if (player_veh != 0 && VEHICLE::DOES_VEHICLE_HAVE_WEAPONS(player_veh))
+            {
+                for (int i = 0; i < 3; i++)
+                {
+                    VEHICLE::SET_VEHICLE_WEAPON_RESTRICTED_AMMO(player_veh, i, -1);
+                }
+
+                VEHICLE::SET_VEHICLE_BOMB_AMMO(player_veh, -1);
+                VEHICLE::SET_VEHICLE_COUNTERMEASURE_AMMO(player_veh, -1);
+            }
+        }
+
+		virtual void on_disable() override
+		{
+			Vehicle player_veh = self::veh;
+
+			if (player_veh != 0 && VEHICLE::DOES_VEHICLE_HAVE_WEAPONS(player_veh))
+			{
+				for (int i = 0; i < 3; i++)
+				{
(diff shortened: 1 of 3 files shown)


gir489returns:
Don't we already have this?

lonelybud:
should also close #1637

rkwapisz:
> Don't we already have this?

Surprisingly, no. Not for vehicle weapons. WEAPON::SET_PED_INFINITE_AMMO does not apply to vehicle weapons.
#1060 went stale and was never committed.

src/backend/looped/vehicle/infinite_veh_ammo.cpp
```diff
@@ -0,0 +1,40 @@
...
+            		}
+        	}
+
+		virtual void on_disable() override
+		{
+			if (self::veh && VEHICLE::DOES_VEHICLE_HAVE_WEAPONS(self::veh))
+			{
+				for (int i = 0; i < 3; i++)
+				{
+					VEHICLE::SET_VEHICLE_WEAPON_RESTRICTED_AMMO(self::veh, i, 20);
```
Yimura, line 30: Are they always capped at `20`?

rkwapisz: No. I chose 20 because that's the missile capacity of the Op Mk2, but it's not necessarily the capacity of other vehicle weapon ammo. For example, the Deluxo can hold up to 30. I wasn't able to find a convenient method of pulling the vehicle weapon's ammo capacity from the index, which is what I'd prefer to do for a "Refill Vehicle Ammo" feature.

We could potentially set ammo capacity based on the vehicle itself, but that would require maintaining a database of all vehicles with weapons.

I think that just giving a flat quantity of 20 to all vehicle weapons after disabling Vehicle Infinite Ammo is a good compromise.

Yimura: Alright that's fine by me.

Yimura (approved):

PR #2818, opened 2024-03-11
Nightly but better

xiaoxiao921:
Closes #2787

PR #3022, opened 2024-05-02
Add .pdb file to ci workflow

pierrelasse:
This allows the download of the .pdb file from pull requests.


gir489returns:
Duplicate of #2818?

xiaoxiao921:
Not duplicate, but I'd rather have the github bot comment being modified instead with two links, one with just the binary.zip and one with binary_and_pdb.zip. Is that possible with the github action?

gir489returns:
Why not just upload one zip for the DLL and one zip for the PDB?

pierrelasse:
Why zip the files individually? Not that they're a few GB.

xiaoxiao921:
Because I don't want to download the pdb if I don't want it, simple as that.

Yimura:
Yeah that's what I was expecting, the build isn't even creating a PDB to begin with.

xiaoxiao921:
For finishing this PR, do the following change:
```
      - name: Generate CMake project
        run: cmake -D CMAKE_BUILD_TYPE=RelWithDebInfo -D OPTIMIZE=YES -S. -Bbuild -G Ninja

      - name: Build 64bit release DLL
        run: cmake --build ./build --config RelWithDebInfo --target YimMenu --
```
       
and

```
      - name: Upload Artifact
        uses: actions/upload-artifact@v3
        with:
          name: dll
          path: |
              path_to_dll
``` 

```
      - name: Upload PDB Artifact
        uses: actions/upload-artifact@v3
        with:
          name: pdb
          path: |
            path_to_pdb
```

Yimura:
Indeed for CI we don't really need a Release build as it's really just meant for debugging.

Though I do agree with @xiaoxiao921's comment on separating the two as users/tests don't necessarily need the pdb file.

xiaoxiao921:
The fact that the comment bot use Release instead of RelWithDebInfo is more of a bug than anything. the nighlty already use RelWithDebInfo so that the pdb is generated inside the artifact.

pierrelasse:
Please feel free everyone to edit the workflow to put it into two separate zips or close this PR.

gir489returns:
We're not going to do your work for you.

xiaoxiao921:
Pretty sure what I gave can be copy pasted and it'll work

PR #3024, opened 2024-05-02
Add more translation key

sch-lda:


Rxann (commented):
No major issues that jumped out at me while looking at it.

Yimura (approved):

Yimura (approved):

PR #3052, opened 2024-05-04
Finished 3 TODO`s

timdev-ger:
ImageButton - imgui.hpp
Image - imgui.hpp

system.hpp


gir489returns (approved):
System looks fine to me, xiao needs to approve the Lua changes.

xiaoxiao921 (dismissed):

xiaoxiao921:
Approved the lua change but realized this wasnt even tested for compilation, tempting to just close this

src/util/system.hpp
```diff
@@ -5,30 +5,23 @@
...
 {
-	inline void dump_entry_points()
-	{
-		// TODO!
-#if 0
-		DWORD64 base_address = memory::module("GTA5.exe").begin().as<DWORD64>();
+    inline void dump_entry_points()
+    {
+        const auto file_path = g_file_manager.get_project_file("./entrypoints.txt");
+        std::ofstream file(file_path.get_path(), std::ios::out | std::ios::trunc);
```
maybegreat48: This should stay disabled since there is no real crossmap anymore. The invoker now uses indices that map directly to translated hashes

Yimura (changes requested):
Please take a look at the compiler output, there appears to be a missing include that's required to make this change working.

src/util/system.hpp
maybegreat48: Just revert this and the code would compile

timdev-ger:
Backuped the system.hpp due an error.

As @maybegreat48 already said, there is no real crossmap anymore.
This was my mistake. It should compile now.

Yimura:
Did you compile this on your system before pushing? There are still compilation issues with the arguments being passed through.

gir489returns:
@Yimura These changes were done on GitHub. 

image

You can also tell because the stupid GitHub editor demands that the EOL character change for the EOF.

gir489returns:
ImGui::Image and ImGui::ImageButton require passing a ImTextureID parameter, not just the "string" of the texture ID. This is quite an involved process to get working, and you can see on the ImGui documentation, this is probably something we want to remove from the Lua layer: https://github.com/ocornut/imgui/blob/master/docs/FAQ.md#q-how-can-i-display-an-image-what-is-imtextureid-how-does-it-work

PR #3053, opened 2024-05-05
PlayerDB Delete Untrusted Players

rkwapisz:
As a player, I currently enjoy the option to clear the Player Database of player entries that are marked as modders. YimMenu adds these entries automatically as reactions to events such as attacking players with god mode, superjump, kick/crash attempts, and more.

However, I also use the Player Database to keep track of trusted players and friends, especially since I find it much easier to track and join friends using YimMenu rather than the in-game browser. Finding friends in the Player Database can become difficult as it fills up, and I don't want to delete friends with the Remove All option.

This PR adds another button called "Remove Untrusted" which removes any players from the Player Database who are not marked "Trust". This allows players like me to enjoy the advantages of YimMenu's player database with friends and other trusted players, without having to navigate through large lists of modders every time.

### DIFF
--- a/src/services/player_database/player_database_service.cpp
+++ b/src/services/player_database/player_database_service.cpp
@@ -219,6 +219,33 @@ namespace big
 		return player;
 	}
 
+	void player_database_service::remove_untrusted_players()
+	{
+		for (auto it = m_players.begin(); it != m_players.end();)
+		{
+			if (!it->second->is_trusted)
+			{
+				it = m_players.erase(it);
+			}
+			else
+			{
+				++it;
+			}
+		}
+
+		for (auto it = m_sorted_players.begin(); it != m_sorted_players.end();)
+		{
+			if (!it->second->is_trusted)
+			{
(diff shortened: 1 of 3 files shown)


rkwapisz:
Teeny tiny GUI update:
image

ghost:
Perhaps we should add the ability to remove players from other categories?  adding a "delete" button and a drop-down list with categories such as modder, untrusted, blocked, etc

rkwapisz:
> Perhaps we should add the ability to remove players from other categories? adding a "delete" button and a drop-down list with categories such as modder, untrusted, blocked, etc

I'm not trying to implement SQL here. I just want the ability to easily differentiate between clean players and modders.
At some point in the future I may revisit this and see about adding an entirely separate menu for friends, but for now I enjoy being able to clear the DB of modders because I generally don't keep that information for longer than a few sessions. Nobody's bothered me to the point where I need them on a permanent shitlist.

Yimura (approved):

PR #3056, opened 2024-05-05
Kick order changed

ghost:
I think it would be wiser to first send a desync kick that is not detected by others and only then other methods


gir489returns:
Desync kick takes a couple of seconds to activate as it fills the buffer, the order of operations is irrelevant.

ghost:
> Desync kick takes a couple of seconds to activate as it fills the buffer, the order of operations is irrelevant.

then maybe it’s worth just removing nfkick oomkick endkick and leaving only desync since they don’t work anyway and everyone has protection from them and it’s only a waste of time to trigger their protection? Well, of course, leave the host methods

rkwapisz:
I think this is a good commit. Host and desync kicks are, in my experience, the only reliable kicks in YimMenu.

Any reason to keep the others?

ghost:
It's always better to have too much weapons than not enough. Also, there is a lot of work behind these kicks, it would be a shame to just remove them like this.

ghost:
> It's always better to have too much weapons than not enough. Also, there is a lot of work behind these kicks, it would be a shame to just remove them like this.

kicks will not be deleted anywhere except the kick reaction.  The fact is that the reaction is used mainly on the modder automatically and so that he cannot find out who kicked him, it is better to leave only the desync kick.  and from other kicking methods he has 99% protection from them and he finds out who tried to kick and maybe the reaction to kick me will also work for him and as a result, his desync kick worked earlier and I got out of sync (sry for my bad england)

rkwapisz:
> It's always better to have too much weapons than not enough. Also, there is a lot of work behind these kicks, it would be a shame to just remove them like this.

Absolutely not. There is no point in maintaining support for things that no longer work.
And, as I've said above, every menu blocks these other kicks.

xiaoxiao921:
> > It's always better to have too much weapons than not enough. Also, there is a lot of work behind these kicks, it would be a shame to just remove them like this.
> 
> Absolutely not. There is no point in maintaining support for things that no longer work. And, as I've said above, every menu blocks these other kicks.

I totally agree, time spent on a feature doesn't matter, only the end result / product matter, and if things doesn't work or are useless they should be trimmed off the code base.

rkwapisz:
> > It's always better to have too much weapons than not enough. Also, there is a lot of work behind these kicks, it would be a shame to just remove them like this.
> 
> kicks will not be deleted anywhere except the kick reaction. The fact is that the reaction is used mainly on the modder automatically and so that he cannot find out who kicked him, it is better to leave only the desync kick. and from other kicking methods he has 99% protection from them and he finds out who tried to kick and maybe the reaction to kick me will also work for him and as a result, his desync kick worked earlier and I got out of sync (sry for my bad england)

If we're getting rid of certain kicks as a reaction we should remove those kicks from the player menu as well, assuming we're removing those kicks because they don't do the job.

maybegreat48:
I mean the desync kick almost always works, while the other kicks can be blocked. The other kicks are much faster though, so the desync is essentially the last resort if the modder doesn't leave. Now, you can be karma'd or crashed, but this is generally rare. 

> since they don’t work anyway and everyone has protection from them

There are still popular menus that do not/cannot block some of these kicks such as Kiddions and several pastes

rkwapisz:
> I mean the desync kick almost always works, while the other kicks can be blocked. The other kicks are much faster though, so the desync is essentially the last resort if the modder doesn't leave. Now, you can be karma'd or crashed, but this is generally rare.
> 
> > since they don’t work anyway and everyone has protection from them
> 
> There are still popular menus that do not/cannot block some of these kicks such as Kiddions and several pastes

I prefer silently kicking people if they're causing problems. I can't always know what menu someone is using, and I don't want to risk failing a particular kick that gets me flagged to them and possibly have my RID posted to some P2C menu's Discord somewhere. Some players are so salty and toxic that they'll just follow you around from session to session and, if they can't crash you, they'll just grief everyone else.

So is there more benefit to an instant, detectable, and blockable kick versus one that is slightly delayed but undetectable and unblockable? I'll go with the latter 100% of the time. For me, kicks are a defensive tool so I want to be as protected and anonymous as possible when using one.

rkwapisz:
That being said, you can't desync kick a host so without the other kicks available there's no way for Kick Host on Join to work.

But I'd argue that's a super toxic feature and shouldn't be in YimMenu.

lonelybud:
> Desync kick takes a couple of seconds to activate as it fills the buffer, the order of operations is irrelevant.

i think this should have just closed the pr. its up to the user whatever he wants to do. I believe `multi kick` was invented for lazy ones no offense.

Edit:  As other said their are menus like kiddion still unprotected by existing kicks.

PR #3061, opened 2024-05-06
Remove useless kicks

tiCkuhAr:
Similar to https://github.com/YimMenu/YimMenu/pull/3056, however, I disagree with how they wanted to do it; removing all the other kicks would make Multi kick unable to kick the session host. It would not even be able to kick a normal player with host, let alone a modder, which would completely break auto kick host. Also, the other kicks still work on some menus, like modest. I removed the two most blocked kicks. I don't even know if bail kick even worked anymore. The last time I tried it, nothing happened. I did not even find a reference to it in the code  other than for Lua.


tiCkuhAr:
I left NF Kick for people that want to use it in their Lua scripts, as it still works and seems to be the only other kick that supports kick all.

rkwapisz:
Regardless of my own position on kicks, there's definitely no sense in keeping non-working features in the code. If it's plain broken, get rid of it.

lonelybud:
> Regardless of my own position on kicks, there's definitely no sense in keeping non-working features in the code. If it's plain broken, get rid of it.

just keep that sweet oom & desync kick and remove the rest?

tiCkuhAr:
That's what I was thinking about doing.

maybegreat48:
Don't have a strong opinion on this, but I guess null function can go

tiCkuhAr:
Alright, I removed it. I only left it because it's the only kick that supports kick all for people making lua scripts.

maybegreat48:
You can definitely add an end kick all option by modifying the [net array](https://github.com/maybegreat48/YimMenu/blob/main/src/hooks/toxic/broadcast_net_array.cpp#L20) hook. Just be sure to disable it after 10 seconds or so to avoid kicking joining players

tiCkuhAr:
It's not a big deal. I was just saying that since null function kick was already part of the menu and supported kick all for Lua, I was going to leave it for the Lua devs.

Anyway, this should be ready to be merged, unless you want something changed.

ghost:
Can you add a check that if the player is the host, then kick him with an oom kick? Thus, if the player is not the host, then we can kick him without triggering his reaction and he will not understand who kicked him. I hope you understood what I meant. Something like this:
`
 if (player && player->is_host() && player->is_valid())
	          {
		          dynamic_cast<player_command*>(command::get("oomkick"_J))->call(player, {});
		          script::get_current()->yield(700ms);
	          }
`

tiCkuhAr:
Due to the fact that it sometimes backfires, I would like it to just be there as a last resort. However, if one of the YimMenu maintainers thinks it should be the only kick for people that are not host, then I will change it to that.

Rxann:
> Due to the fact that it sometimes backfires, I would like it to just be there as a last resort. However, if one of the YimMenu maintainers thinks it should be the only kick for people that are not host, then I will change it to that.

Ever since maybegreat48 improved the oom, i don't think it really backfires anymore. Not in my experience atleast and i use it a LOT

rkwapisz:
> Due to the fact that it sometimes backfires, I would like it to just be there as a last resort. However, if one of the YimMenu maintainers thinks it should be the only kick for people that are not host, then I will change it to that.

Maybe we should clarify a bit more when we want to use a particular kick? Admit I'm starting to lose track. Though this logic primarily affects multi-kick, we should still add a host check for desync so users don't flag themselves trying to desync a host (it'll never work).

- Host kick: default and first kick option when YOU are the host, so if you're host, multikick == host kick only
- OOM kick: default and first kick option when ANOTHER PLAYER is the host, so if they're host, multikick == OOM kick only
- desync kick: default and first kick option when NEITHER YOU NOR THE PLAYER are the host, so multikick == desync + OOM

Make sense?

ghost:
> > Due to the fact that it sometimes backfires, I would like it to just be there as a last resort. However, if one of the YimMenu maintainers thinks it should be the only kick for people that are not host, then I will change it to that.
> 
> Maybe we should clarify a bit more when we want to use a particular kick? Admit I'm starting to lose track. Though this logic primarily affects multi-kick, we should still add a host check for desync so users don't flag themselves trying to desync a host (it'll never work).
> 
> * Host kick: default and first kick option when YOU are the host, so if you're host, multikick == host kick only
> * OOM kick: default and first kick option when ANOTHER PLAYER is the host, so if they're host, multikick == OOM kick only
> * desync kick: default and first kick option when NEITHER YOU NOR THE PLAYER are the host, so multikick == desync + OOM
> 
> Make sense?

For a player who is not a host, a desync kick will be enough to avoid triggering his reaction.  oom leave exclusively against hosts

Yimura (approved):

PR #3063, opened 2024-05-06
Spam in chat as moder detection

### LINKED ISSUE #3026: [Improvement]: Add more options to react to spammers
### Which feature

As far as I understand, we can only kick the chat of spammers, but there is no way, for example, to add them to the database, set a timeout, block joining, etc. I think it’s not difficult to add a “chat spam” section to the reaction and the player will independently choose what to do. And remove the checkbox for automatic kicking of spammers in the Network section and set it as default in reactions.

### Reason

I think it would be more logical.

### Additional context

Or, to avoid the hassle altogether, move it to the Reactions/Moder Detection section.

ghost:
Closes #3026
I don’t think there’s any point in making a separate reaction, just adding it as a modder-detection


gir489returns:
I'm not opposed to this, but this could start to make YimMenu clients suspectial to retaliation attacks from those automated chat spammer bots that just join the lobby start spamming and leave. Since Rockstar has upped their spam detection, normal people don't see it, only we do, so this would turn YimMenu into an offensive menu instead of a defense one, and that I am against.

src/core/data/infractions.hpp
```diff
@@ -18,6 +18,7 @@ namespace big
 		ATTACKING_WITH_GODMODE,
 		ATTACKING_WITH_INVISIBILITY,
 		ATTACKING_WHEN_HIDDEN_FROM_PLAYER_LIST,
+		CHAT_SPAM,
```
GreySideMM: Don't ever change the order of this enum, it will break the already saved database. Always append.

rkwapisz:
Oh god if we start infracting chat spammers then our player DBs are gonna instantly fill up within a half hour of joining any session. Those bots are everywhere.

I also really don't care about them. I rarely see their messages. If anything, YimMenu spams me more with chat spammer notifications than the actual chat spammers. :) They're otherwise harmless.

lonelybud:
if you join Asian lobby there are 50%+ chat spammers in one session and they hog up sessions. Also now days they are spoofing tokens and disturbing session even more. There is no point of kicking them but we can block them via `handle_join_request` hook once we have them in db

PR #3065, opened 2024-05-06
[Workaround]: Remove Start LS Customs Button from LS Customs Menu

rkwapisz:
After making changes to a vehicle via the LSC script, the game stops responding to mouse and keyboard events. It gets stuck in some sort of transition phase between LSC and freemode. Maybe it's waiting for the drive-out cutscene to finish before returning player control? I'm not sure, but it's 100% reproducible.

Unless someone wants to take a crack at fixing this script, I think we should prevent users from activating it since they're spending actual GTA$.

### DIFF
--- a/src/views/debug/view_debug_misc.cpp
+++ b/src/views/debug/view_debug_misc.cpp
@@ -60,6 +60,10 @@ namespace big
 				DLC::ON_ENTER_SP();
 			});
 
+			components::button("START_LS_CUSTOMS"_T, [] {
+				g.vehicle.ls_customs = true;
+			});
+
 			components::button("SKIP_CUTSCENE"_T, [] {
 				CUTSCENE::STOP_CUTSCENE_IMMEDIATELY();
 			});
--- a/src/views/vehicle/view_lsc.cpp
+++ b/src/views/vehicle/view_lsc.cpp
@@ -192,7 +192,15 @@ namespace big
 			});
 		}
 
+		// LSC script is broken, release builds should hide the button from users but we will keep it for debugging
+		#ifndef NDEBUG
+		components::button("START_LS_CUSTOMS"_T, [] {
+			g.vehicle.ls_customs = true;
+		});
+
 		ImGui::SameLine();
+		#endif
+
 		if (components::button("MAX_VEHICLE"_T))
 		{
 			g_fiber_pool->queue_job([] {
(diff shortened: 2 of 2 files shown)


Yimura (dismissed):
Yeah for now this is the correct way of handling this.

Long ago this worked beautifully however R* made some changes causing the script to become significantly harder to trigger and use properly.
Ideally if possible another effort should be made to redo this feature by analyzing the scripts and trigger it without doing janky script patches (which are hard to maintain).

T14D3:
Suiciding works to regain control, and for some use cases the button is pretty handy, so I'd prefer if we just either put a big warning label on it or move it for example to the "Debug" section

Yimura:
> Suiciding works to regain control, and for some use cases the button is pretty handy, so I'd prefer if we just either put a big warning label on it or move it for example to the "Debug" section

Yes indeed however that shouldn't be the solution to this, ideally it would be reworked entirely.

Also I'm interested in these use cases where you'd consider locking your game up would be useful.

T14D3:
> > Suiciding works to regain control, and for some use cases the button is pretty handy, so I'd prefer if we just either put a big warning label on it or move it for example to the "Debug" section
> 
> Yes indeed however that shouldn't be the solution to this, ideally it would be reworked entirely.

Properly functioning rework would definitely be better, yes, but I (personally) feel like making such a function harder to find is better than just outright removing them.

As for use cases, It sometimes works when you are in the bounds of an actual LSC with a vehicle that is too big for the LSC to normally accept it - rare, and highly not stable/repeatable, but IMO enough reason to keep it _somewhere_

Yimura:
I'm fine with moving it to debug, @rkwapisz I'm fine with either way. If you could move it to debug somewhere I'd appreciate it.

Yimura (dismissed):

rkwapisz:
I think I misinterpreted the ask. I wrapped the button in a NDEBUG check, but I think what you asked for was to put the button inside the actual YimMenu debug window.

Yimura:
> I think I misinterpreted the ask. I wrapped the button in a NDEBUG check, but I think what you asked for was to put the button inside the actual YimMenu debug window.

Yeah indeed, as debug isn't really useful unless we listen to the asserts that ImGui is throwing at us and fix them.

Yimura (approved):

PR #3066, opened 2024-05-07
Less noticeable multi-kick

ghost:
this way the oom kick will only work on hosts, and in all other cases a less noticeable desync kick will be used


rkwapisz (commented):
`player->is_host() && player->is_valid()`

This is redundant.

ghost:
> `player->is_host() && player->is_valid()`
> 
> This is redundant.

sometimes it happened that the player managed to leave the session before the if was triggered and a crash occurred.  I think it won't be superfluous

gir489returns:
> > `player->is_host() && player->is_valid()`
> > This is redundant.
> 
> sometimes it happened that the player managed to leave the session before the if was triggered and a crash occurred. I think it won't be superfluous

No it won't. The scope for the delay is inside the if check, the conditions for the if check have already been satisfied and are reduplicated on the subsequent line. He is correct.

ghost:
> `player->is_host() && player->is_valid()`
> 
> This is redundant.

`player->is_host()` is not superfluous.  for example, a scenario when I and the player are not the host, then oom and desync were launched, but I don’t need it.  I only need to run desync so I added a check for the host

lonelybud:
is it possible to desync someone when host. host kick is bad in terms it will make others like voted out when kicked once.

lonelybud:
what about `gta_util::get_network_player_mgr()->RemovePlayer(selected_player->get_net_game_player());` to allow them to come back

ghost:
> is it possible to desync someone when host. host kick is bad in terms it will make others like voted out when kicked once.

nah we cannot do desync when we are the host

ghost:
> what about `gta_util::get_network_player_mgr()->RemovePlayer(selected_player->get_net_game_player());` to allow them to come back

If we kick out a player, I don't think we should allow him to join again.

lonelybud:
> `gta_util::get_network_player_mgr()->RemovePlayer(selected_player->get_net_game_player());`

it works. player can rejoin an wont see voted out message

gir489returns:
> > what about `gta_util::get_network_player_mgr()->RemovePlayer(selected_player->get_net_game_player());` to allow them to come back
> 
> If we kick out a player, I don't think we should allow him to join again.

That is not for you to decide. It is up to the user to make that decision.

ghost:
I don't understand what you're talking about. If you mean that host-kick causes issues with reconnecting, then I removed it and left breakup.

lonelybud:
look `oom kick` is detected and blocked. Only kick you are left with is host kick only when you say try to kick yimmenu user for example. Host kick when host. Desync when not host. All other kicks detected.

As said before multi-kick is for lazy people who don't know what to use. 

Edit: one should easily kick god mod players by any kick who are detected by `modder detection` reaction

xiaoxiao921:
Honestly a single kick button should be in the menu, if any of the kick can backfire then this is terrible user experience.

ghost:
> look `oom kick` is detected and blocked. Only kick you are left with is host kick only when you say try to kick yimmenu user for example. Host kick when host. Desync when not host. All other kicks detected.
> 
> As said before multi-kick is for lazy people who don't know what to use.

I understand, but I just wanted to make it so that, for example, the multi-kick we use as a reaction was not noticeable to the victim, or at least to those victims who are not the host.

lonelybud:
> I understand, but I just wanted to make it so that, for example, the multi-kick we use as a reaction was not noticeable to the victim, or at least to those victims who are not the host.

when host if oom kick is used it shall be blocked and may backfire. Host kick will do the job. When not host you left with desync and oom. Now oom is still detectable and can backfire, desync is fine but again desync can desync you back.

Edit: dont host kick your friends they will not be able to join back

ghost:
> > I understand, but I just wanted to make it so that, for example, the multi-kick we use as a reaction was not noticeable to the victim, or at least to those victims who are not the host.
> 
> when host if oom kick is used it shall be blocked and may backfire. Host kick will do the job. When not host you left with desync and oom. Now oom is still detectable and can backfire, desync is fine but again desync can desync you back.
> 
> Edit: dont host kick your friends they will not be able to join back

Indeed, but 'oom-kick' is the only way to kick the host and will be used solely for that purpose. In other cases, we'll resort to 'breakup' or 'desync'. We won't use host-kick if there are issues with players reconnecting.

lonelybud:
> Indeed, but 'oom-kick' is the only way to kick the host and will be used solely for that purpose. In other cases, we'll resort to 'breakup' or 'desync'. We won't use host-kick if there are issues with players reconnecting.

not sure about `breakup` whether it is detected by P2Cs like stand

ghost:
> player->is_valid()

I'm not sure about this, but the breakup-kick seems more acceptable; there won't be any issues with reconnecting. Especially when we're hosts, modders won't be able to do anything to us. I still wanted to clarify about checking for `player && player->is_valid()` wouldn't it result in crashes if the check is absent, similar to what happened before #2106? If everything is fine, then it's possible to merge

rkwapisz:
> I still wanted to clarify about checking for `player && player->is_valid()` wouldn't it result in crashes if the check is absent, similar to what happened before #2106? If everything is fine, then it's possible to merge

You still want the check, and the check itself was fixed in https://github.com/YimMenu/YimMenu/pull/3068.

src/backend/commands/player/kick/multi_kick.cpp
```diff
@@ -18,12 +18,12 @@ namespace big
...
 		{
 			if (g_player_service->get_self()->is_host())
-				dynamic_cast<player_command*>(command::get("breakup"_J))->call(player, {}),
-				    dynamic_cast<player_command*>(command::get("hostkick"_J))->call(player, {});
-
-			dynamic_cast<player_command*>(command::get("oomkick"_J))->call(player, {});
-			script::get_current()->yield(700ms);
+				dynamic_cast<player_command*>(command::get("breakup"_J))->call(player, {});
 
+			if (player->is_host()) {
```
rkwapisz: You should still check player. You just don't need the is_valid() check.

`if (player && player->is_host())`

src/backend/commands/player/kick/multi_kick.cpp
```diff
@@ -18,12 +18,12 @@ namespace big
...
+				dynamic_cast<player_command*>(command::get("oomkick"_J))->call(player, {});
+				script::get_current()->yield(700ms);
```
rkwapisz: Why are we yielding? If we're oom-kicking, then we won't be following up with a desync because you're already verified that the kick target is the host, so your next check for !player->is_host() will return false and therefore never call the desync.

src/backend/commands/player/kick/multi_kick.cpp
```diff
@@ -18,12 +18,12 @@ namespace big
...
+			}
 
 			if (player && !g_player_service->get_self()->is_host() && player->is_valid() && !player->is_host())
```
rkwapisz: I'd recommend re-ordering the checks so they're a bit easier to follow.
First check that we're not the host, then check that player is not null, then if they're valid, then if they're not host.

if (!g_player_service->get_self()->is_host() && player && player->is_valid() && !player->is_host())

src/backend/commands/player/kick/multi_kick.cpp
```diff
@@ -18,14 +18,12 @@ namespace big
...
+				dynamic_cast<player_command*>(command::get("breakup"_J))->call(player, {});
 
-			dynamic_cast<player_command*>(command::get("oomkick"_J))->call(player, {});
-			script::get_current()->yield(700ms);
+			if (!g_player_service->get_self()->is_host() && player && player->is_valid() && player->is_host())
+				dynamic_cast<player_command*>(command::get("oomkick"_J))->call(player, {});
 
-
-			if (player && !g_player_service->get_self()->is_host() && player->is_valid() && !player->is_host())
+			if (!g_player_service->get_self()->is_host() && player && player->is_valid() && !player->is_host())
```
Yimura: 3/4th of the condition in these two if conditions are duplicates.

This whole function can be simplified to the following:
```cpp
if (g_player_service->get_self()->is_host())
{
	dynamic_cast<player_command*>(command::get("breakup"_J))->call(player, {});
}
else if (player && player->is_valid())
{
	if (player->is_host())
		dynamic_cast<player_command*>(command::get("oomkick"_J))->call(player, {});
	else
		dynamic_cast<player_command*>(command::get("desync"_J))->call(player, {});
}
```

Yimura (changes requested):

lonelybud:
- rename `multi_kick` to `smart_kick` and add these changes
```C++
if (g_player_service->get_self()->is_host())
	dynamic_cast<player_command*>(command::get("breakup"_J))->call(player, {});
else if (player->is_host())
	dynamic_cast<player_command*>(command::get("oomkick"_J))->call(player, {});
else
	dynamic_cast<player_command*>(command::get("desync"_J))->call(player, {});
```

- add `player && player->is_valid()` checks in individual kick files
- add one more kick called `block kick` which should add player in player database with block true and does `host kick` or `desync kick` depending on whether you host or not

ghost:
> * rename `multi_kick` to `smart_kick` and add these changes
> 
> ```c++
> if (g_player_service->get_self()->is_host())
> 	dynamic_cast<player_command*>(command::get("breakup"_J))->call(player, {});
> else if (player->is_host())
> 	dynamic_cast<player_command*>(command::get("oomkick"_J))->call(player, {});
> else
> 	dynamic_cast<player_command*>(command::get("desync"_J))->call(player, {});
> ```
> 
> * add `player && player->is_valid()` checks in individual kick files
> * add one more kick called `block kick` which should add player in player database with block true and does `host kick` or `desync kick` depending on whether you host or not

Agreed, but why do we need another kick if we're already sending a multi-kick as a reaction?

lonelybud:
> Agreed, but why do we need another kick if we're already sending a multi-kick as a reaction?

no need i guess for now

ghost:
can we merge this or should we wait?

rkwapisz:
Desync kick was reworked in https://github.com/YimMenu/YimMenu/pull/3073 so make sure you're not reverting the author's changes.

ghost:
> Desync kick was reworked in #3073 so make sure you're not reverting the author's changes.

It doesn't revert

rkwapisz (commented):
Looks good to me. I like the rename of multi kick to smart kick.

Yimura (approved):

PR #3070, opened 2024-05-08
Refactor of Player Wanted Level

### LINKED ISSUE #3040: [Bug]: Has anyone encountered the wanted level bug?
### Describe the bug

I once dragged the police wanted level slider, first to 5 and then to zero. And now all the time when I restart the game or re-inject anti-cheat, the police do not react to me at all. Unless I get into someone’s car who is already wanted by the police.

### Steps To Reproduce

1. Move the slider from 0 to any value
2. Play for a while and return to zero and your wanted level will disappear.
3. Restart the game and inject anti-cheat and the police will not respond to you.

### Expected Behavior

I think it’s worth adding the ability to hide the slider, for example if you add a checkbox “Set a specific wanted level” or something like that and then the slider appears. If the checkbox is inactive, it will be possible to come to the attention of the police. It seems to me that this is simply saved to the settings file, but the checkbox itself is inactive. Clicking the "Clear Wanted Level" button does not help.

### Logs

No logs sry(

### Screenshots / Video

pic1

### Store

Steam

### Language

Russian

### YimMenu Language

Русский (RU)

### Additional context

_No response_

### Build

### Coload

rkwapisz:
This PR is for an almost total refactor of the Never Wanted feature in the Self menu, including the button that brought up some additional Wanted Level manipulation options.

IMO the previous implementation was unclear, self-conflicting, and poorly coded and documented. This PR replaces all that with code that provides the following capabilities (mostly familiar, but they all work better alone and with each other):

- A Never Wanted toggle just like the old version, but which overrides and also hides the other elements of the feature that become irrelevant when Never Wanted is enabled
- A Clear Wanted Level button just like the old version, but which resets a Wanted Level to 0 even if the Force Wanted Level slider (described below) is configured to a non-zero Wanted Level value
- A Wanted Level slider that behaves better than the old version. This version is much more intuitive, and can be used to non-permanently set the player's Wanted Level which means the Wanted Level can continue to naturally increase or decay with the player's actions. The slider also updates to reflect the player's actual Wanted Level if it's not being forced by Never Wanted or Force Wanted Level, so the Wanted Level shown in the slider should always be consistent with the game.
- A Force Wanted Level checkbox that can be enabled to permanently set or lock the player's Wanted Level to the value specified in the slider. If you want to forever be chased by hungry 5-star tactical teams, go for it.

Although Never Wanted and Clear Wanted could be feasibly handled by the slider and Force Wanted Level toggle themselves, I don't want users to have to learn how to use the slider if they don't have to.

Lots of old code was outright deleted. I don't think anything else used the old police cpp and header files. I have absolutely no idea why the old code was byte patching to restore wanted levels after the menu was unloaded. Everything can be handled by manipulating the Wanted Level nodes in g_local_player->m_player_info. Writing to m_is_wanted after m_wanted_level may be redundant, but it works fine the way it is now. Shrug.

These features now have their own section in the Self menu, which I think is a large improvement from the previous button.

A few new translations will be PR'd tomorrow.

image

image

Should close https://github.com/YimMenu/YimMenu/issues/3040.


Yimura (approved):

Yimura:
Much needed refactoring of this code, one of many things that needed it.

I'm happy someone is taking it up, I'm not too active and usually prefer reworking large and complex pieces of the code base.

PR #3072, opened 2024-05-08
[Hotfix]: Wanted Level

rkwapisz:
Couple things slipped through my fingers the other night that I'm resolving with this PR:

- Never Wanted will now properly reset the slider and force wanted level buttons, which I think is better behavior because we don't want to restore a previously-configured wanted level when turning off Never Wanted
- I renamed the clear wanted level command for self to "clearwantedself" since I failed to notice that the "clearwanted" command was already being used by the player menu's clear wanted option, which broke it
- Also updated the hotkey to properly call the new command

### DIFF
--- a/src/backend/looped/self/wanted_level.cpp
+++ b/src/backend/looped/self/wanted_level.cpp
@@ -26,7 +26,7 @@ namespace big
 		}
 	};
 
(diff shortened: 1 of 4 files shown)


gir489returns (approved):

kozhiro11:
in new update clear wanted level hotkey will crash game

rkwapisz:
> in new update clear wanted level hotkey will crash game

It works fine for me with these changes. Are you using the build with the hotfix?

kozhiro11:
i use yimmenu [b7d75d4] download yesterday

PR #3073, opened 2024-05-09
Assorted fixes

### LINKED ISSUE #2685: [Bug]: Kick Everyone from Interiors
### Describe the bug

Simply pressing the button crashes you.

### Steps To Reproduce

Network > Griefing > Kick Everyone From Interiors

### Expected Behavior

To kick everyone from their apartments?

### Logs and/or screenshots

```
<paste your logs here>
```

### Store

Epic Games

### Language

English

### YimMenu Language

English (US)

### Additional context

_No response_

### Build

### Coload

### LINKED ISSUE #3046: [Improvement]: Send SEXT - Toxic command
### Which feature

Choose player ->Toxic -> SEND SEXT
https://github.com/YimMenu/YimMenu/blob/d9576584f5750a69495e60ba002ea2c7cba9ddf4/src/backend/commands/player/toxic/send_sext.cpp#L14

### Reason

I don't want to be recognized as a modder, I prefer to calmly clean up the lobby and, for example, warn people when they do something wrong, not in the chat, and if there is such an option as SMS, it can be developed.
Obviously, when someone knows that a modder often looks askance wanna money or others ;d or u are attacked from other modders ;d.

Maybe we can do something like "Private messange" and at this SMS on phone add a choosen IMAGE (if it's possible) and input for own text of messange?

Sending an image is supposed to work, but it doesn't. Instead of the player's name, you can add a title of your choice, e.g. "ADMIN" "WARN", or your own nickname (of the person sending the message).

### Additional context

`Sends a random sext (with an image attachment) to the player` but messange was blink
image

### LINKED ISSUE #3062: [Improvement]: Remove looped::system_desync_kick_protection()?
### Which feature

Recent discussions have lead me into investigating our various kicks and protections. In YimMenu we have the following call in backend::loop()

`looped::system_desync_kick_protection();`

Which brings us to YimMenu's complaint handler, and I can see that we're simply setting m_host_tokens_complained to zero every game tick. However, never in my time with YimMenu has desync protection actually worked. If it did something for us in the past, it seems like it no longer does so. Is this something we need to keep looping? Am I misunderstand the purpose of this code?

### Reason

Code maintenance, spending time doing unnecessary things, etc.

### Additional context

_No response_

maybegreat48:
Fixes #3046 
Fixes #2685
Closes #3062 (function used for desync kick)

### DIFF
--- a/src/services/script_patcher/script_patcher_service.cpp
+++ b/src/services/script_patcher/script_patcher_service.cpp
@@ -14,6 +14,9 @@ namespace big
 
 	script_patcher_service::~script_patcher_service()
 	{
+		for (auto& patch : m_script_patches)
+			patch.disable(get_data_for_script(patch.get_script()));
+
 		m_script_data.clear();
 		m_script_patches.clear();
 		g_script_patcher_service = nullptr;
--- a/src/services/script_patcher/script_patcher_service.cpp
+++ b/src/services/script_patcher/script_patcher_service.cpp
@@ -14,9 +14,6 @@ namespace big
 
 	script_patcher_service::~script_patcher_service()
 	{
-		for (auto& patch : m_script_patches)
-			patch.disable(get_data_for_script(patch.get_script()));
-
 		m_script_data.clear();
 		m_script_patches.clear();
 		g_script_patcher_service = nullptr;
--- a/src/util/chat.hpp
+++ b/src/util/chat.hpp
@@ -123,6 +123,9 @@ namespace big::chat
 {
 	inline SpamReason is_text_spam(const char* text, player_ptr player)
 	{
+		if (player->is_friend())
+			return SpamReason::NOT_A_SPAMMER; // don't filter messages from friends
+
 		if (g.session.use_spam_timer)
 		{
 			if (player->last_message_time.has_value())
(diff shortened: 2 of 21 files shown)


src/services/script_patcher/script_patcher_service.cpp
```diff
@@ -14,6 +14,9 @@ namespace big
 
 	script_patcher_service::~script_patcher_service()
 	{
+		for (auto& patch : m_script_patches)
+			patch.disable(get_data_for_script(patch.get_script()));
+
```
gir489returns, line 19: This is the same code pending from #2967. It hasn't been merged, as it currently doesn't solve the issue at hand.

src/util/chat.hpp
```diff
@@ -123,6 +123,9 @@ namespace big::chat
 {
 	inline SpamReason is_text_spam(const char* text, player_ptr player)
 	{
+		if (player->is_friend())
```
gir489returns, line 126: `(plyr->is_trusted || (g.session.trust_friends && plyr->is_friend())))`

gir489returns (changes requested):

Yimura (approved):

PR #3077, opened 2024-05-09
Add Admin Appendage

Rxann:


gir489returns:
Is Admin can be spoofed, saying they're a Rockstar admin is most likely a lie and highly unlikely. That flag is borderline useless.

rkwapisz:
If we don't already infract users with R* admin tags as modders then I think that would be more useful.

Rxann:
> If we don't already infract users with R* admin tags as modders then I think that would be more useful.

What if they are R* admins though.

gir489returns:
> > If we don't already infract users with R* admin tags as modders then I think that would be more useful.
> 
> What if they are R* admins though.

What if I'm a Rockstar admin? Not an argument.

rkwapisz:
> > If we don't already infract users with R* admin tags as modders then I think that would be more useful.
> 
> What if they are R* admins though.

Good question. If we can't tell real admins apart from fake ones then what do we do with the information?

Rxann:
> > > If we don't already infract users with R* admin tags as modders then I think that would be more useful.
> > 
> > 
> > What if they are R* admins though.
> 
> Good question. If we can't tell real admins apart from fake ones then what do we do with the information?

I will be closing this in favor of implementing a check in looped_modder_detection once #3094 is mereged to see if the player with the r* clan logo has the rid in the admin list that I will create. I will close this PR in about 12 hours to see if anyone has any opinions on this.

rkwapisz:
> the rid in the admin list that I will create

Eagerly waiting to see how effective this is.

Rxann:
> > the rid in the admin list that I will create
> 
> Eagerly waiting to see how effective this is.

Yes, I am struggling to figure out a way to get the clan data of a player.

PR #3093, opened 2024-05-11
Fixing invalid sanity-checks

GreySideMM:
There have been some invalid nullptr checks (which apparently even slip throu #3066). Unless I heavily overslept this should fix any possible crashes.

(+ It took me an eternity to not have unwanted whitespace changes on the last lines.....)


xiaoxiao921:
These changes are simply incorrect. Please learn boolean  logic

Yimura:
Yeah these conditions are incorrect, you want the condition to become true ASAP whenever something isn't as it should be.

Making these condition && will make crashes more frequent.

PR #3100, opened 2024-05-12
[Enhancement]: Session Switcher Combo

rkwapisz:
Instead of the "Session Switcher" combo drop-down always just saying "Regions", now it will show which region the user had selected.

image

### DIFF
--- a/src/views/network/view_network.cpp
+++ b/src/views/network/view_network.cpp
@@ -138,7 +138,7 @@ namespace big
 				if (id == eSessionType::LEAVE_ONLINE && gta_util::get_network_player_mgr()->m_player_count == 0) // Don't show a Leave Online option in single player (it actually sends us INTO online)
 					continue;
 
-				ImGui::BeginDisabled(selected_region_index == -1 && id != eSessionType::LEAVE_ONLINE); // Leave online is always enabled
+				ImGui::BeginDisabled(selected_region_index == -1 && id != eSessionType::LEAVE_ONLINE); // Leave Online is always enabled in online sessions since we don't care about the selected region
 				components::selectable(g_translation_service.get_translation(name), false, [&id] {
 					session::join_type(id);
 				});
--- a/src/views/network/view_network.cpp
+++ b/src/views/network/view_network.cpp
@@ -97,11 +97,10 @@ namespace big
 			{
 				for (const auto& region_type : regions)
 				{
-                    components::selectable(region_type.name, *g_pointers->m_gta.m_region_code == region_type.id, [&region_type] 
-                    {
-                        *g_pointers->m_gta.m_region_code = region_type.id;
-                        region_updated = true;
-                    });
+					components::selectable(region_type.name, *g_pointers->m_gta.m_region_code == region_type.id, [&region_type] {
+						*g_pointers->m_gta.m_region_code = region_type.id;
+						region_updated                   = true;
+					});
 				}
 				ImGui::EndCombo();
 			}
--- a/src/views/network/view_network.cpp
+++ b/src/views/network/view_network.cpp
@@ -84,10 +84,11 @@ namespace big
 		if (g_pointers->m_gta.m_region_code == nullptr)
 			return;
 
-		static int selected_region_index     = -1;
-		static bool region_updated = false;
+		static int selected_region_index = -1;
+		static bool region_updated       = false;
 
-        std::string region_str = (selected_region_index == -1) ? "SESSION_SELECT_COMBO"_T.data() : regions[*g_pointers->m_gta.m_region_code].name;
+		std::string region_str =
+		    (selected_region_index == -1) ? "SESSION_SELECT_COMBO"_T.data() : regions[*g_pointers->m_gta.m_region_code].name;
 
 		ImGui::BeginGroup();
 		components::sub_title("SESSION_SELECT"_T);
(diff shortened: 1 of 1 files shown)


src/views/network/view_network.cpp
```diff
@@ -81,18 +81,24 @@ namespace big
...
+			if (ImGui::BeginCombo("##regionswitcher", region_str.c_str()))
 			{
 				for (const auto& region_type : regions)
 				{
 					components::selectable(region_type.name, *g_pointers->m_gta.m_region_code == region_type.id, [&region_type] {
 						*g_pointers->m_gta.m_region_code = region_type.id;
 					});
 				}
+
+				selected_region_index = *g_pointers->m_gta.m_region_code;
```
gir489returns, line 100: You are not compensating for the user not selecting a region here.

src/views/network/view_network.cpp
```diff
@@ -81,18 +81,24 @@ namespace big
...
 	{
+		static int selected_region_index     = -1;
+        std::string region_str = (selected_region_index == -1) ? "REGIONS"_T.data() : regions[*g_pointers->m_gta.m_region_code].name;
+
 		ImGui::BeginGroup();
 		components::sub_title("SESSION_SWITCHER"_T);
 		if (ImGui::BeginListBox("###session_switch", get_listbox_dimensions()))
 		{
-			if (ImGui::BeginCombo("##regionswitcher", "REGIONS"_T.data()))
+			if (ImGui::BeginCombo("##regionswitcher", region_str.c_str()))
```
gir489returns, line 91: This change doesn't make sense to me. There's a drop down box, with no elements near it, besides the join action on the subsequent line. In a perfect UX world, one interaction = one reaction from the software. Since we're not doing that now, we should instead inform the user that selection a new join action will attempt to take them to the new region with a tooltip or something.

gir489returns (changes requested):

src/views/network/view_network.cpp
rkwapisz: You're right. It was causing crashes for me in single player. Updating right now.

src/views/network/view_network.cpp
rkwapisz: I may be misunderstanding the ask here. There is no functional change to the behavior; selecting a region from the drop-down doesn't currently do anything besides set the "destination" for the join options on the subsequent lines.

We're just providing feedback to the user that a selection has been made in the menu, without them having to open the menu again to verify.

rkwapisz: Here is what I'm proposing to guide the user through a typical session join/create workflow.

First-time view:
- The "Select a Region" default drop-down text and tooltip inform the user to take an action
- Until a region is selected, all join/create options are greyed out to prevent behavior that wasn't explicitly intended by the user

image

Once a region is selected from the drop-down, it will remain displayed and the previously greyed-out options will be enabled.

image

I think this is enough to let the user know what needs to be done to execute the feature. Any further thoughts?

rkwapisz (commented):
Compensated for user not selecting a region, plus additional workflow improvements.

src/views/network/view_network.cpp
```diff
@@ -81,22 +81,43 @@ namespace big
...
 				for (const auto& region_type : regions)
 				{
-					components::selectable(region_type.name, *g_pointers->m_gta.m_region_code == region_type.id, [&region_type] {
-						*g_pointers->m_gta.m_region_code = region_type.id;
-					});
+                    components::selectable(region_type.name, *g_pointers->m_gta.m_region_code == region_type.id, [&region_type] 
+                    {
+                        *g_pointers->m_gta.m_region_code = region_type.id;
+                        region_updated = true;
+                    });
```
Yimura: clang-format hasn't ran on this.

Yimura (changes requested):

src/views/network/view_network.cpp
rkwapisz: Ugh. I need to get better about this. Updated.

src/views/network/view_network.cpp
```diff
@@ -81,22 +81,42 @@ namespace big
 
 	void render_session_switcher()
 	{
+		if (g_pointers->m_gta.m_region_code == nullptr)
+			return;
+
+		static int selected_region_index     = -1;
+		static bool region_updated = false;
+
+        std::string region_str = (selected_region_index == -1) ? "SESSION_SELECT_COMBO"_T.data() : regions[*g_pointers->m_gta.m_region_code].name;
```
Yimura: Same here, sorry for not catching this earlier however using clang-format would've fixed this.

You can get it by installing the LLVM package on Windows (you need to add it to your path for it to start working).
clang-format-llvm

Yimura (changes requested):

src/views/network/view_network.cpp
rkwapisz: I missed this one as well, which actually brings me to a general question about code styling moving forward.

Would you prefer to see clang formatting applied against entire source files for consistency, or only to new code? If I were to execute clang formatting against the entire source file, we'd see a LOT more changes which would be functionally unnecessary, but would make the old code conform to the style.

Yimura: @rkwapisz, I have clang-format tied to my document saving so it will affect the whole file whether I like it or not.

In general it's better as old code will also conform to the styling guidelines, however gir489 might see some of these as unnecessary.

Yimura (approved):

src/views/network/view_network.cpp
maybegreat48: The region pointer is broken, that's why you're crashing here

PR #3103, opened 2024-05-12
Hide vehicle godmode

### LINKED ISSUE #2641: [Request]: 'Hide God Mode' also hides vehicle godmode
### Problem

Modders can see your vehicle being in godmode

### Solution

Hide the godmode status of your vehicle like that of your character

### Reason

abcdefg

### Additional context

_No response_

maybegreat48:
* Adds an option to hide vehicle god mode

Closes #2641

### DIFF
--- a/src/hooking/hooking.hpp
+++ b/src/hooking/hooking.hpp
@@ -197,7 +197,7 @@ namespace big
 
 		static void format_int(int64_t integer_to_format, char* format_string, size_t size_always_64, bool use_commas);
 
-		static void write_physical_script_game_state_data_node(__int64 iface, CPhysicalScriptGameStateDataNode* node);
+		static void write_physical_script_game_state_data_node(rage::CPhysical* this_ptr, CPhysicalScriptGameStateDataNode* node);
 	};
 
 	class minhook_keepalive
--- a/src/hooking/hooking.hpp
+++ b/src/hooking/hooking.hpp
@@ -38,6 +38,7 @@ class Network;
 class GtaThread;
 class CNetworkPlayerMgr;
 class CNetworkObjectMgr;
+class CPhysicalScriptGameStateDataNode;
 
 enum class eAckCode : uint32_t;
 
@@ -195,6 +196,8 @@ namespace big
 		static bool can_create_vehicle();
 
 		static void format_int(int64_t integer_to_format, char* format_string, size_t size_always_64, bool use_commas);
+
+		static void write_physical_script_game_state_data_node(__int64 iface, CPhysicalScriptGameStateDataNode* node);
 	};
 
 	class minhook_keepalive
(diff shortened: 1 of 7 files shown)


lonelybud:
closes #2641

src/hooking/hooking.hpp
```diff
@@ -195,6 +196,8 @@ namespace big
 		static bool can_create_vehicle();
 
 		static void format_int(int64_t integer_to_format, char* format_string, size_t size_always_64, bool use_commas);
+
+		static void write_physical_script_game_state_data_node(__int64 iface, CPhysicalScriptGameStateDataNode* node);
```
Rxann: could iface be a void or void*?

maybegreat48: It can, but it's eventually casted as an `__int64` anyway

Yimura: > could iface be a void or void*?

Normal void is not possible (I'm not sure the compiler even allows it), it needs to be a value that holds 64 bits to take the whole value and to pass it completely unchanged to the original function value. `iface` is probably a ptr to a class object however we probably haven't mapped it out yet and don't really need to as we're not modifying it.

gir489returns: iface is actually the [this_ptr](https://learn.microsoft.com/en-us/cpp/cpp/this-pointer?view=msvc-170) of the class that this function belongs to, probably CPhysicalObject or something like that. However, since he doesn't need it/care, and is only telling the compiler that the first register used in the __fastcall (all 64-bit functors are __fastcalls) is just some value, and what we really care about is the 2nd register, I feel this is fine. Void is actually not a real type, and void* just tells the compiler the same exact thing, Register 1 is just 64-bits wide.

src/hooks/spoofing/write_physical_script_game_state_data_node.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include <netsync/nodes/physical/CPhysicalScriptGameStateDataNode.hpp>
+
+namespace big
+{
+	void hooks::write_physical_script_game_state_data_node(__int64 iface, CPhysicalScriptGameStateDataNode* node)
+	{
+		g_hooking->get_original<hooks::write_physical_script_game_state_data_node>()(iface, node);
+
+		auto entity = reinterpret_cast<CVehicle*>((((rage::netObject*)(iface - 0x130)))->GetGameObject());
```
gir489returns, line 12: Actually, if you change iface to CVehicle, you should be able to just use that instead of doing this hacky kinda weird way of grabbing the root object from the 2nd parameter.

gir489returns (changes requested):

gir489returns:
I made some changes to make the parameter work, but once I basically copied what the function was doing from HexRays into Yim, the pointers matched and I no longer got exceptions coming from the function, and it matched to the vehicle I was currently in. However, I didn't test if someone else with a cheat could see if my vehicle was still in godmode or not, I'll leave that up to someone else.

Yimura (approved):

Issue #3108, opened 2024-05-13
[Bug]: Train commands only partially functional


KernelFrog:
### Describe the bug

YimMenu has Train Speed and Drive Train options (under World ->  Train).

Neither of these appear to have any affect on the train,

We should either fix these or remove them.

### Steps To Reproduce

1.  Inject YimMenu
2.  Climb on train
3.   Select Hijack Train, followed by Drive Train and/or Set train speed
4.  Note that train can't be driven/ speed doesn't change.

No messsages/errors in log.

### Expected Behavior

Should be possible to drive the train and/or change its speed.

### Logs

[06:29:50.6264099][INFO/main.cpp:156] Settings Loaded and logger initialized.
[06:29:50.6264320][INFO/main.cpp:158] Yim's Menu Initializing
[06:29:50.6264339][INFO/main.cpp:159] Git Info
        Branch: master
        Hash:   ef70781926c25066ac0455883fc5d58ce2029df1
        Date:   Mon May 13 08:45:57 2024
[06:29:50.6265325][INFO/main.cpp:171] Windows 10 Pro Version 22H2 (OS Build 19045.4291)
[06:29:50.6292861][DEBUG/thread_pool.cpp:22] Resizing thread pool from 0 to 4
[06:29:50.6294776][INFO/main.cpp:179] Thread pool initialized.
[06:29:50.6298141][INFO/pointers.hpp:94] GTA5 pointers cache is up to date, using it.
[06:29:50.6298187][INFO/pointers.cpp:1839] Loading pointers instance from cache
[06:29:50.6298196][INFO/pointers.cpp:1848] Pointers cache: Loading 218 fields from the cache
[06:29:50.6298277][DEBUG/module.cpp:55] Waiting for socialclub.dll...
[06:29:50.6299426][INFO/pointers.hpp:94] Social Club pointers cache is up to date, using it.
[06:29:50.6299449][INFO/pointers.cpp:1839] Loading pointers instance from cache
[06:29:50.6299460][INFO/pointers.cpp:1848] Pointers cache: Loading 5 fields from the cache
[06:29:50.6299781][INFO/main.cpp:182] Pointers initialized.
[06:29:50.6299936][INFO/main.cpp:189] Disabled anticheat gameskeleton.
[06:29:50.6301401][INFO/main.cpp:192] Byte Patch Manager initialized.
[06:29:50.6306147][INFO/main.cpp:195] Renderer initialized.
[...]

PR #3115, opened 2024-05-13
fix delete train

GwyrddGlas:
The delete train feature was non function because of an incorrect condition (!self::veh) in the delete_train function. 

This PR fixes the issue by removing the problematic condition and simplifying the delete_train function. Now, it only checks if get_closest_train() returns a non-zero value to determine if a train is found.

In reference to #3108

### DIFF
--- a/src/util/train.hpp
+++ b/src/util/train.hpp
@@ -38,7 +38,7 @@ namespace big::train
 
 	inline void delete_train()
 	{
-		if (!self::veh && get_closest_train() != 0)
+		if (get_closest_train() != 0)
 		{
 			VEHICLE::DELETE_ALL_TRAINS();
 		}


gir489returns:
This seems to counteract the idea of what the cheat tells you, which is you can't delete the train while inside of it. self::veh should be populated while inside the train. What is the purpose of this change?

rkwapisz:
There are other issues with this feature. Feel free to take a stab at them or just ignore them entirely. I don't think many people use this feature.

1) get_closest_train() does not do what it implies. There is no distance check. It's simply returning the first VEHICLE_FREIGHT it finds in the pool which may not be the nearest one to the player.

2) Delete Train invokes the DELETE_ALL_TRAINS natives which does what it suggests. If we want to delete only the train the player is in, then we should be getting the vehicle handle and deleting it directly.

3) My expectation would be that when you "hijack" a train, you're able to immediately control it. The "drive train" feature is confusing and should simply be default behavior.

4) Unsure of what value Exit Train is supposed to add. The player character is able to leave the train like any other vehicle (though the animation is broken).

gir489returns:
Yeah that's what I thought.

PR #3132, opened 2024-05-15
[Enhancement]: Stability Improvements to Black Hole Feature

rkwapisz:
The Black Hole feature was reworked since it was incredibly spammy and would cause a crash nearly 100% of the time within a few seconds of activating due to the sheer density of shit we were yeeting around. I've clamped the Black Hole feature to trigger only once every second, and only to the first 30 elements of the entity vector. The result is a Black Hole feature that is still really fun to use, but won't severely chug performance and possibly cause a crash unless your machine is a potato.

EDIT: Additional discussion regarding ped/veh features that call get_entities() is now out-dated.

> I've occasionally observed undefined behavior leading to crashes when using features that call entity::get_entities every tick. Since I haven't been able to find a consistent pattern to this behavior, after some discussion I believe that we're most likely simply creating unsafe race conditions. Greatest impacts were from the ped features in the World tab and Black Hole.
> 
> This PR attempts to mitigate this issue by using chrono to enforce a maximum frequency of calls to entity::get_entities in these features. We're now reading the results of entity::get_entities into a separate local vector at regular intervals and using that instead for our various native calls. While this likely doesn't eliminate a potential race condition entirely, it significantly reduces the chances of it happening. I chose chrono instead of just counting game ticks for better consistency and more uniform behavior across different machines.
> 
> In ped-related functions, I've clamped calls to every 100ms with no noticeable impacts to ped feature behavior.

### DIFF
--- a/src/backend/looped/weapons/flying_axe.cpp
+++ b/src/backend/looped/weapons/flying_axe.cpp
@@ -20,9 +20,6 @@ namespace big
 
 	class flying_axe : looped_command
 	{
-		std::vector<Entity> entity_list;
-		std::chrono::steady_clock::time_point last_call_time;
-
 		int get_around_position_iterator = 0;
 		Vector3 get_around_position(int ent, float far_distance, float up_down_distance)
 		{
@@ -153,19 +150,7 @@ namespace big
 					play_bone_looped_ptfx(entity_axe, 0, "scr_powerplay", "scr_powerplay", "sp_powerplay_beast_appear_trails", 0.5f, false, 0.f, 0.f, 0.f);
 					play_non_loop_ptfx(entity_axe, "scr_paletoscore", "scr_paletoscore", "scr_paleto_box_sparks", 0.1f, false, 0.f, 0.f, 0.f);
 				}
-
-				auto current_time = std::chrono::steady_clock::now();
-				auto elapsed_time = std::chrono::duration_cast<std::chrono::milliseconds>(current_time - last_call_time).count();
-
-				if (elapsed_time >= 100)
-				{
-					// Mitigate thread-unsafe behavior of get_entities by not calling it every tick (should be minimal impact to in-game ped behavior)
-					entity_list    = entity::get_entities(true, true);
-					last_call_time = current_time;
-				}
-
-				for (auto entity : entity_list)
-				{
+				for (auto entity : entity::get_entities(true, true)) {
 					function_axe_reaction(entity);
 				}
 
--- a/src/backend/looped/weapons/flying_axe.cpp
+++ b/src/backend/looped/weapons/flying_axe.cpp
@@ -20,6 +20,9 @@ namespace big
 
 	class flying_axe : looped_command
 	{
+		std::vector<Entity> entity_list;
+		std::chrono::steady_clock::time_point last_call_time;
+
 		int get_around_position_iterator = 0;
 		Vector3 get_around_position(int ent, float far_distance, float up_down_distance)
 		{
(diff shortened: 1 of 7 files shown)


xiaoxiao921:
Hum, does this really happen when it's called from the script thread? The discussion we had yesterday iirc was related to you using it outside of that thread, no?

Yimura:
All of the features in question that have been touched are running from within a script thread, we're really gonna need some actual stack traces that show these features crashing or a proof of concept that can consistently cause a crash.

Edit:
And adding a timeout will not eliminate the race condition, probably only make it _less_ likely.

rkwapisz:
> All of the features in question that have been touched are running from within a script thread, we're really gonna need some actual stack traces that show these features crashing or a proof of concept that can consistently cause a crash.

Sure, this makes sense. I was unaware that the script thread should be providing some safety.
I can backtrack some of those updates, but I'd recommend still keeping the black hole change since that there are other issues with it.

rkwapisz:
Thanks all for the clarifications and understanding. I've walked back changes to the ped/veh features but kept some in for Black Hole since, regardless of its access to get_entities(), it's extremely spammy and causes crashes.

xiaoxiao921:
From my understanding looking at the decompilation of the function you gave yesterday in the log it has to be a racing condition, the ptr get invalidated in the same loop iteration

rkwapisz:
> From my understanding looking at the decompilation of the function you gave yesterday in the log it has to be a racing condition, the ptr get invalidated in the same loop iteration

Yourself and Yim are pointing me in the right direction; I'm still learning some of YimMenu's architecture.

At risk of hijacking the conversation, I think what I've learned is that my issue the other day was calling get_entities() from inside esp::draw(), which itself it being called by the GUI renderer here inside gui.cpp:

`g_renderer.add_dx_callback(esp::draw, eRenderPriority::ESP); // TODO: move to ESP service`

So now I can see the obvious. The GUI renderer is asyncronous with GTA, so making calls to GTA functions from within the renderer is inherently risky. A safer approach for me is to let get_entities() write to a shared entity vector in YimMenu, then accessing that vector with the renderer instead of trying to call get_entities from the rendered directly. There may still be conflicts, but I'd imagine the worst thing that can happen is we draw an ESP somewhere that it shouldn't for a tick.

Whereas the script thread (which I'm just learning now is a thing) is syncronous with the game, so any calls we make are in series with other calls. Am I getting the right idea here, or am I still missing some details?

xiaoxiao921:
This is actually a quite common architecture for games to have.
```cpp
// Called by a thread on a fixed rate
game_script_manager::update();

// Rendering thread, called as much as we can
renderer::update();
```

Using a vector that you only access with a std::scoped_lock + mutex combo is what I do most of the time for cases like this

Rxann:
> This is actually a quite common architecture for games to have.
> 
> ```cpp
> 
> // Called by a thread on a fixed rate
> 
> game_script_manager::update();
> 
> 
> 
> // Rendering thread, called as much as we can
> 
> renderer::update();
> 
> ```
> 
> 
> 
> Using a vector that you only access with a std::scoped_lock + mutex combo is what I do most of the time for cases like this
> 
> 

I was going to comment this earlier but forgot, a mutex and a lock would fix the issue well most likely.

rkwapisz:
I think the PR is ready for a final review. The scope of work has changed but the update to Black Hole has eliminated the crashes I've experienced with it.

Special thanks to Xiao and Yim for the brief lesson in safe threading.

Yimura (approved):

gir489returns (approved):
Another banger.

PR #3133, opened 2024-05-16
Optimize OOM Kick

Rxann:
Uses 200 messages instead of 2000. Tested against detected modders and works the same.


rkwapisz:
Is there a measured performance improvement for YimMenu? If the functionality is *not* changing, what are we gaining?

Rxann:
> Is there a measured performance improvement for YimMenu? If the functionality is *not* changing, what are we gaining?

Well, for one it is about 10x less resource intensive and it is noticeably faster.

PR #2929, opened 2024-04-08
Give optimisation hints to compiler

Yimura:
I split up my work into multiple commits as I did NO TESTING whatsoever to see if they make a difference, they can be reverted at any point in time.

This should definitely not be forced on every branch that exists in YimMenu, hot code paths should be prioritized (game thread mainly and expensive operations).

PR #3136, opened 2024-05-16
move session lock logic to handle_join_request

### LINKED ISSUE #3135: [Improvement]: add better session lock
### Which feature

the current handling of session lock is in `assign_physical_index` 
https://github.com/search?q=repo%3AYimMenu%2FYimMenu%20lock_session%20&type=code

 which result in joining and kicking of player and give the bad idea to player that the host tried to kick them for no reason. Instead handle the session lock logic in `handle_join_request` so that the player is not allowed to even join - 

eg -

https://github.com/lonelybud/lonelyMenu/blob/master/src/hooks/protections/handle_join_request.cpp

### Reason

improve quality of life

### Additional context

_No response_

lonelybud:
closes #3135

### DIFF
--- a/src/hooks/player_management/assign_physical_index.cpp
+++ b/src/hooks/player_management/assign_physical_index.cpp
@@ -115,16 +115,7 @@ namespace big
 					}
 
 					if (plyr->block_join && *g_pointers->m_gta.m_is_session_started)
-					{
-						if (g_player_service->get_self()->is_host())
-						{
-							dynamic_cast<player_command*>(command::get("breakup"_J))->call(plyr, {});
-						}
-						else if (g.protections.desync_kick)
-						{
-							dynamic_cast<player_command*>(command::get("desync"_J))->call(plyr, {});
-						}
-					}
+						dynamic_cast<player_command*>(command::get("smartkick"_J))->call(plyr, {});
 
 					if (is_spoofed_host_token(plyr->get_net_data()->m_host_token))
 					{
--- a/src/hooks/player_management/assign_physical_index.cpp
+++ b/src/hooks/player_management/assign_physical_index.cpp
@@ -120,25 +120,9 @@ namespace big
 						{
 							dynamic_cast<player_command*>(command::get("breakup"_J))->call(plyr, {});
 						}
-						else
+						else if (g.protections.desync_kick)
 						{
 							dynamic_cast<player_command*>(command::get("desync"_J))->call(plyr, {});
-						}
-					}
-
-					if (g.session.lock_session && g_player_service->get_self()->is_host() && *g_pointers->m_gta.m_is_session_started)
-					{
-						if ((plyr->is_friend() && g.session.allow_friends_into_locked_session) || plyr->is_trusted)
-						{
-							g_notification_service.push_success("LOBBY_LOCK"_T.data(),
-							    std::vformat("LOBBY_LOCK_ALLOWED"_T.data(),
-							        std::make_format_args(plyr->get_net_data()->m_name)));
-						}
-						else
-						{
-							dynamic_cast<player_command*>(command::get("smartkick"_J))->call(plyr, {});
-							g_notification_service.push_warning("LOBBY_LOCK"_T.data(),
-							    std::vformat("LOBBY_LOCK_DENIED"_T.data(), std::make_format_args(plyr->get_net_data()->m_name)));
 						}
 					}
 
--- a/src/hooks/protections/handle_join_request.cpp
+++ b/src/hooks/protections/handle_join_request.cpp
@@ -1,5 +1,6 @@
 #include "hooking/hooking.hpp"
 #include "pointers.hpp"
+#include "services/friends/friends_service.cpp"
 #include "services/player_database/player_database_service.hpp"
 
 #include <network/CJoinRequestContext.hpp>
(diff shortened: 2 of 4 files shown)


src/hooks/player_management/assign_physical_index.cpp
```diff
@@ -120,28 +120,12 @@ namespace big
 						{
 							dynamic_cast<player_command*>(command::get("breakup"_J))->call(plyr, {});
 						}
-						else
+						else if (g.protections.desync_kick)
```
Rxann: smart kick should probably be subbed in here so we can call that once and it will handle the host logic.

src/services/friends/friends_service.hpp
```diff
@@ -13,6 +13,7 @@ namespace big
 		friends_service& operator=(const friends_service&)     = delete;
 		friends_service& operator=(friends_service&&) noexcept = delete;
 
+		[[nodiscard]] static bool is_friend(int64_t rockstar_id);
```
Rxann, line 16: We could probably keep the original and just have an overload for the rid

Rxann (commented):
I don't think the assign_player_index code entirely needs to be removed, if for example a user were able to somehow enter and then they would need to be kicked.

src/services/friends/friends_service.hpp
lonelybud: i can declare this one inline

lonelybud:
> I don't think the assign_player_index code entirely needs to be removed, if for example a user were able to somehow enter and then they would need to be kicked.

in the worst case scenario it can happen if you have disabled `idlekick` and the focus from game window is lost for a long time. There is still check present in `assign_physical_index` to kick block players if they happen to join when we are host.

Rxann:
> > I don't think the assign_player_index code entirely needs to be removed, if for example a user were able to somehow enter and then they would need to be kicked.
> 
> 
> 
> in the worst case scenario it can happen if you have disabled `idlekick` and the focus from game window is lost for a long time. There is still check present in `assign_physical_index` to kick block players if they happen to join when we are host.

That should not be relied apon.

src/hooks/player_management/assign_physical_index.cpp
```diff
@@ -115,16 +115,7 @@ namespace big
...
-						if (g_player_service->get_self()->is_host())
-						{
-							dynamic_cast<player_command*>(command::get("breakup"_J))->call(plyr, {});
-						}
-						else if (g.protections.desync_kick)
-						{
-							dynamic_cast<player_command*>(command::get("desync"_J))->call(plyr, {});
-						}
-					}
+						dynamic_cast<player_command*>(command::get("smartkick"_J))->call(plyr, {});
```
Rxann, line 118: Is this indented properly?

lonelybud: yes

Rxann: Oh yeah sorry, I was looking on github mobile. this can be resolved.

src/hooks/protections/handle_join_request.cpp
```diff
@@ -10,19 +11,41 @@ namespace big
...
 			g_notification_service.push("BLOCK_JOIN"_T.data(),
-			    std::vformat("BLOCK_JOIN_INFO"_T, std::make_format_args(player->name)));
+			    std::vformat("BLOCK_JOIN_INFO"_T, std::make_format_args(player_info->m_name)));
 			return false;
 		}
-		else [[likely]]
+		else if (g.session.lock_session) [[unlikely]]
 		{
-			return g_hooking->get_original<hooks::handle_join_request>()(network, session, player_info, ctx, is_transition_session);
+			if ((is_friend && g.session.allow_friends_into_locked_session) || is_trusted)
```
Rxann, line 36: Not required but [[unlikely]] and [[likely]] could probably be used. May want to wait for yimura or gir to approve that idea so it isn't a waste of time.

gir489returns: This references a configuration boolean, and as such we cannot make assumptions about what the user will do, thus we can't make assumptions for the compiler as which path will be mostly taken. We made a core set of rules about when to use and when not to use the branch prediction adjustor modifiers in the original ticket we submitted the change in. https://github.com/YimMenu/YimMenu/pull/2929#issuecomment-2045245855

lonelybud:
Edited
this hook as tested is only called when you are host. It wont be called when someone say your friend try to click join button from social club and you are not host of the session at that time

lonelybud:
> > > I don't think the assign_player_index code entirely needs to be removed, if for example a user were able to somehow enter and then they would need to be kicked.
> > 
> > 
> > in the worst case scenario it can happen if you have disabled `idlekick` and the focus from game window is lost for a long time. There is still check present in `assign_physical_index` to kick block players if they happen to join when we are host.
> 
> That should not be relied apon.

it will work fine as intended don't worry

Rxann:
@gir489returns what do you think about this

src/hooks/protections/handle_join_request.cpp
```diff
@@ -10,19 +11,41 @@ namespace big
...
+		if (block_join) [[unlikely]]
+		{
+			send_response(db_player->block_join_reason);
 			g_notification_service.push("BLOCK_JOIN"_T.data(),
-			    std::vformat("BLOCK_JOIN_INFO"_T, std::make_format_args(player->name)));
+			    std::vformat("BLOCK_JOIN_INFO"_T, std::make_format_args(player_info->m_name)));
 			return false;
 		}
-		else [[likely]]
+		else if (g.session.lock_session) [[unlikely]]
```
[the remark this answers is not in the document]
lonelybud: yes. i get your point. Edit: in my fork i am not using any of these for now bcz i dont completely understand it

Rxann:
@gir489returns should we keep the code in the assign_player_index as a fall through?

Yimura (dismissed):

gir489returns (approved):

Yimura (approved):

Yimura (approved):

PR #3137, opened 2024-05-16
improve smart kick

lonelybud:
improvements

- safely kick player with godmode infraction
- only use desync when protection is enabled for self

### DIFF
--- a/src/backend/commands/player/kick/smart_kick.cpp
+++ b/src/backend/commands/player/kick/smart_kick.cpp
@@ -1,8 +1,10 @@
 #include "backend/player_command.hpp"
+#include "core/data/infractions.hpp"
 #include "core/scr_globals.hpp"
 #include "natives.hpp"
 #include "pointers.hpp"
 #include "script.hpp"
+#include "services/player_database/player_database_service.hpp"
 
 namespace big
 {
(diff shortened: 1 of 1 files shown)


Rxann (commented):
Should something be done if they have desync protections aswell?

lonelybud:
> Should something be done if they have desync protections aswell?

need to figure out

Rxann:
> > Should something be done if they have desync protections aswell?
> 
> 
> 
> need to figure out

If they have attack w gmode or desync protections use the OOM

rkwapisz:
Why OOM for "attack with god mode"?

Every menu has a god mode option, but not all of them have desync protections.

xiaoxiao921:
What's even the point of using desync inside the smart kick if oom is stronger anyway?

lonelybud:
> What's even the point of using desync inside the smart kick if oom is stronger anyway?

oom is blocked my most menus and will also let the target know who attacked him.

lonelybud:
> Why OOM for "attack with god mode"?
> 
> Every menu has a god mode option, but not all of them have desync protections.

yes you are right. There is one exception where yim users dont spoof god mod and they will be marked with god mod infraction hmn then this pr is still ineffective bcz oom kick is blocked my the menus anyways

PR #3141, opened 2024-05-17
Fix fake ban message

ghost:


Rxann:
Nice

Rxann (commented):
Very subtle change. Looks good.

gir489returns (approved):

gir489returns:
```cpp
union Args {
	struct
	{
		int64_t event;
		int64_t id;
		int64_t playerId;
	};
	char message[15]; // if (HUD::GET_LENGTH_OF_LITERAL_STRING_IN_BYTES(sParam1) > 15) return;
};

const size_t arg_count = 9;
Args args = {(int64_t)eRemoteEvent::SendTextLabelSMS, self::id, 1 << player->id()};
std::string str        = "HUD_ROSBANPERM";
std::copy(str.begin(), str.end(), args.message);

g_pointers->m_gta.m_trigger_script_event(1, (int64_t*)&args, arg_count, 1 << player->id(), (int)eRemoteEvent::SendTextLabelSMS);
```

We can also do it this way.

gir489returns:
image

This feautre doesn't work the way it used to. I guess they patched it? Before, it used to open a dialog window saying you were banned. Now it just does this,... Not exactly gonna make them panic or anything. I think this feature should be removed.

rkwapisz:
> image
> 
> This feautre doesn't work the way it used to. I guess they patched it? Before, it used to open a dialog window saying you were banned. Now it just does this,... Not exactly gonna make them panic or anything. I think this feature should be removed.

Agreed. Get rid of it. I wouldn't even blink at an in-game text like that.

ghost:
We could create a checkbox and add a loop to it that would display a warning message `HUD::SET_WARNING_MESSAGE_WITH_HEADER_AND_SUBSTRING_FLAGS("WARN", "JL_INVITE_ND", 2, "", true, -1, -1, "You have been permanently banned from entering Grand Theft Auto Online Mode.", " Return to Grand Theft Auto V.", true, 0);` but I don't know how to do it, maybe someone else will create another PR.

maybegreat48:
> it used to open a dialog window saying you were banned

It never did that to my knowledge

gir489returns:
> > it used to open a dialog window saying you were banned
> 
> It never did that to my knowledge

The one I had did a while ago did. Must've gotten patched.

PR #3143, opened 2024-05-17
Automatic Vectorization

Rxann:
I became aware of this compiler flag that seemingly helps speed up YimMenu code via vectorization, so I added it.


gir489returns:
This flag is extremely dangerous, as we manually control the timing of our threads. Have you tested this at all???

Rxann:
> This flag is extremely dangerous, as we manually control the timing of our threads. Have you tested this at all???

Yes, I have tested it with no observable side effects. That could also mean that the flag is somehow not actually vectorizing anything.

gir489returns:
Rejected. Flag is too dangerous, and no way to prove any discernable gains without side effects.

PR #3156, opened 2024-05-19
Improving Release Post

ghost:
If there have been commits in the last 24 hours, the release is recreated; if not, everything remains as is. This will prevent people from seeing an old release date on the main GitHub page and thinking that the project has no longer been updated
dick_pick


xiaoxiao921:
Was this tested?

ghost:
> Was this tested?

Sure

xiaoxiao921:
Do you have an actual proof I can check so that I know it was tested?

ghost:
Here is how to create a release from scratch
first_creation
Further work in case there are no new commits:
no_commits_last_2mins
Work in case there were commits:
release_and_ddl_were_recreated_after_commit
[The repository where the tests were](https://github.com/Farfour-1-4-8-8/TestNightly/actions)

ghost:
I have a question. Why do we always build the project first and then check if there were new commits? Maybe it would be better to do it the other way around, and then we wouldn't need to start the build, or did I misunderstand something?

GreySideMM:
Sry for OT but WTF is this branch name? (And the name of your test account, too.)

ghost:
I switched the order of checking for the latest commits and the build process. Now, if there are no commits, the build will not start. I think this will be more logical behavior.The newly conducted tests are below.
1. If there have been new commits, then a release is created.
if_there_were_commits
2. If there have been no new commits, the build and release are skipped.
if_there_were_no_commits

.github/workflows/nightly.yml
```diff
@@ -50,28 +69,37 @@ jobs:
...
+          # Check if release exists and delete it
+          if [ "$RELEASE_ID" != "null" ]; then
+            curl -s -X DELETE -H "$AUTH_HEADER" "$GITHUB_API_URL/repos/$REPO/releases/$RELEASE_ID"
+            echo "Deleted release with ID $RELEASE_ID"
+          else
+            echo "No existing release to delete"
+          fi
+
+          # Delete the tag
+          curl -s -X DELETE -H "$AUTH_HEADER" "$GITHUB_API_URL/repos/$REPO/git/refs/tags/$TAG" || true
```
Yimura, line 101: Why use raw curl requests when this can be done using a package?
https://github.com/actions/github-script

Yimura (changes requested):

Yimura:
> I have a question. Why do we always build the project first and then check if there were new commits? Maybe it would be better to do it the other way around, and then we wouldn't need to start the build, or did I misunderstand something?

A long time ago when YimMenu didn't disable the anti-cheat we had an action step that would check against Rockstar's signature database, this is why we would require a rebuild of the binary to verify that the menu was still undetected, if a signature from Rockstar's database was detected in YimMenu we would just remove or rewrite to code to get rid of the detection. This ordering of action steps has remained ever since even though I removed the signature check when they retired their old signature scanning system.

In regards to you reporting the comment of GreySideMM for abuse I find this incredibly immature on your part as your username at best is edgy and at worst an insult. I can look past that however your branch name is INCREDIBLY immature and you should know better. Next time I see this I will revoke your access to the YimMenu organisation and report it to Github support.

GreySideMM:
> In regards to you reporting the comment of GreySideMM for abuse I find this incredibly immature on your part as your username at best is edgy and at worst an insult. I can look past that however your branch name is INCREDIBLY immature and you should know better. Next time I see this I will revoke your access to the YimMenu organisation and report it to Github support.

I'm sorry again for continuing OT but I should clarify my previous post: His testusers name (see link & screenshots) is "Farfour-1-4-8-8". 1488 is a neonazi/rasist code which stands for "[Fourteen Words Hail Hitler](https://en.wikipedia.org/wiki/Fourteen_Words)" and Farfour is part of an Hamas-near propaganda child series "[Tomorrow's Pioneers](https://en.wikipedia.org/wiki/Tomorrow's_Pioneers)". That was further part of why I posted.

PR #3159, opened 2024-05-19
Lessen Detections

Rxann:
Only activate desync protections(when not host) and only spoof host token when not host to stop other players from detecting us as modders for unnecessary reasons.


rkwapisz:
Host token needs to be set before joining a session.

Mr-X-GTA:
This will not change anything

PR #3161, opened 2024-05-20
Harass players and spam killfeed

maybegreat48:
* Harass Players: Makes other players think that you are following them around using position and vehicle spoofing. Note that this breaks many things
* Spam Killfeed: Spams the player's killfeed with fake death messages
* Refactored many node writers into a single function

Thanks to @Rxann and @Mr-X-GTA for extensive testing


gir489returns (approved):
I like this better than the way we were getting the netObject for the vehicle node writer before because of its multiinheritence.

Yimura (approved):

PR #3167, opened 2024-05-21
Session multiplexer and more

### LINKED ISSUE #3166: [Bug]: Various Crashes
### Describe the bug

Any protection against these crashes?  its always these 2/3 that crash me doesn't matter if i am host or not, either.
Screenshot_64

### Steps To Reproduce

Go into a session with a modder, you don't have to do anything since they are severely toxic most of the time.

### Expected Behavior

To block these crashes as it says its doing, but seems to do nothing.

### Logs

[18:46:41.4598744][WARN/notify.cpp:35] Blocked player ped removal crash from ChadddyBoyyyy (162939655)
[18:46:41.4599264][WARN/notify.cpp:35] Blocked invalid weapon type crash from ChadddyBoyyyy (162939655)
[18:46:41.4599452][WARN/reaction.cpp:59] Received Sound Spam from ChadddyBoyyyy (162939655)

### Store

Epic Games

### Language

English

### YimMenu Language

English (US)

### Additional context

_No response_

### Build

### Coload

maybegreat48:
* Session multiplexer: Makes your session much more visible to other players by submitting multiple sessions to the matchmaking server. High values generally fill your session up in seconds
* Increase session limit: Increases session limit from 30 to 32 players when you are the host
* Block CEO creation: Prevents players from forming orgs through the interaction menu. Other methods work
* Fixed several bugs with session attribute spoofing

Fixes #3166, but for how long?


src/hooks/matchmaking/process_matchmaking_find_response.cpp
```diff
@@ -62,6 +62,7 @@ namespace big
 			int i = 0;
 			for (auto result = node->get_child_node("Results")->m_child; result; result = result->m_sibling)
 			{
+				LOG(INFO) << result->get_child_node("Attributes")->m_value;
```
BugisoftRSG: // TODO: Remove me

cmake/gtav-classes.cmake
```diff
@@ -2,8 +2,8 @@ include(FetchContent)
 
 FetchContent_Declare(
     gtav_classes
-    GIT_REPOSITORY https://github.com/Yimura/GTAV-Classes.git
-    GIT_TAG        46c1c63396cf40b206b9f5dd612f293809eefcd1
+    GIT_REPOSITORY https://github.com/maybegreat48/GTAV-Classes.git
+    GIT_TAG        7620f104072a8c13a54587c0bbadedc22d9c870b
```
gir489returns, line 6: Revert this, and fastforward the tag hash.

src/backend/looped/session/block_ceo_creation.cpp
```diff
@@ -0,0 +1,37 @@
...
+		using looped_command::looped_command;
+
+		virtual void on_enable() override
+		{
+			g_script_patcher_service->update();
+		}
+
+		virtual void on_tick() override
+		{
+			if (*g_pointers->m_gta.m_is_session_started && !STREAMING::IS_PLAYER_SWITCH_IN_PROGRESS() && SCRIPT::GET_NUMBER_OF_THREADS_RUNNING_THE_SCRIPT_WITH_THIS_HASH("maintransition"_J) == 0)
```
gir489returns, line 21: Player Switch In Progress is a candidate for the unlikely suggestion to the compiler. Break these out if you agree.

src/services/matchmaking/matchmaking_service.cpp
```diff
@@ -113,4 +138,162 @@ namespace big
...
+			m_multiplexed_sessions.erase(it);
+			return false;
+		}
+		else
+		{
+			if (g.spoofing.multiplex_session)
+				for (auto& [_, children] : m_multiplexed_sessions)
+					for (auto& session : children)
+						if (session.m_data2 == id->m_data2 && session.m_data3 == id->m_data3)
+							return true; // prevent auto cleanup
```
gir489returns, line 274: I don't like the way this looks, I think you should add curley braces. Anything more than 3 nested statements becomes too much.

src/services/matchmaking/matchmaking_service.cpp
```diff
@@ -113,4 +138,162 @@ namespace big
...
+		// this can be fire and forget, but it's probably a good idea to be notified if something goes wrong
+		if (auto it = m_multiplexed_sessions.find(rage::joaat(id->m_data1)); it != m_multiplexed_sessions.end())
+		{
+			if (!g.spoofing.multiplex_session)
+			{
+				// option disabled mid-session
+				return;
+			}
+
+			int i = 0;
```
gir489returns, line 230: Static? Otherwise the compiler will discard this after it falls out of scope.

gir489returns (changes requested):

src/services/matchmaking/matchmaking_service.cpp
maybegreat48: It gets captured by the lambda so it wouldn't fall out of scope

gir489returns: But the lambda is executed asynchronously. You may want to consider using atomic_bool or something, otherwise you may have something like 

```
update_session_advertisement returned false for multiplex task 3
update_session_advertisement returned false for multiplex task 4
update_session_advertisement returned false for multiplex task 5
update_session_advertisement returned false for multiplex task 4
update_session_advertisement returned false for multiplex task 5
```

maybegreat48: > But the lambda is executed asynchronously

The fiber pool isn't asynchronous, and it is guaranteed that the fibers won't run until the matchmaking hook returns. The only annoying thing is the order of the prints, but that's essentially it

gir489returns (approved):

Rxann:
I can testify for the crazy effectiveness of the multiplexer. I also noticed that the multiplexer works more and more once the number of players in the session increase.

maybegreat48:
> I also noticed that the multiplexer works more and more once the number of players in the session increase.

I recommend spoofing the session player count to 25 to increase its effectiveness. I've looked at the sorting code and there is definitely room for improvement

Rxann:
> > I also noticed that the multiplexer works more and more once the number of players in the session increase.
> 
> I recommend spoofing the session player count to 25 to increase its effectiveness. I've looked at the sorting code and there is definitely room for improvement

This is funny, but I was actually about to edit my comment and say to spoof it to 25 because I found it to work.

lonelybud:
> Block CEO creation: Prevents players from forming orgs through the interaction menu. Other methods work

can we have protection against it i mean it is like disturbing the session. Does it mean it stops players from becoming ceo or MC club ceo?

lonelybud:
> Makes your session much more visible to other players by submitting multiple sessions to the matchmaking server. High values generally fill your session up in seconds

what about poor hosts who enable this. Also what could be the impact when most of the players enables it. Also what about the sessions who are getting players without using this how they will be affected. I mean sometimes it feels like tinkering with already existing system can create problems. Just an opinion.

Rxann:
> Has anyone encountered the issue where, during the testing of this PR, attempting to teleport everyone to the center of the map causes players to start leaving the session one by one?

I don't use this feature, but I will test it.

maybegreat48:
> Has anyone encountered the issue where, during the testing of this PR, attempting to teleport everyone to the center of the map causes players to start leaving the session one by one?

I'm pretty sure they left on their own

lonelybud:
> Has anyone encountered the issue where, during the testing of this PR, attempting to teleport everyone to the center of the map causes players to start leaving the session one by one?

when you annoy them they know what to do

ghost:
> > Has anyone encountered the issue where, during the testing of this PR, attempting to teleport everyone to the center of the map causes players to start leaving the session one by one?
> 
> I'm pretty sure they left on their own

So, in general, this function only works if the force script host function is enabled. If it is not active, players are kicked out of the session without being teleported.

Yimura (approved):

PR #3190, opened 2024-05-26
Block join reason to reaction

### LINKED ISSUE #1991: [Low Priority]: Add block join alert and chat command permission to reaction
### Problem

Maybe someone will be bored and want to add it. I don't think it's very hard to implement.

### Solution

The block join alert combobox should be displayed below if the block join checkbox is checked in the reaction section. And, if possible, add a combobox of chat permission commands in the reaction so that it would be possible to prevent modders from using my commands in the chat, or at least assign access.

### Reason

Will prevent modders from using my chat commands and will give some access. Allows you to change the join blocking alert automatically instead of doing it manually in the player database. Perhaps someone wants to make their first PR and is looking for an easy task.

### Additional context

_No response_

### LINKED ISSUE #3131: [Request]: Add block join reason in the reactions
### Problem

There is no way to automatically specify the reason for the block, such as Cheater, Session Full, etc

### Solution

Just add a dropdown list with block join reasons under the 'block join' checkbox.

### Reason

It would be quite logical compared to setting it manually for each player, and I don't think it's very difficult to implement.

### Additional context

_No response_

ghost:
Closes #3131
Closes #1991
pic

### DIFF
--- a/cmake/async-logger.cmake
+++ b/cmake/async-logger.cmake
@@ -1,10 +1,12 @@
 include(FetchContent)
 
+add_compile_definitions(CXX_FORMAT_SUPPORT)
+
 message("AsyncLogger")
 FetchContent_Declare(
     AsyncLogger
     GIT_REPOSITORY https://github.com/Yimura/AsyncLogger.git
-    GIT_TAG v0.0.6
+    GIT_TAG 80ce938277acd44767f858099920ae20f1df42ca
     GIT_PROGRESS TRUE
 )
 FetchContent_MakeAvailable(AsyncLogger)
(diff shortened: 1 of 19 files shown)


maybegreat48 (dismissed):

binded-zz:
this looks totally awsome 
small request to add too this if at all possible 
any way to add grouping  in player database  like show by group 
so you put them in cheater block show only cheaters 


i tired to do something about a year ago but this wasnt even a option then it was like blocked friends and modders  but i couldnt make it work out i had colors and other stuff circles in circles and it was a no go 

but maybe just a drop list under search but not only block reasons maybe add friends modders trusted and command use as well then just show online offline invite only for that group

Yimura:
@binded-zz, don't mistake this PR is only for this feature being added. Not a place to request other features on top of what has already been made.

binded-zz:
sorry about the feature request on top of the PR

this looks so awesome

Yimura (approved):

PR #3192, opened 2024-05-27
Chat spam to reaction

### LINKED ISSUE #3026: [Improvement]: Add more options to react to spammers

ghost:
Closes #3026
Removed the spammers autokick checkbox and moved it to reactions

pic1


src/views/settings/view_reaction_settings.cpp
rkwapisz, line 1: If we're updating these reactions, might as well organize them alphabetically so it's easier to find a particular setting. What's the point of that separator anyway?

rkwapisz (commented):
Looks good. This will be useful. Huge influx of spammers recently.

rkwapisz:
One more thing I'd recommend doing is breaking out Block Joins from the Add Player to DB so players can more easily find the option to block spammers. Right now Block Joins is hidden from view unless Add Player to DB is enabled. Functionally it makes sense (player needs to be in the DB for them to be blocked from joining) but it's not intuitive.

Get rid of the conditional for the menu item in view_reaction_settings.cpp so the reaction is always available for toggling:

```
			if (reaction.add_to_player_db)
				ImGui::Checkbox("REACTION_BLOCK_JOINS"_T.data(), &reaction.block_joins);
```

And then in reaction.cpp, change if (add_to_player_db) to if (add_to_player_db || block_joins) so block_joins will automatically add the player to DB and then block them as well in the future. This removes the need for Add Player to DB to be enabled in order for block joins to work. Fewer clicks.

```
		if (add_to_player_db || block_joins)
		{
			auto entry = g_player_database_service->get_or_create_player(player);

			if (block_joins)
			{
				entry->block_join = true;
				g_player_database_service->save();
			}
		}
```

lonelybud:
> Looks good. This will be useful. Huge influx of spammers recently.

in china chat spammers hosts the lobby for more than hours sometimes.

ghost:
> One more thing I'd recommend doing is breaking out Block Joins from the Add Player to DB so players can more easily find the option to block spammers. Right now Block Joins is hidden from view unless Add Player to DB is enabled. Functionally it makes sense (player needs to be in the DB for them to be blocked from joining) but it's not intuitive.
> 
> Get rid of the conditional for the menu item in view_reaction_settings.cpp so the reaction is always available for toggling:
> 
> ```
> 			if (reaction.add_to_player_db)
> 				ImGui::Checkbox("REACTION_BLOCK_JOINS"_T.data(), &reaction.block_joins);
> ```
> 
> And then in reaction.cpp, change if (add_to_player_db) to if (add_to_player_db || block_joins) so block_joins will automatically add the player to DB and then block them as well in the future. This removes the need for Add Player to DB to be enabled in order for block joins to work. Fewer clicks.
> 
> ```
> 		if (add_to_player_db || block_joins)
> 		{
> 			auto entry = g_player_database_service->get_or_create_player(player);
> 
> 			if (block_joins)
> 			{
> 				entry->block_join = true;
> 				g_player_database_service->save();
> 			}
> 		}
> ```

I added this in [another PR](https://github.com/YimMenu/YimMenu/pull/3190) to avoid problems with merging both

src/services/player_database/persistent_player.cpp
```diff
@@ -22,6 +22,7 @@ namespace big
 	    {Infraction::INVALID_PLAYER_MODEL, "Had used an invalid player model"},
 	    {Infraction::SUPER_JUMP, "Had used super jump"},
 	    {Infraction::UNDEAD_OTR, "Had used undead OTR"},
```
maybegreat48, line 24: You need a null desc between UNDEAD_OTR and CHAT_SPAM or the game will crash

maybegreat48: `{Infraction::CUSTOM_REASON, ""},`

Rxann: Why does it crash in the first place? I feel like that should be addressed and I can work on it and make a separate pr if needed

Yimura (approved):

maybegreat48 (approved):

PR #3194, opened 2024-05-28
P2905R2 Fix - Refactor rvalue arguments into lvalues for std::make_format_args

rkwapisz:
`Rejecting temporaries in make_format_args is an (intentionally) breaking change.`

*thumbs up*

Refactored all calls to std::make_format_args to use lvalue args so that YimMenu can compile on VS 17.10.0

### DIFF
--- a/cmake/async-logger.cmake
+++ b/cmake/async-logger.cmake
@@ -1,10 +1,12 @@
 include(FetchContent)
 
+add_compile_definitions(CXX_FORMAT_SUPPORT)
+
 message("AsyncLogger")
 FetchContent_Declare(
     AsyncLogger
     GIT_REPOSITORY https://github.com/Yimura/AsyncLogger.git
-    GIT_TAG v0.0.6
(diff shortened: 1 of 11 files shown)


gir489returns:
This change effectively makes no sense. I don't understand why we need to break out the variables,

maybegreat48:
> This change effectively makes no sense. I don't understand why we need to break out the variables,

I'm assuming that recent versions of MSVC don't allow rvalue references (such as passing stuff directly into the function) into std::vformat because the C++ committee said so

[This](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2905r2.html#problems) appears to be the paper associated with this change

xiaoxiao921:
Just bump the git hash of the asybc logger dependency.

xiaoxiao921:
Somehow missed the other errors, my bad.

gir489returns (approved):

PR #3199, opened 2024-05-29
Make Player Stats Retrievable Within Lua

USBMenus:
lets the user retrieve other players rank and rp within lua
adds 2 network commands
get_player_rank()
get_player_rp()

### DIFF
--- a/src/lua/bindings/network.cpp
+++ b/src/lua/bindings/network.cpp
@@ -1,21 +1,53 @@
 #include "network.hpp"
 
 #include "../../script.hpp"
+#include "core/scr_globals.hpp"
 #include "hooking/hooking.hpp"
 #include "pointers.hpp"
 #include "services/player_database/player_database_service.hpp"
-#include "util/notify.hpp"
 #include "util/chat.hpp"
+#include "util/notify.hpp"
 #include "util/scripts.hpp"
 #include "util/session.hpp"
 #include "util/system.hpp"
 #include "util/teleport.hpp"
+#include <script/globals/GPBD_FM.hpp>
 
 namespace lua::network
 {
 	// Lua API: Table
 	// Name: network
 	// Table containing helper functions for network related features.
+
+	// Lua API: Function
+	// Table: network
+	// Name: get_player_rank
+	// Param: pid: int
+	// Call get_player_rank(playerID)
+	static int get_player_rank(int pid)
+	{
+		if (big::g_player_service->get_by_id(pid))
+		{
+			auto& stats = big::scr_globals::gpbd_fm_1.as<GPBD_FM*>()->Entries[pid].PlayerStats;
+			return stats.Rank;
+		}
+		return -1;
+	}
+
+	// Lua API: Function
+	// Table: network
+	// Name: get_player_rp
+	// Param: pid: int
+	// Call get_player_rp(playerID)
+	static int get_player_rp(int pid)
+	{
+		if (big::g_player_service->get_by_id(pid))
+		{
+			auto& stats = big::scr_globals::gpbd_fm_1.as<GPBD_FM*>()->Entries[pid].PlayerStats;
+			return stats.RP;
+		}
+		return -1;
+	}
 
 	// Lua API: Function
 	// Table: network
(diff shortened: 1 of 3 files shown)


gir489returns:
Is this really necessary when you can just read the globals?

maybegreat48:
You'll have to update the globals if you use them directly. A menu update is all that's required with this one (feel free to close again if you disagree)

xiaoxiao921:
That global is thoroughly used in the c++ menu code already so might aswell expose it

xiaoxiao921 (changes requested):
```cpp
// Lua API: Function
// Table: network
// Name: get_player_rank
// Param: player_id: integer: id of the player
// Returns: Rank of the player if the player id is valid, nil otherwise.
static sol::object get_player_rank(int player_id, sol::this_state state_)
{
	if (big::g_player_service->get_by_id(player_id))
	{
		auto& stats = big::scr_globals::gpbd_fm_1.as<GPBD_FM*>()->Entries[player_id].PlayerStats;
		return sol::make_object(state_.L, stats.Rank);
	}

	return sol::nil;
}
```

Let's make it a bit cleaner. Do the same for the other function please.

gir489returns:
> ```c++
> // Lua API: Function
> // Table: network
> // Name: get_player_rank
> // Param: player_id: integer: id of the player
> // Returns: Rank of the player if the player id is valid, nil otherwise.
> static sol::object get_player_rank(int player_id, sol::this_state state_)
> {
> 	if (big::g_player_service->get_by_id(player_id))
> 	{
> 		auto& stats = big::scr_globals::gpbd_fm_1.as<GPBD_FM*>()->Entries[player_id].PlayerStats;
> 		return sol::make_object(state_.L, stats.Rank);
> 	}
> 
> 	return sol::nil;
> }
> ```
> 
> Let's make it a bit cleaner. Do the same for the other function please.

This is inconconsistent with the other functions in the class.

get_flagged_modder_reason returns a blank string, and get_selected_database_player_rockstar_id returns -1. If we're going to change the incoming, we should change them for all, as all the functions should behave the same. Not sure why get_player_rank needs to return an object? What other properties besides the rank value as an integer does the callee need?

src/lua/bindings/network.cpp
```diff
@@ -227,6 +259,8 @@ namespace lua::network
 
 		auto ns = state["network"].get_or_create<sol::table>();
 
+		ns["get_player_rank"]                          = get_player_rank;
+		ns["get_player_rp"]                            = get_player_rp;
```
gir489returns, line 263: Can you put this in chronological order that they appear in the class?

gir489returns (changes requested):

xiaoxiao921:
> > ```c++
> > // Lua API: Function
> > // Table: network
> > // Name: get_player_rank
> > // Param: player_id: integer: id of the player
> > // Returns: Rank of the player if the player id is valid, nil otherwise.
> > static sol::object get_player_rank(int player_id, sol::this_state state_)
> > {
> > 	if (big::g_player_service->get_by_id(player_id))
> > 	{
> > 		auto& stats = big::scr_globals::gpbd_fm_1.as<GPBD_FM*>()->Entries[player_id].PlayerStats;
> > 		return sol::make_object(state_.L, stats.Rank);
> > 	}
> > 
> > 	return sol::nil;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Let's make it a bit cleaner. Do the same for the other function please.
> 
> This is inconconsistent with the other classes in the Lua exporter. I think we should have an all or nothing paradigm. Either all of them use Sol objects, or none of them do.

Hmm not wrong, never realized that, well I guess that's whatever then

gir489returns:
> Hmm not wrong, never realized that, well I guess that's whatever then

We should move that to an epic(Yim calls them Chores?), but for this PR, I think what he has is fine, bar a few nitpicks. I really hate that god damn clang formatter, but I'm screaming into the void at this point with it for something that's superfical.

Yimura:
So is this fine?

xiaoxiao921:
The doc left to be desired and it's not generated. (Python script not ran). Other than that yeah it's ok

src/lua/bindings/network.cpp
USBMenus: aren't they already in chronological order?

USBMenus: done

gir489returns (approved):

xiaoxiao921 (approved):

PR #3204, opened 2024-05-30
Add new spam words

### LINKED ISSUE #3202: [Improvement]: Please improve chat spam protection
### Which feature

Is it possible to block links to discord?

### Reason

Many people simply advertise discord with mods/cheats
screen

### Additional context

_No response_

Rxann:
closes #3202

### DIFF
--- a/src/util/chat.hpp
+++ b/src/util/chat.hpp
@@ -112,7 +112,7 @@ namespace
 	    "\xE5\x81\xB7\xE6\x8B\x8D",             // AV
 	    "\xE4\xBC\xA0\xE7\x85\xA4",             // AV
 	    "\xE4\xB9\xB1\xE8\xAE\xBA",             // AV
-	    "情色影院：сб54.сo | >",                // 6 is cyrillic letter for B
+	    "\xE6\x83\x85\xE8\x89\xB2",             // Erotic
 	};
 }
 
--- a/src/util/chat.hpp
+++ b/src/util/chat.hpp
@@ -44,6 +44,7 @@ namespace
 	    "Money\\\\",
 	    "Money\\",
 	    ".gg",
+	    ".GG",
 	    "--->",
 	    "shopgta5",
 	    "doit#",
--- a/src/util/chat.hpp
+++ b/src/util/chat.hpp
@@ -28,23 +28,17 @@ namespace
 
 	static const char* spam_texts[] = {
 	    "qq", //a chinese chat app
-	    "QQ",
-	    "WWW.",
 	    "www.",
 	    ".cn",
-	    ".CN",
 	    ".cc",
-	    ".CC",
-	    ".TOP",
-	    ".COM",
+	    ".com",
 	    ".top",
 	    "\xE3\x80\x90", //left bracket in Chinese input method
-	    "/Menu",
-	    "Money/",
-	    "Money\\\\",
-	    "Money\\",
+	    "/menu",
+	    "money/",
+	    "money\\\\",
+	    "money\\",
 	    ".gg",
-	    ".GG",
 	    "--->",
 	    "shopgta5",
 	    "doit#",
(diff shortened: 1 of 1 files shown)


gir489returns:
Instead of this manual nonsense we keep doing, why don't we just reimplement the rockstar original callback for incoming chat messages? Also, yet again with the nonsense formatting.

Rxann:
> Instead of this manual nonsense we keep doing, why don't we just reimplement the rockstar original callback for incoming chat messages? Also, yet again with the nonsense formatting.

This sounds like a good idea, but for now I think the new words should just be merged until that can be done. Per usual, I think you already will guess what I am going to say. I run the formatter.

Edit:
Won't we still need to have a manual list though? It's not like the spam detects itself afaik.

gir489returns:
@maybegreat48 Would the return value of receive_net_message indicate if Rockstar has blocked/banned the isue from chat or if they reject the message? I remember Yim telling me something about how he removed the Rockstar API hit for incoming/outgoing chat requests and manually just pushed it to the screen, but I guess they sped up the response time of the API as it seems OK now when I play without Yim. Are these spammers just getting passed Rockstar's defenses, or are we still manually pushing the text to the screen?

maybegreat48:
> @maybegreat48 Would the return value of receive_net_message indicate if Rockstar has blocked/banned the isue from chat or if they reject the message? I remember Yim telling me something about how he removed the Rockstar API hit for incoming/outgoing chat requests and manually just pushed it to the screen, but I guess they sped up the response time of the API as it seems OK now when I play without Yim. Are these spammers just getting passed Rockstar's defenses, or are we still manually pushing the text to the screen?

The hook actually calls the original for `MsgTextMessage`, so we're already filtering messages though the Rockstar API, which obviously does not filter spam. The original handler has started to report messages through metrics, so it might be a good idea to revert the change and just display messages directly

gir489returns:
> The hook actually calls the original for `MsgTextMessage`, so we're already filtering messages though the Rockstar API

After we run the static filter, which this is updating. Shouldn't it check against Rockstar's API first, then do static analysis?

maybegreat48:
> After we run the static filter, which this is updating. Shouldn't it check against Rockstar's API first, then do static analysis?

Rockstar's API doesn't actually block any of these, so it wouldn't be useful

src/util/chat.hpp
```diff
@@ -72,6 +72,8 @@ namespace
 	    "0 Ban Risk",
 	    "Discord For Cheap Money",
 	    "10-30m",
+	    "Hey Guys! Tired of being poor?",
+	    "GTA CASH",
```
maybegreat48: While you're at it. can you block `情色影院：сб54.сo | >`? Note that the "6" here is actually the Cyrillic letter for B

Rxann: Yeah sure

gir489returns: Question! Instead of having .gg and .GG why don't we just cast the incoming message tolower() and then compare against the list? I don't know how this will affect the 1989 Tiananmen Square Massacre languages, but you know. It would solve at least capitalization permutations.

Rxann: I can implement this

Rxann: Should I implement a manual check(not in the list, and in the for loop itself for the cyrillic text) if the tolower messes it up? Another list that isn't casted could be made too. Let me know your opinions.

ghost:
Apparently GitHub didn’t like obscure ppl’s insults and deleted both issues including #3202

src/util/chat.hpp
```diff
@@ -116,6 +112,7 @@ namespace
 	    "\xE5\x81\xB7\xE6\x8B\x8D",             // AV
 	    "\xE4\xBC\xA0\xE7\x85\xA4",             // AV
 	    "\xE4\xB9\xB1\xE8\xAE\xBA",             // AV
+	    "情色影院：сб54.сo | >",                // 6 is cyrillic letter for B
```
maybegreat48: You're supposed to convert it to hex bytes like the others up there. I recommend sigging just `情色` (erotic in Chinese) and not the whole message i.e. `\xE6\x83\x85\xE8\x89\xB2`

maybegreat48 (approved):

Yimura (approved):

PR #3207, opened 2024-05-31
Input for xml transport

ghost:
I think it will be useful for those people who have 1000+ XML transports (including myself).
pic1
pic2


gir489returns:
Why not implement the same thing I did in #2042?

Yimura (approved):

PR #2488, opened 2023-12-06
Added more spam messages to the filter.

gir489returns:

PR #3250, opened 2024-06-12
improve(chat filter): Block all anonymous chat messages

sch-lda:
Although I don't know why, there are many spammers that yimmenu cant get their player info, and some whose names can't even be found in the player list. These ads are not filtered to be blocked and still show up on the screen, interfering with the gaming experience
After testing, I don't see any messages from normal players reaching here, so I think we can just return true

image

image

### DIFF
--- a/src/hooks/protections/receive_net_message.cpp
+++ b/src/hooks/protections/receive_net_message.cpp
@@ -262,6 +262,7 @@ namespace big
 				}
 				break;
 			}
+			case rage::eNetMessage::MsgTextMessage: return true;
 			case rage::eNetMessage::MsgNonPhysicalData:
 			{
 				buffer.Read<int>(7); // size


lonelybud:
where the player is not known i am only able to figure out their rockstar id. This anonymous messages will only happen when you are joining session and not joined yet. After you have joined their names will appear in list or they might have left by that time.

so you have basically handled the case of spam where player has not been added in player list

gir489returns (approved):
I remember asking @maybegreat48 to look into this ages ago, but all he said was that the player was null and didn't know how to deal with that situation.

lonelybud:
@sch-lda just wanted to know how much time do you think it takes on an average to type one Chinese letter using keyboard. I was thinking of improving the https://github.com/YimMenu/YimMenu/pull/2488 like based on ratio (how many characters are typed divided by time difference). On the internet they say 1 character per second.

sch-lda:
> @sch-lda just wanted to know how much time do you think it takes on an average to type one Chinese letter using keyboard. I was thinking of improving the #2488 like based on ratio (how many characters are typed divided by time difference). On the internet they say 1 character per second.

Typing speeds vary widely, but are generally in the range of 30-80 words per minute. Court stenographers may be able to reach 200 WPM

sch-lda:
> Do you have a chat.log you could share that shows people rapidly dispatching Chinese chats? We could probably calculate an average time between chats with the timestamps.

It may not be easy to find a scenario where a person is continuously sending messages rather than engaging in a dialogue. Due to the uncertainty of in-game matchmaking, players do not know the languages of others, so they typically use English for simple communication. Other players tend to form teams within chat groups on platforms like QQ or Discord, and directly utilize the voice chat functions provided by these tools for more efficient communication.

As for public sessions, due to the constant spam, there are very few people willing to chat within  the game

PR #3251, opened 2024-06-13
Added loop to kick reaction

### LINKED ISSUE #3120: [Bug]: Problems with reaction kick upon joining
### Describe the bug

It's not that this is a bug, but when I have the Modder Detection/Kick option enabled and I join with token spoofing and host auto-kick, players who are identified as modders (for example, those with desync protection) are not kicked during my connection but remain in the session. As far as I understand, when I join the session, a desync kick is triggered, and before I manage to become the host or something like that, they are protected from it, the kick doesn't have time to do its job

### Steps To Reproduce

1. Enable Modder Detection/Kick reaction
2. Enable Token Spoofing and Auto-kick host
3. Join random session with modders(which have an enabled function Desync Protection or Super Jump or etc.)
4. During the joining, we observe that they are identified as modders, but they do not leave even after you become the host

### Expected Behavior

Modders should be kicked the hell out. Maybe we should delay the reactions until a certain moment? For example, like with the auto-kick of the host?

### Logs

Logs are not needed, the bug is fully reproducible

### Store

Steam

### Language

English

### YimMenu Language

English (US)

### Additional context

_No response_

### Build

### Coload

ghost:
Closes #3120
Just sends a kick until the player disappears or we do.

### DIFF
--- a/src/backend/reactions/reaction.cpp
+++ b/src/backend/reactions/reaction.cpp
@@ -34,7 +34,11 @@ namespace big
 		if (kick)
 		{
 			g_fiber_pool->queue_job([player] {
-				dynamic_cast<player_command*>(command::get("smartkick"_J))->call(player, {});
+				while (player && player->get_net_data() && player->is_valid())
+				{
+					dynamic_cast<player_command*>(command::get("smartkick"_J))->call(player, {});
+					script::get_current()->yield(700ms);
+				}
 			});
 		}
 


lonelybud:
<del>its not what you think like. You should know that smart kick implements OOM & desync by default. Breakup kick will work when you are host. OOM kick works definitely. Desync kick is implemented in loop already.</del>

does not seems to be good solution.

ghost:
> ~its not what you think like. You should know that smart kick implements OOM & desync by default. Breakup kick will work when you are host. OOM kick works definitely. Desync kick is implemented in loop already.~
> 
> does not seems to be good solution.

during joining, desync is triggered, as soon as we become the host, desync stops and the player still remains in the session. and in this case both the desync and then the breakup will work.

P.S. it would be possible to implement something like a delay after we load into the session and then call kicks, but this sounds like nonsense

maybegreat48:
> Just sends a kick until the player disappears or we do.

Sending a kick 100 times wouldn't change anything

gir489returns:
This is a pointless change.

PR #3257, opened 2024-06-16
Filtering blocked modders in session browser

### LINKED ISSUE #3222: [Improvement]: Increase your chances of joining
### Which feature

How do we know that in order to successfully be in a session and not be kicked, you need to become a session host, right? But considering that we are not alone, I suggest adding a filter for modder sessions to the session browser. how will this work? We have a database of players whom we mark as modders, if the modder sessions filter checkbox is active, then we will exclude sessions hosted by a modder. this will be very useful for those who already have a fairly large database of cheaters (for me, for example). But how do we know that a modder owns the session? Without joining, we can get the host's RID and compare it with the database and then filter it. At least that’s what a stand can do (it can get all the information about the session and also the host’s nickname).

### Reason

This way we will reduce the time required to find a suitable session in which we can become the host and will not be desynchronized by others.

### Additional context

_No response_

ghost:
Sorry for the previous closed PR, I wanted to add another commit and accidentally closed the PR 😔
Closes #3222
It will be very useful for those who play in the Chinese region. The positive effect will be noticeable if you block all chat spammers from joining as they often host a lobby with a multiplexer and you will surf in sessions as often as possible. The screenshots show how we delete constantly duplicate lobbies with the same spammer
pic1
pic2

### DIFF
--- a/src/core/settings.hpp
+++ b/src/core/settings.hpp
@@ -1041,9 +1041,9 @@ namespace big
 			int sort_direction = 0;
 
 			bool replace_game_matchmaking = false;
-			bool without_host_modders     = false;
+			bool exclude_modder_sessions     = false;
 
-			NLOHMANN_DEFINE_TYPE_INTRUSIVE(session_browser, region_filter_enabled, region_filter, language_filter_enabled, language_filter, player_count_filter_enabled, player_count_filter_minimum, player_count_filter_maximum, sort_method, sort_direction, replace_game_matchmaking, pool_filter_enabled, pool_filter, without_host_modders)
+			NLOHMANN_DEFINE_TYPE_INTRUSIVE(session_browser, region_filter_enabled, region_filter, language_filter_enabled, language_filter, player_count_filter_enabled, player_count_filter_minimum, player_count_filter_maximum, sort_method, sort_direction, replace_game_matchmaking, pool_filter_enabled, pool_filter, exclude_modder_sessions)
 		} session_browser{};
 
 		struct session_protection
--- a/src/core/settings.hpp
+++ b/src/core/settings.hpp
@@ -1041,8 +1041,9 @@ namespace big
 			int sort_direction = 0;
 
 			bool replace_game_matchmaking = false;
+			bool without_host_modders     = false;
 
-			NLOHMANN_DEFINE_TYPE_INTRUSIVE(session_browser, region_filter_enabled, region_filter, language_filter_enabled, language_filter, player_count_filter_enabled, player_count_filter_minimum, player_count_filter_maximum, sort_method, sort_direction, replace_game_matchmaking, pool_filter_enabled, pool_filter)
+			NLOHMANN_DEFINE_TYPE_INTRUSIVE(session_browser, region_filter_enabled, region_filter, language_filter_enabled, language_filter, player_count_filter_enabled, player_count_filter_minimum, player_count_filter_maximum, sort_method, sort_direction, replace_game_matchmaking, pool_filter_enabled, pool_filter, without_host_modders)
 		} session_browser{};
 
 		struct session_protection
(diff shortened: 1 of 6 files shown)


src/views/network/view_session_browser.cpp
```diff
@@ -33,6 +34,12 @@ namespace big
 					if (!session.is_valid)
 						continue;
 
+					auto host_rid = session.info.m_net_player_data.m_gamer_handle.m_rockstar_id;
+
+					if (g.session_browser.without_host_modders && g_player_database_service->get_player_by_rockstar_id(host_rid)
+					    && g_player_database_service->get_player_by_rockstar_id(host_rid)->block_join)
```
Yimura: You're calling `get_player_by_rockstar_id`, this can be reduced to a single function call and saved into a variable before composing this statement.

src/views/network/view_session_browser.cpp
```diff
@@ -158,6 +165,8 @@ namespace big
 				ImGui::Combo("###pooltype", &g.session_browser.pool_filter, pool_filter_options.c_str());
 			}
 
+			ImGui::Checkbox("WITHOUT_HOST_MODDERS"_T.data(), &g.session_browser.without_host_modders);
```
Yimura: No translation PR open yet.

src/hooks/matchmaking/start_matchmaking_find_sessions.cpp
```diff
@@ -28,6 +29,13 @@ namespace big
 					{
 						if (g_matchmaking_service->get_found_sessions()[i].is_valid)
 						{
+							auto host_rid =
+							    g_matchmaking_service->get_found_sessions()[i].info.m_net_player_data.m_gamer_handle.m_rockstar_id;
+
+							if (g.session_browser.without_host_modders && g_player_database_service->get_player_by_rockstar_id(host_rid)
+							    && g_player_database_service->get_player_by_rockstar_id(host_rid)->block_join)
```
Yimura: You're calling `get_player_by_rockstar_id`, this can be reduced to a single function call and saved into a variable before composing this statement.

src/core/settings.hpp
```diff
@@ -1041,8 +1041,9 @@ namespace big
 			int sort_direction = 0;
 
 			bool replace_game_matchmaking = false;
+			bool without_host_modders     = false;
 
-			NLOHMANN_DEFINE_TYPE_INTRUSIVE(session_browser, region_filter_enabled, region_filter, language_filter_enabled, language_filter, player_count_filter_enabled, player_count_filter_minimum, player_count_filter_maximum, sort_method, sort_direction, replace_game_matchmaking, pool_filter_enabled, pool_filter)
+			NLOHMANN_DEFINE_TYPE_INTRUSIVE(session_browser, region_filter_enabled, region_filter, language_filter_enabled, language_filter, player_count_filter_enabled, player_count_filter_minimum, player_count_filter_maximum, sort_method, sort_direction, replace_game_matchmaking, pool_filter_enabled, pool_filter, without_host_modders)
```
Yimura: `without_host_modders`, this variable just doesn't sound right. I obviously know what it means within the context of this PR.

My suggestion would be to change this to `exclude_modder_sessions`.

Yimura (changes requested):

src/hooks/matchmaking/start_matchmaking_find_sessions.cpp
Rxann: Instead of using the block_join wouldn't is_modder be better?

src/views/network/view_session_browser.cpp
Rxann: Instead of using the block_join wouldn't is_modder be better?

src/hooks/matchmaking/start_matchmaking_find_sessions.cpp
ghost: > Instead of using the block_join wouldn't is_modder be more fitted?

If we block a join, that means we must also block our connection to the player, sounds more logical, doesn’t it? Then let's just rename it to `EXCLUDE_MODDER_SESSIONS` and add a description.

Yimura (approved):

PR #3268, opened 2024-06-20
Filters to the player database

ghost:
A small number of filters to the player database and removed "Remove Untrusted"
pic1
pic2
pic3

### DIFF
--- a/src/services/player_database/player_database_service.cpp
+++ b/src/services/player_database/player_database_service.cpp
@@ -219,11 +219,13 @@ namespace big
 		return player;
 	}
 
-	void player_database_service::remove_untrusted_players()
+	void player_database_service::remove_filtered_players(bool filter_modder, bool filter_trust, bool filter_block_join, bool filter_track_player)
 	{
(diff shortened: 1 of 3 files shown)


binded-zz:
omg i tired to do this and failed so bad this is nice and neat looking 
not that i use chat commands but also one for command access would make it easier to find who  can do that if you set it for per user 
i would say by level but i doubt that matters

cringearab man thanks i hope this gets pulled

ghost:
> omg i tired to do this and failed so bad this is nice and neat looking not that i use chat commands but also one for command access would make it easier to find who can do that if you set it for per user i would say by level but i doubt that matters
> 
> cringearab man thanks i hope this gets pulled

I was thinking about adding infraction and command access level to the filter, but it seemed a bit too complicated for me. Anyway, it's not really necessary right now. I think we can consider this for the next PR or maybe someone else will do it.

binded-zz:
i was  going to ask about that very thing but i got yelled at last time  so i didnt do that this time :-)

Yimura (approved):

PR #3280, opened 2024-06-22
update outfit slot name global

USBMenus:
update the global for outfit slot names

### DIFF
--- a/src/util/outfit.hpp
+++ b/src/util/outfit.hpp
@@ -56,7 +56,7 @@ namespace big::outfit
 
 	inline char* get_slot_name_address(int slot)
 	{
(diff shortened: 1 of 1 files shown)


USBMenus:
i noticed in the decompiled scripts that the size of the first offset was `5569`, but `5568` seems to work the same

gir489returns (approved):

gir489returns:
> i noticed in the decompiled scripts that the size of the first offset was `5569`, but `5568` seems to work the same

That is because we add 1 to it, for using a size. Look at the implementation of at with 2 parameters in the scr_globals class.

Rxann (approved):

PR #3299, opened 2024-07-04
static link the vcruntime due to some people not having latest vcruntime dll installed on their machine

xiaoxiao921:


gir489returns:
Question. Why are we using the debug variant for release mode?

xiaoxiao921:
it appends Debug only if the selected config is Debug

gir489returns:
Our Debug build is busted, so I don't see any use case for using the D variant of the MSVC DLLs. I don't even know why we have the Debug build in the first place.

xiaoxiao921:
Yeah it is busted but can still serve some use in case someone wanna debug some stuff that get completly optimized away, it's just that so far we've been lazy and never bothered fixing the asserts that debug builds produce, but it's doable

Yimura (approved):

Yimura (approved):

PR #3300, opened 2024-07-04
CPlayerInfo update.

gir489returns:


gir489returns:
I tested this online by shooting someone, getting 1 star, and then clearing it. I haven't tested the other members of CPlayerInfo, but I only saw that there was 0x10 padding added between 2F0 and 4C0.

Yimura (approved):

PR #3301, opened 2024-07-05
feat: dynamicly hooking x64 functions from lua at runtime with arbitrary signatures.

xiaoxiao921:


gir489returns:
I like this feature, but does it really need ASMJIT? Can’t we do this with an existing dependency like MinHook or something?

Yimura (approved):
Gonna be interesting what kind of features people will implement with this.

Sure I get that this adds another dependency however the implementation was taken from somewhere else. Completely converting it to use the pre-existing dependencies in YimMenu may require more extensive work. 

It brings up the conversation on the current dependency stack and if what benefits there are to gain from switching it up.

PR #3322, opened 2024-07-09
Removes size requirement on view_world check boxes affecting translations.

gir489returns:


gir489returns:
@CringeArab I just pushed another commit to address the two menus of issue, hopefully that will solve the problem. If you find more spacing issues, let me know.

gir489returns:
@CringeArab I reduced it by another 100, however, I don't think that's going to work for the first translation it's way too long, anyway we can make it more succinct?

gir489returns:
@CringeArab I am not a huge fan of the translation now, because it doesn't explicitly say it's for joining notification, it (roughly?) translates to Blocking Notifications.

gir489returns:
> > @CringeArab I am not a huge fan of the translation now, because it doesn't explicitly say it's for joining notification, it (roughly?) translates to Blocking Notifications.
> 
> It's hard to describe it in two words. `Причина блокировки` will be better

You have to tell them it's join blocking notifications.

gir489returns:
Why don't we do this? "Предупреждение о\r\nблокировке присоединения"

It would put it on a new line, but it's better than having a translation that makes no fucking sense.

gir489returns:
image

Looks fine to me. Are there any other issues, @CringeArab before we merge this?

gir489returns:
@CringeArab Once the builder bot finishes this, would you mind going through the menu and seeing if you can find anything untranslated that shouldn't be left latinized?

PR #3325, opened 2024-07-09
more player stats lua bindings

USBMenus:
adds 5 more lua bindings in network
get_player_money
get_player_wallet
get_player_bank
get_player_language_id
get_player_language_name

### DIFF
--- a/src/lua/bindings/network.cpp
+++ b/src/lua/bindings/network.cpp
@@ -1,6 +1,7 @@
 #include "network.hpp"
 
 #include "../../script.hpp"
+#include "core/data/language_codes.hpp"
 #include "core/scr_globals.hpp"
 #include "hooking/hooking.hpp"
 #include "pointers.hpp"
@@ -12,6 +13,7 @@
 #include "util/system.hpp"
 #include "util/teleport.hpp"
 #include <script/globals/GPBD_FM.hpp>
+#include <script/globals/GPBD_FM_3.hpp>
 
 namespace lua::network
 {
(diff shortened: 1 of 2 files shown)


src/lua/bindings/network.cpp
```diff
@@ -234,6 +236,81 @@ namespace lua::network
...
+		}
+		return -1;
+	}
+
+	// Lua API: Function
+	// Table: network
+	// Name: get_player_language_name
+	// Param: pid: integer: Index of the player.
+	// Call get_player_language_name(playerID)
+	static std::string get_player_language_name(int pid)
```
maybegreat48, line 304: Can't you do this in Lua?

gir489returns: I made the same argument for literally any of these in his last PR.

USBMenus: > Can't you do this in Lua?

yes, but that would require the globals to be updated every time there's an update, this way it's easier for scripts to stay up to date

maybegreat48: >  this way it's easier for scripts to stay up to date

I'm talking about the language id to string conversion

USBMenus: > I'm talking about the language id to string conversion

oh mb, yea you could create a table and use the id for the index, i can remove it if you want

gir489returns: Same argument here applies as the globals, you could do it in Lua, but we should be consistent. If we're exposing crap internally to the Lua layer they could be doing as an assistant, then we should expose as much as possible.

gir489returns (approved):

PR #3346, opened 2024-07-14
Fix host token logic.

### LINKED ISSUE #3345: [Bug]: false positive spoofed host token reaction when yim user is host
### Describe the bug

https://github.com/YimMenu/YimMenu/blob/796f6209f5bb05e0c311470270b74fa2ab893752/src/hooks/misc/add_gamer_to_session.cpp#L17

I believe there is a bug here. It should be `g.session.original_host_token  & 0xFFFFFFFF00000000`

### Steps To Reproduce

host a session with HIDE_TOKEN_SPOOFING_WHEN_HOST checked (true by default) and dont spoof host token
ask any yim user to join
note that they will get reaction that you spoofed host token

### Expected Behavior

should not trigger false positive due to incorrect peer value sent by us during spoofing. 

### Logs

.

### Store

Steam

### Language

English

### YimMenu Language

English (US)

### Additional context

_No response_

### Build

### Coload

gir489returns:
Closes #3345

### DIFF
--- a/src/hooks/misc/add_gamer_to_session.cpp
+++ b/src/hooks/misc/add_gamer_to_session.cpp
@@ -14,7 +14,7 @@ namespace big
 		    && info->m_gamer_info.m_gamer_handle.m_rockstar_id == g_player_service->get_self()->get_rockstar_id())
 		{
 			info->m_gamer_info.m_host_token = g.session.original_host_token;
-			info->m_gamer_info.m_peer_id = (info->m_gamer_info.m_peer_id & 0xFFFFFFFF) | (g.session.original_host_token << 32); // TODO: P2pSecurity info message
+			info->m_gamer_info.m_peer_id = (info->m_gamer_info.m_peer_id & 0xFFFFFFFF) | (g.session.original_host_token & 0xFFFFFFFF00000000); // TODO: P2pSecurity info message
 		}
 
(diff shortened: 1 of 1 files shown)


src/hooks/misc/add_gamer_to_session.cpp
```diff
@@ -14,7 +14,7 @@ namespace big
 		    && info->m_gamer_info.m_gamer_handle.m_rockstar_id == g_player_service->get_self()->get_rockstar_id())
 		{
 			info->m_gamer_info.m_host_token = g.session.original_host_token;
-			info->m_gamer_info.m_peer_id = (info->m_gamer_info.m_peer_id & 0xFFFFFFFF) | (g.session.original_host_token << 32); // TODO: P2pSecurity info message
```
maybegreat48, line 17: Or `>> 32`, but both are fine

lonelybud: `>> 32` will move the higher 32 bits to bottom but we want to set the lower 32 bits to 0. The pr looks fine.

gir489returns: >>  32 will require the compiler to generate a ROR instruction, as opposed to just bit-banging the upper 32-bits.

Yimura (approved):

PR #3348, opened 2024-07-14
YimCLI

Tayipc38:
Ill be trying to make this thing actually work nice.

Lets be fancy and call this YimCLI

### DIFF
--- a/src/backend/commands/session/join_player.cpp
+++ b/src/backend/commands/session/join_player.cpp
@@ -1,0 +1,96 @@
+#include "backend/command.hpp"
+#include "network/CNetworkPlayerMgr.hpp"
+#include "pointers.hpp"
+#include "services/players/player_service.hpp"
+#include "socialclub/FriendRegistry.hpp"
+#include "services/player_database/player_database_service.hpp"
+#include "services/api/api_service.hpp"
+#include "util/session.hpp"
+
+namespace big
+{
+	class join_player : command
+	{
+		using command::command;
+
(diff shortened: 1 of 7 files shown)


ghost:
I tried to type the entire word SINGLEPLAYER but after pressing TAB for autocomplete I see this
mv

Tayipc38:
> I tried to type the entire word SINGLEPLAYER but after pressing TAB for autocomplete I see this
> mv
> 

Yes im aware of this. Its got worse cases, im trying to figure out why.

src/backend/commands/session/join_player.cpp
```diff
@@ -0,0 +1,96 @@
...
+			{
+				for (const auto& [name, player] : g_player_database_service->get_sorted_players())
+					if (name == args[0])
+						rid = player->rockstar_id;
+			}
+
+			// If the player is not a friend or in the database, fetch rid from the API
+			if (rid == 0)
+			{
+				auto fetched = g_api_service->get_rid_from_username(args[0], rid);
```
Yimura, line 66: I know we coded this together, however this will cause a deadlock until the Rockstar API returns some data.

If we want to make this work we'll have to make it so suggestions can be given in an asynchronous way.
Alternatively if we were to support std::string (mainly making sure that the string survives longer than the argument creation context), then we could pass the string to the execute method.

Or as a botch fix we can make a member variable to pass data from the parse method to the execute method.

Yimura (approved):

PR #3363, opened 2024-07-16
Refactor no_idle_kick to not store the address of the tunables.

### LINKED ISSUE #3347: [Bug]: Tunable library can sometimes return a pointer that points to an invalid section of memory.
### Describe the bug

While playing SP and loading into a save, I noticed an error inside my developers console. It's complaining that one of the tunable addresses for idle kick was invalid. We already do a nullptr check, so it must be returning a non-null value that points to an invalid section of memory.

### Steps To Reproduce

1. Go into SP.
2. Load a save.
3. Bug appears.

### Expected Behavior

Should return a nullptr if the engine is loading or the pointers need to be refreshed.

### Logs

[cout.log](https://github.com/user-attachments/files/16215446/cout.log)

### Store

Steam

### Language

English

### YimMenu Language

English (US)

### Additional context

_No response_

### Build

### Coload

gir489returns:
Closes #3347

### DIFF
--- a/src/backend/looped/tunables/no_idle_kick.cpp
+++ b/src/backend/looped/tunables/no_idle_kick.cpp
@@ -30,7 +30,7 @@ namespace big
 						did_fail = true;
 					}
 				}
-				m_backed_up = did_fail;
+				m_backed_up = !did_fail;
 			}
 			else
 			{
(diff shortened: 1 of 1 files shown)


Yimura (approved):

src/backend/looped/tunables/no_idle_kick.cpp
```diff
@@ -1,56 +1,61 @@
...
 		}
 
 		virtual void on_disable() override
 		{
-			for (int i = 0; m_ready_to_use && i < m_restore.size(); ++i)
+			if (m_backed_up)
 			{
-				if (m_tunables[i])
-					*m_tunables[i] = m_restore[i];
+				*g_tunables_service->get_tunable<int*>("IDLEKICK_WARNING1"_J)        = m_restore[0];
```
lonelybud: could probably convert to -
```C++
				for (int i = 0; i < m_restore.size(); ++i)
					*g_tunables_service->get_tunable<int*>(m_tunable_hashes[i]) = m_restore[i];
```

Yimura (approved):

PR #3389, opened 2024-07-20
Revert changes to Superherofly

Tayipc38:
The changes done after the initial contribution broke more than half of the features in superherofly.

### DIFF
--- a/src/backend/looped/self/super_hero_fly.cpp
+++ b/src/backend/looped/self/super_hero_fly.cpp
@@ -399,7 +399,7 @@ namespace big
 
 		virtual void on_enable() override
 		{
-			g_notification_service.push("Super hero fly", "SUPER_HERO_FLY_ENABLE_NOTIFICATION"_T.data());
+			g_notification_service.push("SUPER_HERO_FLY"_T.data(), "SUPER_HERO_FLY_ENABLE_NOTIFICATION"_T.data());
 		}
 
 		virtual void on_disable() override
--- a/src/backend/looped/self/super_hero_fly.cpp
+++ b/src/backend/looped/self/super_hero_fly.cpp
@@ -1,6 +1,6 @@
-#include "backend/looped_command.hpp"
 #include "backend/bool_command.hpp"
 #include "backend/float_command.hpp"
+#include "backend/looped_command.hpp"
 #include "gta/enums.hpp"
 #include "natives.hpp"
 #include "util/entity.hpp"
@@ -22,7 +22,7 @@ namespace big
 
 		bool m_timer = false;
 		std::chrono::system_clock::time_point m_time_stamp;
-		std::chrono::milliseconds m_duration;
+		std::chrono::duration<double> m_duration;
 
 		bool m_charging             = false;
 		double m_charge_intensity   = 1;
@@ -239,6 +239,14 @@ namespace big
 
 		virtual void on_tick() override
 		{
+			if (PED::IS_PED_IN_ANY_VEHICLE(self::ped, false))
+			{
+				if (m_flying || m_charging || m_landing || m_launching)
+					reset();
+
+				return;
+			}
+
 			//Both anims are required in order to consider the player ped flying.
 			m_anims_are_playing = ENTITY::IS_ENTITY_PLAYING_ANIM(self::ped, "missfam5_yoga", "c8_to_start", 3) && ENTITY::IS_ENTITY_PLAYING_ANIM(self::ped, "skydive@parachute@first_person", "chute_idle_alt_lookright", 3);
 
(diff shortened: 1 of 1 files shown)


iwanttodo1t:
I really liked the previous flying function, it was more fun, but this superhero mode feels no different from free movement.
Don't know why that was deleted.

src/backend/looped/self/super_hero_fly.cpp
```diff
@@ -391,7 +399,7 @@ namespace big
 
 		virtual void on_enable() override
 		{
-			g_notification_service.push("SUPER_HERO_FLY"_T.data(), "SUPER_HERO_FLY_ENABLE_NOTIFICATION"_T.data());
+			g_notification_service.push("Super hero fly", "SUPER_HERO_FLY_ENABLE_NOTIFICATION"_T.data());
```
xiaoxiao921: No

xiaoxiao921 (changes requested):

src/backend/looped/self/super_hero_fly.cpp
Yimura: No

Tayipc38: Lol

Tayipc38:
> I really liked the previous flying function, it was more fun, but this superhero mode feels no different from free movement. Don't know why that was deleted.

Could you try it again after this revert? I do not share your opinion.

Yimura (approved):

src/backend/looped/self/super_hero_fly.cpp
```diff
@@ -22,7 +22,7 @@ namespace big
 
 		bool m_timer = false;
 		std::chrono::system_clock::time_point m_time_stamp;
-		std::chrono::milliseconds m_duration;
+		std::chrono::duration<double> m_duration;
```
gir489returns, line 25: std::chrono::milliseconds types as std::chrono::duration<long long>, why are you changing it to [double](https://en.wikipedia.org/wiki/Double-precision_floating-point_format), which is a double precision float?

src/backend/looped/self/super_hero_fly.cpp
```diff
@@ -249,7 +257,7 @@ namespace big
 			PAD::DISABLE_CONTROL_ACTION(0, (int)ControllerInputs::INPUT_COVER, false);
 
 			if (m_timer)
-				m_duration = std::chrono::duration_cast<std::chrono::milliseconds> (std::chrono::system_clock::now() - m_time_stamp);
+				m_duration = std::chrono::system_clock::now() - m_time_stamp;
```
gir489returns, line 260: Why? If you're worried about accuracy being lost in calculations between calls, you can keep a std::chrono::high_resolution_clock type, and then set the m_duration to a std::chrono::duration_cast at the last moment.

src/backend/looped/self/super_hero_fly.cpp
```diff
@@ -289,7 +297,7 @@ namespace big
 			}
 
 			if (PAD::IS_DISABLED_CONTROL_PRESSED(0, (int)ControllerInputs::INPUT_COVER) && !m_flying
-			    && g.self.super_hero_fly.charge)
+			    && g.self.super_hero_fly.charge && !PED::IS_PED_IN_ANY_VEHICLE(self::ped, false))
```
gir489returns, line 300: This change doesn't make any sense, you added a return check at the head of the function to see if you're in any vehicle, why are you checking it again?

gir489returns (changes requested):

src/backend/looped/self/super_hero_fly.cpp
Tayipc38: One of these changes fixes the features, just reverted it.

src/backend/looped/self/super_hero_fly.cpp
Tayipc38: Funny thing, it still won't work lol

Tayipc38:
Btw, I might try to make this a lua instead, sounds much more appropriate for a feature such as this.

gir489returns:
> Btw, I might try to make this a lua instead, sounds much more appropriate for a feature such as this.

I'll take your word that this has been tested and fixed, although like you mentioned this might be better suited as a Lua script. Yim already approved the code, and the problems I have with it are lexical, however, given the fact this is a "fun" feature and not a core element, I think we can let code quality slide.

src/backend/looped/self/super_hero_fly.cpp
iwanttodo1t: I use this to confirm that the player is inside the vehicle.
scr_globals::globalplayer_bd.as<GlobalPlayerBD*>()->Entries[self::id].PlayerBlip.NumPassengersInVehicle > 0

PR #3404, opened 2024-07-22
Sanitized the project's includes.

gir489returns:
I benchmarked both before and after this change, the compile time did not change, but the link time dropped by 3 seconds on FastCompile. Negligible, but this could have wider ramifications in both development and deployment eventually.

These are only the headers that were not used, or I was able to optimize by using direct headers. If we can figure out which ones are circularly dependent, things that include the settings.hpp when they do not need to, then development could be sped up, instead of 559 CPP files that need to be recompiled, we could probably get that down to 300 or so.

If someone has a heat-map of the CPP files and their headers that are the worst offenders, that would help out a lot, I couldn't run Visual Studio's tools because it's a CMake project, and I don't know how to run it against just YimMenu, it wants to run it against the entire CMake build which encompasses our 9 dependencies and other things we have no control over.


Yimura (approved):
It still compiles so I see this as a win.

SkiddyToast (approved):
LGTM.

PR #3430, opened 2024-07-25
Added Use Weapon's Max Range to Aimbot

gir489returns:


xiaoxiao921 (approved):

src/backend/looped/session/spoof_host_token.cpp
```diff
@@ -19,7 +19,8 @@ namespace big
 
 		g_pointers->m_gta.m_profile_gamer_info->m_peer_id                                          = *g_pointers->m_gta.m_peer_id;
 		g_pointers->m_gta.m_player_info_gamer_info->m_peer_id = *g_pointers->m_gta.m_peer_id;
-		(*g_pointers->m_gta.m_communications)->m_voice.m_connections[0]->m_gamer_info.m_peer_id = *g_pointers->m_gta.m_peer_id;
+		if (g_pointers->m_gta.m_communications != nullptr)
```
lonelybud, line 22: if it is null then token should not be allowed to spoof right?

gir489returns: It was only occurring when ejecting/reinjecting in SP, it probably happened due to the fact this was called out of lockstep with what it should be. Sanity check wouldn't hurt.

Yimura (approved):

PR #3431, opened 2024-07-25
feat(debug): Add Tunables Tab

ShinyWasabi:
Added a Tunables tab similar to Globals/Locals that can be used to read and write tunables by inputting their unsigned/signed or hexadecimal hash, as well as their pre-joaat name. It also displays the offset of the given tunable.

Removed the animation players in the Debug tab as the one in the Self tab makes them redundant.


src/services/tunables/tunables_service.hpp
```diff
@@ -6,6 +6,8 @@
 
 namespace big
 {
+	constexpr int TUNABLE_BASE_ADDRESS = 262145; // This never changes
```
gir489returns, line 9: I think the math might be wrong for this here. The Tunables globals starts at 0x40000, and the first tunable is at 0x40001. 

The way the tunables service generates the indexes is `*script_global(0x40000).at(i).as<int*>();`

So, shouldn't the base address be 0x40000?

gir489returns (changes requested):

Yimura (dismissed):

src/services/tunables/tunables_service.hpp
ShinyWasabi: 0x40000 (262144) will return offset + 1, which is wrong.

This should be 0, not 1.
```c
Global_262145 = NETWORK::_NETWORK_GET_TUNABLES_REGISTRATION_FLOAT(joaat("CASH_MULTIPLIER"), 1f);
```

But I think I can change it to 0x40000 for consistency, and add `(offset - TUNABLE_BASE_ADDRESS) - 1` to the get_tunable_offset function.

Yimura (approved):

PR #3434, opened 2024-07-25
Fixed mission cars not being properly cleaned up.

### LINKED ISSUE #3427: [Bug]: Vehicles not being cleaned up during mission restarts
### Describe the bug

When a mission or heist is restarted (due to a player dying or an objective failing for example), your vehicle does not get removed by the game.

In the past it sometimes showed a notification saying "xyz tried to delete your vehicle", but nowadays it doesnt even show that.

**For example:** the driller failed during the Fleeca Bank job and we restarted from a checkpoint. The Armored Kuruma I was sitting in was not cleaned up by the game, and now its just sitting there as an extra unused vehicle.

car bug

### Steps To Reproduce

1. Start any mission/heist
2. Fail mid-way and restart from checkpoint
3. Your vehicle, regardless of whether you're sitting in or not, get carried over into the restart

### Expected Behavior

The game should delete them.

### Logs

[cout.log](https://github.com/user-attachments/files/16374439/cout.log)

### Store

Steam

### Language

English

### YimMenu Language

English (US)

### Additional context

I have observed this with and without my LUAs being active.

I have all my protection settings disabled (I do not play in public lobbies anyway). I have tried resetting and then disabling them again as well.

I have tried resetting my entire Yim settings as well.

### Build

### Coload

gir489returns:
Closes #3427

### DIFF
--- a/src/hooks/protections/received_clone_remove.cpp
+++ b/src/hooks/protections/received_clone_remove.cpp
@@ -15,11 +15,14 @@ namespace big
 		if (g_local_player && g_local_player->m_vehicle && g_local_player->m_vehicle->m_net_object
 		    && object_id == g_local_player->m_vehicle->m_net_object->m_object_id) [[unlikely]]
 		{
-			if (auto plyr = g_player_service->get_by_id(src->m_player_id))
+			if (!NETWORK::NETWORK_IS_ACTIVITY_SESSION()) //If we're in Freemode.
 			{
-				g.reactions.delete_vehicle.process(plyr);
+				if (auto plyr = g_player_service->get_by_id(src->m_player_id))
+				{
+					g.reactions.delete_vehicle.process(plyr);
+				}
+				return;
 			}
-			return;
 		}
 
 		if (auto object = g_pointers->m_gta.m_get_net_object(*g_pointers->m_gta.m_network_object_mgr, object_id, true)) [[likely]]


Yimura (dismissed):

src/hooks/protections/received_clone_create.cpp
```diff
@@ -1,90 +1,93 @@
...
+		{
+			if (!NETWORK::NETWORK_IS_ACTIVITY_SESSION())     //If we're in Freemode.
+			{
+				if (auto plyr = g_player_service->get_by_id(src->m_player_id))
+				{
+					g.reactions.delete_vehicle.process(plyr);
+				}
+				return;
+			}
+		}
```
gir489returns, line 38: @maybegreat48 Why are we checking for a delete scenario in CLONE_CREATE?

gir489returns: This is a malicious scenario and can be only called by modders.

gir489returns:
> The line endings have changed but LGTM otherwise

Yeah you can thank @Yimura for that amazing change to git. No matter what you do now, it will force every file to be in CR LF. We should just do a giant replace of all the files so we get it over with.

xiaoxiao921:
If such commit happens, you can make it so that the commit doesn't affect the git blame.

Yimura (approved):

gir489returns:
@Yimura The user who reported the original bug has confirmed this fixed it.

Issue #3437, opened 2024-07-26
[Bug]: Cant join ANY Player when Menu is injected


ghost:
### Describe the bug

You cant join players via the player database. It will always say Session doesnt exist anymore.
Also happens to friends if you join via the ingame Tabs
I can join my friends 3/6 times via ingame menu when i dont ahve Yim injected
The log is from a crash i had 2 mins before writing this. I crashed when trying to tp a player tp which also didnt work 

### Steps To Reproduce

1. Go to Playerdatabase and join any player 
2. Wont work and will always fail

### Expected Behavior

Join the player i want to join

### Logs

[cout.log](https://github.com/user-attachments/files/16388521/cout.log)

### Store

Steam

### Language

English

### YimMenu Language

English (US)

### Build

- [X] I've confirmed the issue exists on the latest version of YimMenu

### Coload

- [X] I've tested this without co-loading anything.

ghost:
everything works great for me, maybe the people you are trying to join are on a mission
![image](https://github.com/user-attachments/assets/f464bd93-ae78-4b06-bbe2-876891d9ab25)

or the host is blocking your connection

gir489returns:
> everything works great for me

Same.

Kibasmoothie:
Ill guess its just Gta bugging for me ill close the bug and move on lol. I mean i can get around by injecting AFTER joining my friend

ghost:
Ill guess its just Gta bugging for me ill close the bug and move on lol. I mean i can get around by injecting AFTER joining my friend

PR #3444, opened 2024-07-26
Don't mark players in missions as joinable

ghost:
We mark players in yellow who are in a mission and we will not be able to join them. Related to #3437
image


gir489returns (changes requested):
Can you use ImGui::BeginDisabled/ImGui::EndDisabled if is_joinable_session fails on the Join Session button? Disabling the button entirely makes more sense than having it silently do nothing, that's bad UX.

ghost:
now it looks better

anim

gir489returns (approved):
image

PR #3450, opened 2024-07-27
Kick host during join

ghost:
I don't know why you removed this useful feature. I suggest bringing it back.

anim

### DIFF
--- a/src/backend/looped/session/auto_kick_host.cpp
+++ b/src/backend/looped/session/auto_kick_host.cpp
@@ -13,8 +13,9 @@ namespace big
 		{
 			g_player_service->iterate([](auto& plyr) {
 				// Don't kick trusted players
-				if (plyr.second->is_trusted || (g.session.trust_friends && plyr.second->is_friend()))
-					return;
+				if (plyr.second->is_trusted || (g.session.trust_friends && plyr.second->is_friend()) 
+					|| (plyr.second->is_modder && g.session.exclude_modders_from_kick_host))
+						return;
 
 				if (plyr.second->is_host())
 				{
--- a/src/backend/looped/session/auto_kick_host.cpp
+++ b/src/backend/looped/session/auto_kick_host.cpp
@@ -8,7 +8,7 @@ namespace big
 	static bool bLastKickHost = false;
 	void looped::session_auto_kick_host()
 	{
-		bool kick_host = false; // *g_pointers->m_gta.m_is_session_started && g.session.force_session_host && g.session.kick_host_when_forcing_host;
+		bool kick_host = *g_pointers->m_gta.m_is_session_started && g.session.spoof_host_token_type != 0 && g.session.kick_host_when_forcing_host;
 		if (kick_host && !bLastKickHost) [[unlikely]]
 		{
 			g_player_service->iterate([](auto& plyr) {
--- a/src/backend/looped/vehicle/allow_all_vehicles_in_heists.cpp
+++ b/src/backend/looped/vehicle/allow_all_vehicles_in_heists.cpp
@@ -18,10 +18,8 @@ namespace big
 			{
 				if (auto tunable_ptr = g_tunables_service->get_tunable<PBOOL>(tunable_hash)) [[likely]]
 				{
-					if (*tunable_ptr != FALSE)
-					{
+					if (*tunable_ptr != FALSE) [[unlikely]]
 						*tunable_ptr = FALSE;
-					}
 				}
 			}
 		}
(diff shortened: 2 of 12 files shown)


Yimura (dismissed):

snowleopardd:
Does this check if the host is a detected modder and is spoofing their token?
otherwise this will lead to us getting into a lot of databases + kicked + blocked, these simple checks should be a must tbh

ghost:
> Does this check if the host is a detected modder and is spoofing their token? otherwise this will lead to us getting into a lot of databases + kicked + blocked, these simple checks should be a must tbh

why do we need to check the host to see if he is a modder or what? don't kick the host? if he is a modder, then we won't last more than a minute in the session, and it will have to be changed anyway.

snowleopardd:
> > Does this check if the host is a detected modder and is spoofing their token? otherwise this will lead to us getting into a lot of databases + kicked + blocked, these simple checks should be a must tbh
> 
> why do we need to check the host to see if he is a modder or what? don't kick the host? if he is a modder, then we won't last more than a minute in the session, and it will have to be changed anyway.

Thats it, there's no need and is 100% useless to try to kick a host token spoofer if we will end up getting kicked and permanently blocked by these guys, we got the tools and built-in detections to prevent that, why not?

ghost:
> Thats it, there's no need and is 100% useless to try to kick a host token spoofer if we will end up getting kicked and permanently blocked by these guys, we got the tools and built-in detections to prevent that, why not?

I still don't understand what you want

snowleopardd:
> > Thats it, there's no need and is 100% useless to try to kick a host token spoofer if we will end up getting kicked and permanently blocked by these guys, we got the tools and built-in detections to prevent that, why not?
> 
> I still don't understand what you want

I'm just suggesting adding a check to prevent automatically kicking host token spoofers who are already host, the less we are in their databases, the better, no?, also what if the spoofer is a good guy but we trigger their reactions and get permanently blocked from their sessions?, there's nothing we can lose from adding this logic and it helps a lot...

ghost:
> I'm just suggesting adding a check to prevent automatically kicking host token spoofers who are already host, the less we are in their databases, the better, no?, also what if the spoofer is a good guy but we trigger their reactions and get permanently blocked from their sessions?, there's nothing we can lose from adding this logic and it helps a lot...

Dude, we'll end up in their database anyway because we'll spoof the token too. If we were able to find out that the host is a cheater, we'll add him to the database and won't connect to his session anymore. Even if we don't know this host but he blocked our connection during matchmaking, matchmaking will connect us to the next session in the queue without a banner.

snowleopardd:
> > I'm just suggesting adding a check to prevent automatically kicking host token spoofers who are already host, the less we are in their databases, the better, no?, also what if the spoofer is a good guy but we trigger their reactions and get permanently blocked from their sessions?, there's nothing we can lose from adding this logic and it helps a lot...
> 
> Dude, we'll end up in their database anyway because we'll spoof the token too. If we were able to find out that the host is a cheater, we'll add him to the database and won't connect to his session anymore. Even if we don't know this host but he blocked our connection during matchmaking, matchmaking will connect us to the next session in the queue without a banner.

I was explicitly referring to getting permanently blocked by them or getting kicked which is stupid because we can prevent that easily (if spoofer and host = don't try kick), i see that you don't care about getting kicked and blocked and just move on to a next session, but I'd prefer not, we have different views it's your PR, if you don't want to add it its okay 👍

ghost:
> I was explicitly referring to getting permanently blocked by them or getting kicked which is stupid because we can prevent that easily (if spoofer and host = don't try kick), i see that you don't care about getting kicked and blocked and just move on to a next session, but I'd prefer not, we have different views it's your PR, if you don't want to add it its okay 👍

It's definitely better to be kicked immediately and join the next session than to play in a session with a host modder and be kicked for anything at any time (friendly modders is a good joke), especially since we can't determine whether the host is a modder or not, maybe he didn't even spoof the token and filled the solo session earlier with a multiplexer.

Yimura:
Personally @mikedrugs and @CringeArab are making fairly good arguments. I think there's still room for improvement, perhaps we can offer additional options to users to either always try to kick the host regardless of what kind of protections and spoofing they may have applied or alternatively try not to kick the host if they're token spoofing (they likely are running a mod menu and in superior position making it unlikely we can kick them).

CypherMorgan:
> Personally @mikedrugs and @CringeArab are making fairly good arguments. I think there's still room for improvement, perhaps we can offer additional options to users to either always try to kick the host regardless of what kind of protections and spoofing they may have applied or alternatively try not to kick the host if they're token spoofing (they likely are running a mod menu and in superior position making it unlikely we can kick them).

An option for a user to select if they want to kick the host if they're a modder can be added imo. This would satisfy the needs of both type of users, who always want to try to kick the host no matter if they're modder or not and the ones who just wanna kick non-modders.

Edit: It just got added the same minute I commented this, my bad.

snowleopardd (commented):
Nicee, gj

gir489returns (approved):
image

Looks good to me.

Yimura (approved):

PR #3473, opened 2024-07-30
Refactor how blips work.

### LINKED ISSUE #3465: [Bug]: Autodrive to Objective
### Describe the bug

After pressing To Objective, when trying to autopilot, the vehicle stops every 15-20 seconds and the message "no waypoint found" appears. This happens during business battles or the delivery of all kinds of goods, etc.

### Steps To Reproduce

1. Start a business battle
2. Pick up the goods
3. Press Autopilot/To Objective
4. Wait for some time, 10-20 seconds
5. Vehicle stops
6. Sadness

### Expected Behavior

We must reach the yellow mark.

### Logs

Logs?

### Screenshots / Video

image

### Store

Steam

### Language

English

### YimMenu Language

English (US)

### Additional context

_No response_

### Build

### Coload

### LINKED ISSUE #3469: [Request]: Set time of the Day/Override time with non static clock.
### Problem

The menu does override time but the clock freezes. 

### Solution

If there was a way that we can just rewind / forward time locally (like single player) and after that the clock continues to moves forward in Gta time (30X faster). Also can be implemented the option to slow down time or accelerate time.

### Reason

Helps in role playing. for example start day and go to work at 6am or 9am and then go back to apartment at 5pm etc. 

### Additional context

Got this idea as it was the feature of Wasabi Unlocker menu by @ShinyWasabi. If he can add the same code into the menu or added Lua Script. The time dilation is the idea Model railroad designer use all the time.

gir489returns:
(Potentially) Closes #3465
Depends on https://github.com/Yimura/GTAV-Classes/pull/183


Yimura (approved):
Code looks good, didn't test.

gir489returns:
@Yimura I tested this in SP, and it caught a lot of mission blips that weren't caught before, like during the mission with Beverly, it teleported me to the escape route, despite it being in red and not mission yellow. It also teleported me to one of the assassination blips, despite it being in red too.

Yimura (approved):

PR #3479, opened 2024-07-30
Add Return Value Support for Script Functions

### LINKED ISSUE #3417: [Bug]: Calling Script Functions with Reference Parameters Causes Exception
### Describe the bug

When calling a script function that includes a parameter passed by reference, an EXCEPTION_ACCESS_VIOLATION occurs.

Example:
```c
void func_1460(int iParam0, int iParam1, var uParam2, bool bParam3, bool bParam4, bool bParam5)
```
```cpp
uint64_t transaction_hash = "SERVICE_EARN_DAILY_STASH_HOUSE_COMPLETED"_J;
uint64_t transaction_price = NETSHOPPING::NET_GAMESERVER_GET_PRICE(transaction_hash, "CATEGORY_SERVICE_WITH_THRESHOLD"_J, true);
uint64_t service_index = 0;

scr_functions::fire_service_transaction({ transaction_hash, transaction_price, service_index, true, false, false });
```
In this scenario, service_index (uParam2) must be passed by reference, but the current implementation does not support this, leading to an exception.

### Steps To Reproduce

1. Call a script function with a parameter that should be passed by reference.

### Expected Behavior

The function should execute without causing an EXCEPTION_ACCESS_VIOLATION. The value should be correctly passed to the reference parameter, and any modifications made to the reference parameter within the function should be returned correctly.

### Logs

Log for Lua: [cout.txt](https://github.com/user-attachments/files/16357985/cout.txt)
Log for YimMenu: [cout.log](https://github.com/user-attachments/files/16358208/cout.log)

### Store

Steam

### Language

English

### YimMenu Language

English (US)

### Additional context

_No response_

### Build

### Coload

### LINKED ISSUE #3497: [Bug]: Self -> Graceful Landing not persistent
### Describe the bug

state of Graceful Landing isnt saved

### Steps To Reproduce

tick the checkbox, reinject yimmenu and see that Graceful Landing is set to false

### Expected Behavior

save the state of Graceful Landing

### Logs

no

### Screenshots / Video

image

### Store

Rockstar

### Language

English

### YimMenu Language

English (US)

### Additional context

_No response_

### Build

### Coload

ShinyWasabi:
- Added return value support for script functions and exposed it to Lua.
- Removed Dance Mode & Los Santos Customs.
- Replaced add_clan_logo_to_vehicle with a script function.
- Updated creator stack size for start_creator_script.
- Fixed reset_session_data not being called due to pausemenu_multiplayer being terminated immediately after the we request it. The function is now called from main_persistent.
- Updated load_from_datafile pattern.
- Added get_int & set_int methods to the Lua memory table.

Closes https://github.com/YimMenu/YimMenu/issues/3417
Closes #3497


src/backend/backend.cpp
```diff
@@ -54,7 +54,6 @@ namespace big
 		{
 			looped::self_wanted();
 			looped::self_hud();
-			looped::self_dance_mode();
```
maybegreat48, line 57: Why was this removed?

gir489returns: From Shiny's notes:

> I found two more bugs:
> - dance mode doesn't work, it even breaks your nightclub, you can't enter the property
> - pattern of load_from_datafile is outdated

Mr-X-GTA: It works for me, but the `dance_state` global seems to be incorrect (should be `1943513`?)

ShinyWasabi: The main reason it was removed isn’t solely because it wasn’t working, but because these features can now be converted into a Lua script, as we have exposed script patches and functions. The same applies to LSC. They don't need to be in the base YimMenu.

lonelybud:
clearly lot of stuff has changed in this pr. Please add proper description

src/backend/backend.cpp
sch-lda: This feature is used very infrequently and convert it to Lua should have little impact. Please try not to convert frequently used features that are already located within YimMenu to Lua, not all users follow the PR and Lua repositories and they don't know where to find these features. These features are already localized in other languages, and converting them to Lua will make multilingual support more difficult.

Tayipc38: Can't we add a little window to download these Lua's via the menu itself?

Yimura: I agree with both, however some functionality is quite shit and should be considered for removal whereas some of the others can be reworked a bit and kept in.

Also at some point we should look into making a sort in-menu Lua workshop to install scripts with.

Tayipc38: > Also at some point we should look into making a sort in-menu Lua workshop to install scripts with.

I can do that right now

gir489returns: > > Also at some point we should look into making a sort in-menu Lua workshop to install scripts with.
> 
> I can do that right now

That would be amazing if you could get that implemented.

gir489returns:
@maybegreat48 I added a script function called vehicle_cannot_accept_clan_logo, however in the script they check if NETWORK_IS_GAME_IN_PROGRESS is true and just return TRUE. Is there a way to make the script function caller jump to an arbitrary location? Ideally, we'd just need it to run the switch statement and nothing else from the function.

ShinyWasabi:
I added a script patch to replace the return with a NOP. It shouldn't cause any issues in SP as this function is only used in online. That's the reason of the `if (!NETWORK::NETWORK_IS_GAME_IN_PROGRESS()) return true;` check. It blocks applying clan logo to vehicles in SP.

src/script_function.hpp
```diff
@@ -1,43 +1,71 @@
...
+			tls_ctx->m_is_script_thread_active = og_thread != nullptr;
+
+			if constexpr (!std::is_same_v<Ret, void>)
+			{
+				return *reinterpret_cast<Ret*>(stack + top_stack);
+			}
+		}
+	};
+
+	static inline script_function set_as_ceo("SAC", "freemode"_J, "2D 04 1D 00 00 5D");
```
xiaoxiao921: static inline for global variables in hpp file mean that for every include the ctor is fired. just make them inline.

xiaoxiao921 (dismissed):

src/util/scr_functions.hpp
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+#include "script_function.hpp"
+
+namespace big::scr_functions
```
maybegreat48: Can't we just call the script functions directly? And I think we should store them in `scr_functions` instead of leaving it at `big`

gir489returns: When I called them outside of a fiber, I got issues.

Yimura (dismissed):

gir489returns:
image

I was able to clone another player's car perfectly including their clan logo with this, so I'd say this is ready for merging.

Issue #3273, opened 2024-06-20
[Request]: PTFX spam protection


tiCkuhAr:
### Solution

Prevent people from being able to spam PTFX's on you.

### Reason

Because it's annoying and lowers your frame rate, it could even be considered a crash for people with lower end PCs.

lonelybud:
yeah something like this simple can be added - 

```C++
rate_limiter m_ptfx_ratelimit{1s, 5};
....

		case eNetworkEvents::NETWORK_PTFX_EVENT:
		{
			if (g_debug.log_ptfx_event)
				LOGF(WARNING,
				    "PTFX Event: {} (Dist- {})",
				    plyr->m_name,
				    math::distance_between_vectors(*plyr->get_ped()->get_position(), *g_local_player->get_position()));

			if (plyr->block_ptfx)
				return send_ack_event();

			if (plyr->whitelist_ptfx)
				break;

			if (plyr->m_ptfx_ratelimit.process())
			{
				if (plyr->m_ptfx_ratelimit.exceeded_last_process())
					g_reactions.ptfx_spam.process(plyr);

				return send_ack_event();
			}

			buffer->Seek(0);
			break;
		}
```

tiCkuhAr:
I saw there was a pull request to add it, but it never got merged. 
https://github.com/YimMenu/YimMenu/pull/2873

GreySideMM:
That PR didn't get merged because eNetworkEvents was outdated at this point in time. This enum was updated some short time ago by Mr-X-GTA so this should be no longer an problem.

Issue #1373, opened 2023-05-26
[Request]: Detect modder explosions


manav-souluda01:
### Problem

explosion griefing is the most common forum of griefing in GTA. i have seen stand now that can detect any sort of modder explosion. be it explosive ammo, you spawning an explosion on them, etc. im not a programmer, i don't know what the method is, but its possible.

### Solution

ideally yim menu could detect:
- explosions spawned on us by modders
- blamed explosions (maybe read the kill feed and compare the person who is sending explosions?)
- explosive ammos

### Reason

like above, explosion griefing is toxic and common. we should be able to easily remove these toxic people, but it's not always the most obvious procedure with explosions. it would make life easier if yim could detect these automatically. it'd also give yim an upper hand in detections, since explosive ammo is super common and detecting that is kinda op.

Yimura:
We are detecting blamed explosions by default however, the data fields in the event have never been completely reverse engineered before.

Because of this it's really hard to properly detect if an explosion is actually a blamed explosion, we are detecting blamed explosions but not always correctly so when you see the event pop-up be critical and check to see if it was a modder blame exploding or a false positive.

Due to the nature of how GTA handles creating explosions it's really hard to figure out. Say for example you put a sticky bomb down at the casino and drive away all the way to the main airport. When you want to explode it you're not actually the one creating the explosion it's one of the players close to the explosive. This will set them as the ones setting off the explosive but blamed on you, as you can see this is not a modder blamed explosion but normal behaviour.

PR #3480, opened 2024-07-31
feat(view): add Network Syncs tab

### LINKED ISSUE #3369: [Request]: Explosion spam protection
### Problem

Modders can spam explosions on you.

### Solution

Rate limit explosions and notify the user of the person spamming them.

### Reason

It's a bit annoying.

### Additional context

This should probably only be done in freemode, as there are some missions like Cayo that normally spam explosions on you.

lonelybud:
refractor code
- changes in VIEW_PLAYER_INFO_EXTRA_INFO
  - add block_ptfx option
  - display full vehicle name
- changes in gta_data_service
  - change key type in vehicle_map & ped_map
  - add get_vehicle_full_name
 
added GUI_NETWORK_SYNCS tab to track ped, vehicle, pickups, objects, explosion, ptfx events

closes #3369, #1373, #3273


src/hooks/protections/received_event.cpp
```diff
@@ -867,7 +881,15 @@ namespace big
 				return;
 			}
 
-			scan_explosion_event(source_player, buffer);
+			scan_explosion_event(plyr, source_player, buffer);
+
+			// dont accept event in died state
```
Yimura: Don't let the game handle the event if we're dead you mean? We're still accepting the event (from the attackers POV).

src/hooks/protections/received_event.cpp
```diff
@@ -386,6 +387,19 @@ namespace big
 			return;
 		}
 
+		// logs all type of explosion including owned one, fire, water hydrant etc, one without damage but camerashake, npcs shooting explosive ammo from planes from source client etc
+		if (g.debug.logs.explosion_event && plyr)
+			{
```
Yimura: Mixed indents

Yimura (changes requested):

src/hooks/protections/received_event.cpp
lonelybud: if we are dead the explosion event is useless to us, it will only create explosions around the dead camera scene if intended for us. Also i have added the check later so that the logging can still happen. I am open to your ideas.

lonelybud: hmn i think it will create problem since it is not specific. Some explosion like car being exploded etc which are legit will also get rejected. This is concerning you are right

Yimura: I was more targeting the wording used in the comment, we're really still accepting the event just not letting the game process it and create the explosion on our end.

lonelybud: my main objective was to remove frame drop due to explosion ptfx effect when we are in died state and someone is spamming explosions.

lonelybud: right we are just sending the acknowledgement but in game we are not processing the explosion. I have not tested this with scenarios like when you die and someone else die in meantime bcz of explosion or someone's car get exploded. Therefore i removed it.

lonelybud:
When someone is spamming explosions then in the console the event is logged very fast so checking and then unchecking the toggle becomes necessary

lonelybud:
i was able to track the client (player) who spawned the planes which fired explosive ammos on other players. The question is how was I able to is it because their client was the owner and responsible for sending explosion events?

prince4589:
> i was able to track the client (player) who spawned the planes which fired explosive ammos on other players. The question is how was I able to is it because their client was the owner and responsible for sending explosion events?

is it possible ?? to implement this feature !!
cuz recently i was kicking all god mod abuser as i was host but then someone dropped god mod jets on everyone in lobby and rc tanks resulting in everyone to leave :(

i once tried to spawn jets on a paid modder and he detected that so i thought if thats possible with yim or lua script

lonelybud:
> is it possible ?? to implement this feature !! cuz recently i was kicking all god mod abuser as i was host but then someone dropped god mod jets on everyone in lobby and rc tanks resulting in everyone to leave :(
> 
> i once tried to spawn jets on a paid modder and he detected that so i thought if thats possible with yim or lua script

I have opened this pr so that one can easily log or get notification related to explosions. And yes if you understand the nature and cause of explosions very well then you can detect that player without false positives. I believe explosions are certain and one can get out with lot of clues who caused it. I didn't got time to test all scenarios thats why it is in debug section in this pr and can be used to detect explosions on your own risk with probability of false positives. I was able to detect the player who spawned jet because he was still in control of his jets and explosions happening in his client was emitted to everyone. I got explosion events only from that client and hence kicked him. I was right and after he left those jets stopped firing and left to flew in air.

lonelybud:
may be can add a UI tab - `High Level Detections`.  We will store all explosion, ped, vehicle, pickups spawn events. For example 50 events max of each type in a `unordered_map`. Then if they get full they will push their data to backups of same size. This way we have last 100 events stored for each type. We can use this data and render in `ImGui::BeginChild("ScrollingRegion"` for each type inside this new tab  `High Level Detections` and this way we can identify who is troublemaker. This is the easiest solution.

PR #3484, opened 2024-07-31
Add more lsc colors

V1RU513N:


gir489returns (dismissed):

src/core/data/lsc_types.hpp
```diff
@@ -48,7 +48,7 @@ static const std::map<int, std::string> lsc_classic_colors = {
 	{COLOR_CLASSIC_BLACK, "Black"},
 	{COLOR_CLASSIC_CARBON_BLACK, "Carbon Black"},
 	{COLOR_CLASSIC_GRAPHITE, "Graphite"},
-	{COLOR_CLASSIC_ANHRACITE_GREY, "Anhracite Grey"},
+	{COLOR_CLASSIC_ANTHRACITE_BLACK, "Anhracite Black"},
```
gir489returns: According to the GTA V wiki, this is supposed to be Anthracite Black, yes?

gir489returns (changes requested):
See: https://gta.fandom.com/wiki/Vehicle_Customization_in_GTA_V/Respray_Colors#Classic

src/core/data/lsc_types.hpp
V1RU513N: Yes, the "color name" is “anthracite black” Anthracite grey would be the “technical name”, hence my confusion. I've chosen to stick to the “color name”.

https://docs.google.com/spreadsheets/d/e/2PACX-1vTa7yvIzGZRjK2cWS-jtkpD649KjV6vBR3cLjWQQ0JxCcLIotNGlxBMlqs_Z7c7lk8OQVUkGwgwdKqa/pubhtml

gir489returns: If you want to see how the game gets these values, you can use something like HUD::GET_FILENAME_FOR_AUDIO_CONVERSATION with the internal GXT strings to resolve it at runtime. This has the added benefit of using the game's translations for the colors.

If you need the GXT strings, carmod_shop should have them. I found a few using your spread sheet. func_1496 is Silver through Lime Green. func_1498 is Black through Frost White. func_1507 might be some kind of special color? Maybe pearlescent. It uses a couple of extra parameters, probably for the sheen percentage. It has Brushed Black Steel through Brushed Gold.

gir489returns (approved):

Yimura (approved):

PR #3527, opened 2024-08-06
Better enemy detection for aimbot

gir489returns:


gir489returns:
I've noticed that there are still scenarios where enemies are spawned and shooting at the player, but because they aren't blips on the map, they're considered not enemies from this code, but still turn the crosshair red, so the engine must have some kind of generic enemy determination system...

Yimura (dismissed):

xiaoxiao921:
Looks good to me, just small thing, should the is_ped_a_friend function be inside the blip util file? Maybe ped util is better

gir489returns:
> Looks good to me, just small thing, should the is_ped_a_friend function be inside the blip util file? Maybe ped util is better

I put it in there, since a majority of the code is about utilizing the ped's blip. I'll experiment more by trying to factor that code out.

xiaoxiao921 (dismissed):

Yimura (approved):

PR #3550, opened 2024-08-09
Fixed Weapons JSON not properly using the most recent meta file.

gir489returns:
I noticed an issue with the weapons.json file was using the non-updated WEAPON_MG components, and not the one from update.rpf, where I guess Rockstar had changed the components from COMPONENT_AT_SCOPE_SMALL to COMPONENT_AT_SCOPE_SMALL_02, which would cause Ammunation and Persist Weapons to not work properly.

### DIFF
--- a/src/services/gta_data/gta_data_service.cpp
+++ b/src/services/gta_data/gta_data_service.cpp
@@ -330,9 +330,25 @@ namespace big
 						std::string LocName = item.child("LocName").text().as_string();
 						std::string LocDesc = item.child("LocDesc").text().as_string();
 
-						if (LocName.ends_with("INVALID") || LocName.ends_with("RAIL"))
-						{
+						if (LocName.ends_with("RAIL"))
 							continue;
+
+						if (LocName.ends_with("INVALID"))
+						{
+							while (!SCRIPT::HAS_SCRIPT_LOADED("achievement_controller"))
+								script::get_current()->yield();
+
+							Hash weapon_hash = 0;
+							if (name.starts_with("COMPONENT_KNIFE"))
+								weapon_hash = "WEAPON_KNIFE"_J;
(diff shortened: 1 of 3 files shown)


src/services/gta_data/gta_data_service.cpp
```diff
@@ -329,9 +332,48 @@ namespace big
...
+							if (!SCRIPT::GET_NUMBER_OF_THREADS_RUNNING_THE_SCRIPT_WITH_THIS_HASH(script_hash))
+							{
+								while (!SCRIPT::HAS_SCRIPT_WITH_NAME_HASH_LOADED(script_hash))
+								{
+									SCRIPT::REQUEST_SCRIPT_WITH_NAME_HASH(script_hash);
+									script::get_current()->yield(10ms);
+								}
+								thread_id = SYSTEM::START_NEW_SCRIPT_WITH_NAME_HASH(script_hash, 1424);
+								SCRIPT::SET_SCRIPT_WITH_NAME_HASH_AS_NO_LONGER_NEEDED(script_hash);
+								while (!SCRIPT::GET_NUMBER_OF_THREADS_RUNNING_THE_SCRIPT_WITH_THIS_HASH(script_hash))
```
maybegreat48: You don't need to wait for this since the thread will start running the moment you call `START_NEW_SCRIPT_WITH_NAME_HASH`

maybegreat48: You should also attempt to pause the thread to make sure it doesn't end up killing itself (see the LSC code for more details)

gir489returns: We no longer have that function anymore as we couldn't fix it to work properly without freezing the transaction stack, so we took it out, so I had to go find it in the git history.

Yimura (approved):

PR #3559, opened 2024-08-12
Fixed some weapons whose update files did not follow FIFO.

gir489returns:


gir489returns:
@xiaoxiao921 Changing the return value to sol from vector to set should still just return an (ordered) table to the front-end, right? I don't think anything functional should change on the Lua user's side.

Yimura (approved):

xiaoxiao921 (approved):

PR #3610, opened 2024-08-22
Added protection against remote ped flag setting

41Baloo:
This has been made public now. This weapon damage event is unused by the game and will set a flag on your playerped that causes you to die infinitely. Enabeling godmode in that state won't stop you from dying and if you switch sessions you will be stuck due to the your ped dying before you can load in. You can stop this by just unsetting the flag but since this damage event is completely unused, you can just block it like this

### DIFF
--- a/src/hooks/protections/received_event.cpp
+++ b/src/hooks/protections/received_event.cpp
@@ -27,7 +27,6 @@ namespace big
 
 	static bool is_valid_weapon(rage::joaat_t hash)
 	{
-
 		for (const auto& info : g_pointers->m_gta.m_weapon_info_manager->m_item_infos)
 		{
 			if (info && info->m_name == hash && info->GetClassId() == "cweaponinfo"_J)
--- a/src/hooks/protections/received_event.cpp
+++ b/src/hooks/protections/received_event.cpp
@@ -27,6 +27,11 @@ namespace big
 
 	static bool is_valid_weapon(rage::joaat_t hash)
 	{
+		if ("WEAPON_TRANQUILIZER"_J == weaponType)
+		{
+			return true;	
+		}
+
 		for (const auto& info : g_pointers->m_gta.m_weapon_info_manager->m_item_infos)
 		{
 			if (info && info->m_name == hash && info->GetClassId() == "cweaponinfo"_J)
--- a/src/hooks/protections/received_event.cpp
+++ b/src/hooks/protections/received_event.cpp
@@ -27,10 +27,6 @@ namespace big
 
 	static bool is_valid_weapon(rage::joaat_t hash)
 	{
-		if ("WEAPON_TRANQUILIZER"_J == hash)
-		{
-			return false;	
-		}
 
 		for (const auto& info : g_pointers->m_gta.m_weapon_info_manager->m_item_infos)
 		{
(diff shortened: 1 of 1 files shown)


41Baloo:
Sorry, typed this on phone and i'm not exactly familiar with YimMenus structure. Feel free to implement this in whatever way you see more fit

Prisuhm:
> Sorry, typed this on phone and i'm not exactly familiar with YimMenus structure. Feel free to implement this in whatever way you see more fit

noob!

gir489returns:
Is WEAPON_TRANQUILIZER even a valid weapon? That seems like nonsense they've made to dispatch this event. We should be looking for the damage type they are using instead, which is probably out of bounds or something like the Tazer + an infinite duration.

Mr-X-GTA:
> We should be looking for the damage type they are using instead, which is probably out of bounds or something like the Tazer + an infinite duration.

They check for the weapon hash and set the flag here:
`75 2F 8B 83 ?? ?? ?? ?? F7 D0`

It kills the ped every 1000ms because the flag is never unset:
`48 89 5C 24 08 55 56 57 41 54 41 55 41 56 41 57 48 8D 6C 24 D9 48 81 EC C0 00 00 00 8B 82`

src/hooks/protections/received_event.cpp
```diff
@@ -113,7 +116,6 @@ namespace big
 			    "Blocked WEAPON_DAMAGE_EVENT from {} with invalid weapon hash {:X}",
 			    player->get_name(),
 			    weaponType);
-			g_pointers->m_gta.m_send_event_ack(event_manager, player, target_player, event_index, event_handled_bitset);
```
gir489returns, line 111: Re-add this. Otherwise their client will keep spamming us with requests.

gir489returns (changes requested):

src/hooks/protections/received_event.cpp
```diff
@@ -100,6 +100,13 @@ namespace big
 		damageType = buffer->Read<uint8_t>(2);
 		weaponType = buffer->Read<uint32_t>(32);
 
+		if ("WEAPON_TRANQUILIZER"_J == weaponType)
+		{
+			if (auto plyr = g_player_service->get_by_id(player->m_player_id))
+    			g.reactions.break_game.process(plyr);
```
gir489returns, line 106: Illformed code block.

Should be:

```
if (auto plyr = g_player_service->get_by_id(player->m_player_id))
{
    g.reactions.break_game.process(plyr);
}
```

gir489returns (changes requested):

src/hooks/protections/received_event.cpp
Mr-X-GTA: https://github.com/YimMenu/YimMenu/blob/4046640c1e2e9f7fdd6d51895020889d98ca30b6/src/hooks/protections/received_event.cpp#L873-L881

gir489returns: Ah, yes. My mistake.

BigBootyAlcholic:
you can just unset the flag in `CPed + 0x147C` lol

src/hooks/protections/received_event.cpp
41Baloo: i was under the impression

```
if (auto plyr = g_player_service->get_by_id(player->m_player_id))
    			g.reactions.break_game.process(plyr);
```
    			
was how 1 lined if statements were supposed to be made. Changed it to yours now tho

gir489returns: While it is valid syntax, and it has been seen around YimMenu before, it's not considered proper form. We're trying to move away from it.

gir489returns (approved):

PR #3619, opened 2024-08-24
Fix Thread State for Script Functions and cellphone_flashhand

ShinyWasabi:


tupoy-ya (approved):
I don't think `idle` and `running` were incorrect before  https://github.com/Yimura/GTAV-Classes/pull/186

xiaoxiao921:
Running is definitly 0 in the enum.
The two changes for the script_vm stuff make sense, you want the thread to be running for running the function inside the script vm, at first glance the cellphone change is wrong, but if it has been tested then lgtm

ShinyWasabi:
> at first glance the cellphone change is wrong, but if it has been tested then lgtm

We were previously setting it to 1, which was `running` in the old enum. And the code used to work just fine. We now set it to `idle`, which is 1 in the new enum.

xiaoxiao921 (approved):

gir489returns (approved):

gir489returns:
And this is why I wanted MBG to look it over, but you know, raw testing works too.

Yimura (approved):

PR #3627, opened 2024-08-26
Move Collapsing Headers in Weapons to Separate Tab

ShinyWasabi:
- Moved ammu-nation, persist weapons, and weapon hotkeys to a separate tab named "Ammu-Nation".
- Added a command to reset the player's weapon loadout to default.
- Aligned aimbot checkboxes.


src/core/data/weapon_tints.hpp
```diff
@@ -0,0 +1,50 @@
...
+		"Metallic Gold",
+		"Metallic Platinum",
+		"Metallic Grey & Lilac",
+		"Metallic Purple & Lime",
+		"Metallic Red",
+		"Metallic Green",
+		"Metallic Blue",
+		"Metallic White & Aqua",
+		"Metallic Red & Yellow"
+	});
```
gir489returns, line 49: Now that we have access to the script functions, can't we grab these at runtime so they're translated?

src/backend/commands/self/reset_weapons.cpp
```diff
@@ -0,0 +1,21 @@
...
+	{
+		using command::command;
+
+		virtual void execute(const command_arguments&, const std::shared_ptr<command_context> ctx)
+		{
+			if (*g_pointers->m_gta.m_is_session_started && scripts::is_running("freemode"_J))
+			{
+				*scr_globals::should_reset_fm_weapons.as<PBOOL>() = TRUE;
+				scr_functions::give_owned_freemode_weapons.call<void>();
+			}
```
gir489returns, line 16: Could we tie this into Persist Weapons somehow, that if it detects a weapons malignment, it runs this instead? Make that a user configurable thing. We also want to check for a runaway scenario, where it's stuck in a loop. (It keeps detecting weapons have changed, but running the scr func has no outcome, thus either stop what it's doing or warn the user)

gir489returns (changes requested):

src/core/data/weapon_tints.hpp
ShinyWasabi: Unfortunately, there isn't a suitable function we can use for this. Instead, I've modified the code to generate the translations using `GET_FILENAME_FOR_AUDIO_CONVERSATION` when the user opens the tab.

gir489returns (approved):

PR #3628, opened 2024-08-26
Disable script patches from the Lua layer on eject.

gir489returns:
User testing of [DailyCollectables](https://github.com/YimMenu-Lua/DailyCollectibles) showed that when ejecting a script that contained script patches, it would not properly clean them up, so if the user would then re-enable the script, it would try to patch a new area that the signatures could land on. We might still have a similar issue in the main cheat, but it's not a huge issue since only devs can eject the main module at runtime.


gir489returns:
https://github.com/YimMenu-Lua/DailyCollectibles/issues/28

The PR has been confirmed to fix the issue.

xiaoxiao921 (approved):

PR #3630, opened 2024-08-27
refactor(Lua)!: Use LuaJIT.

tupoy-ya:
## EXPERIMENTAL
This PR is not intended to be merged any time soon. And *will* beak existing scripts.

### Reasoning
The main reason I made this is because LuaJIT has an amazing [FFI Library](https://luajit.org/ext_ffi.html) that allows you to call C functions and read C structs much easier then YimMenu's `memory` library.
It can also load libraries like `libcurl.dll` that GTAV comes with.
While this is all obviously insecure, we already have the `memory` library which can already do most if not all of this.
It would probably also be possible to implement dynamic hooking with this, if we `dllexport` the MinHook API.

<details>
  <summary>FFI Example</summary>

```lua
ffi = require("ffi")
ffi.cdef[[
int MessageBoxA(void *w, const char *txt, const char *cap, int type);
typedef union
{
	float data[4][4];
	struct
	{
		struct
		{
			float x, y, z, w;
		} rows[4];
	};
} fmatrix44;

typedef struct
{
	void* m_pad1;
	void* m_pad2;
	void* m_pad3;
	void* m_pad4;
	uint64_t* m_model_info; //0x0020
	uint8_t m_entity_type; //0x0028
	char gap29; //0x0029
	uint16_t gap2A; //0x002A
	uint32_t m_flags; //0x002D
	uint64_t* m_navigation; //0x0030
	uint16_t gap38; //0x0038
	uint16_t gap3A; //0x003A
	uint32_t gap3C; //0x003C
	uint64_t* m_dynamic_entity_component; //0x0040 (stores attachments and stuff)
	uint64_t* m_draw_data; //0x0048
	uint64_t* gap50; //0x0050
	uint64_t gap58; //0x0058
	fmatrix44 m_transformation_matrix; //0x0060
	void* m_render_focus_entity; //0x00A0
	uint32_t m_render_focus_distance; //0x00A8
	uint32_t m_flags_2; //0x00AC
	uint32_t m_shadow_flags; //0x00B0
	char gapB4[4]; //0x00B4
	uint8_t byteB8; //0x00B8
} fwEntity;
]]

local lua_tab = gui.get_tab("GUI_TAB_LUA_SCRIPTS")
lua_tab:add_imgui(function()
	local player_ped_ptr = memory.handle_to_ptr(PLAYER.PLAYER_PED_ID())
	if ImGui.Button("FFI test") then
		local player_ped = ffi.cast("fwEntity*", player_ped_ptr:get_address())
		local player_pos = player_ped.m_transformation_matrix.rows[3]
		local message = string.format("Our entity type is: %i\nOur pos is: (%f, %f, %f", player_ped.m_entity_type, player_pos.x, player_pos.y, player_pos.z)

		ffi.C.MessageBoxA(nil, message, "Ped info", 0)
	end
end)
```
</details>

### Known incompatibilities with existing scripts:
- Bitwise operators `&` `|` `<<` `>>` are not supported, you need to use the `bit` library.
- `entities.get_all_*_as_handles` returns a usertype which cannot be iterated over using `ipairs`, you need to iterate manually: 
```lua
local all_vehicles = entities.get_all_vehicles_as_handles()
for i = 1, #all_vehicles do
	log.info(all_vehicles[i])
end
```
- `//` operator is not supported
- `<const>` and `<close>` local variables are not supported.


xiaoxiao921:
I could see this getting merged, considering we already have dynamic_call and dynamic_hook implemented right now

edit: nevermind, it would lower the barrier by a lot, not sure then

PR #3634, opened 2024-08-29
Input Method Editor

### LINKED ISSUE #3613: [Improvement]: The input box cannot input Chinese, but can paste
### Which feature

ImGui:input box

### Reason

The input box cannot input Chinese characters. If you want to input Chinese characters, you can only copy them from other places and then paste them into the input box.

### Additional context

_No response_

Mr-X-GTA:
Closes https://github.com/YimMenu/YimMenu/issues/3613

### DIFF
--- a/src/pointers.cpp
+++ b/src/pointers.cpp
@@ -1958,6 +1958,16 @@ namespace big
             {
                 g_pointers->m_gta.m_can_do_damage_to_ped = ptr.add(1).rip().as<functions::can_do_damage_to_ped>();
             }
+        },
+        // Input Method Editor
+        {
+            "IME",
+            "75 25 89 44 24 28",
+            [](memory::handle ptr)
+            {
+                g_pointers->m_gta.m_allow_keyboard_layout_change = ptr.sub(4).rip().as<bool*>();
+                g_pointers->m_gta.m_ime = ptr.add(44).rip().sub(0x278 + 0x8).as<InputMethodEditor*>();
+            }
         }
         >(); // don't leave a trailing comma at the end
 
--- a/src/util/input_method_editor.hpp
+++ b/src/util/input_method_editor.hpp
@@ -3,40 +3,27 @@
 #include "pointers.hpp"
 
 #include <imgui_internal.h>
-
-#pragma pack(push, 8)
-class InputMethodEditor
-{
-public:
-	uint32_t m_count; //0x0000
-	uint32_t m_selected_index; //0x0004
-	wchar_t m_composition_string[31]; //0x0008
-	wchar_t m_candidate_list[9][31]; //0x0046
-	bool m_active; //0x0274
-	char pad_0275[3]; //0x0275
-}; //Size: 0x0278
-static_assert(sizeof(InputMethodEditor) == 0x278);
-#pragma pack(pop)
+#include <misc/InputMethodEditor.hpp>
 
 namespace
 {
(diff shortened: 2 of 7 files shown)


src/util/input_method_editor.hpp
```diff
@@ -0,0 +1,125 @@
...
+public:
+	uint32_t m_count; //0x0000
+	uint32_t m_selected_index; //0x0004
+	wchar_t m_composition_string[31]; //0x0008
+	wchar_t m_candidate_list[9][31]; //0x0046
+	bool m_active; //0x0274
+	char pad_0275[3]; //0x0275
+}; //Size: 0x0278
+static_assert(sizeof(InputMethodEditor) == 0x278);
+#pragma pack(pop)
```
gir489returns, line 19: Since this class is coming from GTA 5 itself, it belongs in the GTAV-Classes project.

src/util/input_method_editor.hpp
```diff
@@ -0,0 +1,125 @@
...
+			state->Stb.cursor += len;
+			state->Stb.has_preferred_x = 0;
+
+			state->CursorFollow = true;
+		}
+
+		ImmReleaseContext(g_pointers->m_hwnd, context);
+	}
+
+	inline void draw_input_method_editor()
```
gir489returns, line 82: Do these methods have to be inlined?

src/util/input_method_editor.hpp
```diff
@@ -0,0 +1,125 @@
...
+		obj->CurLenA += new_text_len_utf8;
+		obj->TextW[obj->CurLenW] = '\0';
+
+		return true;
+	}
+}
+
+namespace big
+{
+	inline void handle_ime_result()
```
gir489returns, line 59: Do these methods have to be inlined?

src/pointers.cpp
```diff
@@ -1958,6 +1958,16 @@ namespace big
...
             }
+        },
+        // Input Method Editor
+        {
+            "IME",
+            "75 25 89 44 24 28",
+            [](memory::handle ptr)
+            {
+                g_pointers->m_gta.m_allow_keyboard_layout_change = ptr.sub(4).rip().as<bool*>();
+                g_pointers->m_gta.m_ime = ptr.add(44).rip().sub(0x278 + 0x8).as<InputMethodEditor*>();
```
gir489returns, line 1969: This is extremely unlikely to be consistent from version to version. I think finding a new signature for InputMethodEditor* would be more appropriate.

gir489returns (changes requested):

src/pointers.cpp
Mr-X-GTA: I looked in 1.54 and it's the same there. 0x278 is just the size of the class of which there are two instances for some reason but we only need the first one

src/util/input_method_editor.hpp
Mr-X-GTA: Yes

gir489returns: They really don't as it's not doing anything in the function scope that has to be done once unrolled, and inline is a suggestion to the compiler, one it's probably ignoring. If it really did have to be inlined you'd be using __forceinline is bad.

Mr-X-GTA: Yeah I mean they don't have to be inlined but the inline keyword is required to avoid multiple definition errors

gir489returns (approved):
Looks good on the surface, untested.

Yimura (approved):

PR #3636, opened 2024-08-29
Updated Vehicle Kick via trigger_script_event with playermask

NGxDTV:
I've adjusted the code for the Vehicle Kick so that it triggers correctly, and the TargetPlayer is thrown out of the car instead of being directly outside. This way, the function is triggered properly.

My Test Version:

```
if (GUI::Option("Kick out of Vehicle", "")) //1.69
{
	int vehicle = PED::GET_VEHICLE_PED_IS_IN(PLAYER::GET_PLAYER_PED(Features::SelectedPlayer), false);

	int playerMask = 0;
	for (int i = 0; i < 32; ++i) {
		int playerIndex = PLAYER::INT_TO_PLAYERINDEX(i);
		if (PED::IS_PED_IN_ANY_VEHICLE(PLAYER::GET_PLAYER_PED(playerIndex), false)) {
			int currentVehicle = PED::GET_VEHICLE_PED_IS_IN(PLAYER::GET_PLAYER_PED(playerIndex), false);
			if (currentVehicle == vehicle) {
				MISC::SET_BIT(&playerMask, i);
			}
		}
	}

	const size_t arg_count = 10;
	uint64_t arguments_array[arg_count] = {
		static_cast<uint64_t>(eRemoteEvents::VehicleKick),
		PLAYER::PLAYER_ID(),
		0,
		0,
		0,
		0,
		-1,
		MISC::GET_FRAME_COUNT(),
		0,
		0
	};

	SCRIPT::TRIGGER_SCRIPT_EVENT(1, arguments_array, arg_count, playerMask, (int)eRemoteEvents::VehicleKick);
}
```


src/backend/commands/player/toxic/kick_from_vehicle.cpp
```diff
@@ -15,15 +15,36 @@ namespace big
 
 		virtual void execute(player_ptr player, const command_arguments& _args, const std::shared_ptr<command_context> ctx) override
 		{
-			auto vehicle = player->get_current_vehicle();
+			int vehicle = PED::GET_VEHICLE_PED_IS_IN(PLAYER::GET_PLAYER_PED(player->id()), false);
```
gir489returns, line 18: Pedantic, but the type returned from the natives is Vehicle not int. (We typedef it as int, because it takes up the same space in the C++ layer, but to the native VM operator it is different)

src/backend/commands/player/toxic/kick_from_vehicle.cpp
```diff
@@ -15,15 +15,36 @@ namespace big
...
 
-			if (!vehicle || !vehicle->m_net_object)
+			if (!vehicle)
```
gir489returns, line 20: This check was there to make sure people didn't execute this in SP. Why did you remove it? You also replaced the vehicle type from pointer to a native type. 0 does not indicate that the object was not found (-1 does), but now that you've changed the fundamental type, you now need to use ENTITY::DOES_ENTITY_EXIST, as the ! check was checking for nullptr.

gir489returns (changes requested):

src/backend/commands/player/toxic/kick_from_vehicle.cpp
NGxDTV: I removed it because there are no negative effects when used in single-player mode. Additionally, I didn't look at the vehicle->m_net_object since I'm working with a different source.

If it’s just the check, it can be put back in; I’m programming based on another source. In my version, I use the TSE Pointer, but I handle it through the native function:

```
static void TRIGGER_SCRIPT_EVENT(int event_group, uint64_t* args, int arg_count, int player_bits, int event_id) { 
	hooking::trigger_script_event((int)event_group, (uint64_t*)args, (int)arg_count, (int)player_bits, (int)event_id); 
}
```

Also, when triggering, if the person is not in a vehicle, there are no errors; the event just goes nowhere.

I only noticed when I briefly used YimMenu that people get thrown out of the vehicle instead of the vehicle slowing down and the person exiting on their own.

I hope I understood you correctly; my English isn't the best :)

src/backend/commands/player/toxic/kick_from_vehicle.cpp
NGxDTV: I tested the trigger for all functions and finally came up with this, here are the codes:

```
//Trigger Function:

void func_1846(int iParam0, int iParam1, float fParam2, int iParam3, int iParam4, int iParam5, bool bParam6)//Position - 0xB0D90
{
	struct<10> Var0;
	
	Var0.f_0 = -503325966;
	Var0.f_1 = PLAYER::PLAYER_ID();
	Var0.f_8 = bParam6;
	Var0.f_5 = iParam1;
	Var0.f_6 = iParam4;
	Var0.f_3 = fParam2;
	Var0.f_4 = iParam3;
	Var0.f_7 = iParam5;
	Var0.f_9 = MISC::GET_FRAME_COUNT();
	if (!iParam0 == 0)
	{
		SCRIPT::_SEND_TU_SCRIPT_EVENT_NEW(1, &Var0, 10, iParam0, Var0.f_0);
	}
}
```

playermask = (int)iParam0

```
//Trigger Example:

void func_2358(int iParam0)//Position - 0xD0E9B
{
	bool bVar0;
	int iVar1;
	
	iVar1 = 0;
	while (iVar1 < 32)
	{
		bVar0 = PLAYER::INT_TO_PLAYERINDEX(iVar1);
		if (func_21(bVar0, 0, 1) && bVar0 != PLAYER::PLAYER_ID())
		{
			if (PED::IS_PED_IN_VEHICLE(PLAYER::GET_PLAYER_PED(bVar0), iParam0, false) && !func_2359(bVar0))
			{
				func_1846(func_24279(bVar0), 0, 0f, 0, 0, 0, -1);
				VEHICLE::SET_VEHICLE_DOORS_LOCKED_FOR_PLAYER(iParam0, bVar0, true);
			}
		}
		iVar1++;
	}
}
```

```
//Other Functions:

int func_134(var uParam0, int iParam1, bool bParam2)//Position - 0x6EEC
{
	if (iParam1 == -1)
	{
		return 1;
	}
	func_135(uParam0, bParam2, 0);
	if (NETWORK::NETWORK_IS_GAME_IN_PROGRESS() && !bParam2)
	{
		if (MISC::ABSI(NETWORK::GET_TIME_DIFFERENCE(NETWORK::GET_NETWORK_TIME(), *uParam0)) >= iParam1)
		{
			return 1;
		}
	}
	else if (MISC::ABSI(NETWORK::GET_TIME_DIFFERENCE(MISC::GET_GAME_TIMER(), *uParam0)) >= iParam1)
	{
		return 1;
	}
	return 0;
}

int func_47(bool bParam0)//Position - 0x1F11
{
	int iVar0;
	
	iVar0 = bParam0;
	if (iVar0 < 0)
	{
		return 0;
	}
	if (iVar0 >= 32)
	{
		return 0;
	}
	return 1;
}

int func_24279(bool bParam0)//Position - 0x705203
{
	var uVar0;
	
	if (func_47(bParam0))
	{
		MISC::SET_BIT(&uVar0, bParam0);
	}
	return uVar0;
}

int func_21(bool bParam0, bool bParam1, bool bParam2)//Position - 0x15C1
{
	int iVar0;
	
	iVar0 = bParam0;
	if (iVar0 != -1)
	{
		if (NETWORK::NETWORK_IS_PLAYER_ACTIVE(bParam0))
		{
			if (bParam1)
			{
				if (!PLAYER::IS_PLAYER_PLAYING(bParam0))
				{
					return 0;
				}
			}
			if (bParam2)
			{
				if (iVar0 == Global_2672855.f_3)
				{
					return Global_2672855.f_2;
				}
				else if (Global_2657971[iVar0 /*465*/] != 4)
				{
					return 0;
				}
			}
			return 1;
		}
	}
	return 0;
}
```

PR #3644, opened 2024-08-31
add checkboxes for some missing protections

### LINKED ISSUE #3564: [Improvement]: Some protections are always active with no way of disabling them
### Which feature
I'd like to be able to disable all protections, except for the crash protections since sometimes, you want to be able to be teleported by other modders or pretend you are not a modder to avoid problems. Even with all protections disabled, other modders can't teleport me and I always get the notification despite having no way to disable it. The only thing I believe should be unchanged are the crash protections of course!

### Reason

In my opinion, it can help some people who want to just play and pretend they are a normal player. I know it's counterintuitive since the menu itself is for protections, but the option to disable the rest of the protections would be nice.

### Additional context

The notification I get the most is the "request control event". I don't know if that event is related to crashes aswell, hence why there isn't an option to disable it. If the devs find this "improvement" useless and counterintuitive, like I have stated before, just close the issue, no need for commenting!

In addition to this, I don't know if it's a bug or not, when sound protection is turned off, I still get "blocked sound spam" notifications when I think I shouldn't?

lonelybud:
- remote_wanted_level, clear_ped_tasks, ragdoll, request_control, vehicle_special_ability
- sound_spam check in NETWORK_PLAY_SOUND_EVENT

closes #3564

### DIFF
--- a/src/core/settings.hpp
+++ b/src/core/settings.hpp
@@ -292,7 +292,6 @@ namespace big
 				bool rotate_cam                 = true;
 				bool send_to_cutscene           = true;
 				bool send_to_location           = true;
-				bool sound_spam                 = true;
 				bool spectate                   = true;
 				bool give_collectible           = true;
 				bool vehicle_kick               = true;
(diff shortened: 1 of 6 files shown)


lonelybud:
I am also thinking to add description for all the checkboxes

lonelybud:
had a doubt related to `CVehicleProximityMigrationDataNode` whether to use both `g.protections.request_control` & `g.protections.clear_ped_tasks`

opt8:
would be great if all these checkboxes were under each other
image

lonelybud:
> would be great if all these checkboxes were under each other

can you clarify little bit more?

opt8:
> > would be great if all these checkboxes were under each other
> 
> can you clarify little bit more?

the signatures of the checkboxes do not fit in the current form (several items in one line). it would be better if each item starts with a new line

PR #3655, opened 2024-09-03
Enable All DLC Vehicles

ShinyWasabi:
Added script patches that allow all vehicles to be available for purchase even if their tunables are disabled.
This should work across different game versions (unless they change the logic). So even if new vehicles with tunables are added in future DLCs, they will automatically be enabled.


gir489returns:
This counts as recovery and should be relegated to a Lua script instead.

PR #3691, opened 2024-09-17
Update for b3323

Mr-X-GTA:
This makes YimMenu usable in SP.
- Disable BattlEye in the Rockstar Games Launcher.
- If the game crashes on injection, delete `%appdata%/YimMenu/settings.json`.

### DIFF
--- a/cmake/gtav-classes.cmake
+++ b/cmake/gtav-classes.cmake
@@ -2,8 +2,8 @@ include(FetchContent)
 
 FetchContent_Declare(
     gtav_classes
-    GIT_REPOSITORY https://github.com/Mr-X-GTA/GTAV-Classes-1.git
-    GIT_TAG        e78c6b2788263cce800971d124b6d1358eeb5cbc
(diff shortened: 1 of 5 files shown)


Yimura:
We should probably look into disabling online usage of YimMenu in GTA V Online, even though we know as developers that it's stupid to go online right now without a proper bypass we can't be sure that our user base has the proper knowledge and insights to realize this.

Mr-X-GTA:
You cannot go online if BattlEye is disabled and if it's enabled you cannot inject.

Yimura:
> You cannot go online if BattlEye is disabled and if it's enabled you cannot inject.

Yes, that makes complete sense. I was a bit blind sighted by the fixing YimMenu for this update and missed the disabling BE part.

Yimura (approved):
LGTM, untested as I'm currently still at work.

gir489returns:
> You cannot go online if BattlEye is disabled and if it's enabled you cannot inject.

Could it be as simple as just spoofing that we are running BattleEye, but disabling it?

gir489returns (approved):

PR #3705, opened 2024-09-20
Fix crash when loading the settings file

Mr-X-GTA:


Yimura (approved):

Yimura:
This bug has been there for a good while, glad you found and reproduced it.

